PR #110, opened 2017-12-23
XCB bugfix: handle uppercase keys with modifiers correctly

stacksmith:


stacksmith:
Sorry, I will merge into previous pull request

PR #154, opened 2018-01-17
Build an executable for the ncurses interface

vindarel:
Hi,

I suggest to distribute Lem executables in addition of the Roswell installation process. This is the way I found and I'd like your feedback. Indeed, downloading an executable is easier and lighter than installing roswell and its dependencies, I'm sure it will make more people try Lem, and it's maybe easier to develop with (I don't know how you develop with Roswell).

A CI system can easily build a new binary at every new tag push.

I have many edits like so

    +(asdf:defsystem "lem-c-mode" 

because I need this to `C-c C-k` in Slime and for the make target, but there may be another way if that bothers you.

I didn't try for Electron yet.

Regards

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,20 @@ all:
 
 %$(EXEEXT): %.ros
 	ros build $<
+
+build:
+	sbcl --load lem-core/lem-core.asd \
+	     --load lem-vi-mode/lem-vi-mode.asd \
+	     --load lem-lisp-mode/lem-lisp-mode.asd \
+	     --load lem-go-mode/lem-go-mode.asd \
+	     --load lem-c-mode/lem-c-mode.asd \
+	     --load lem.asd \
+	     --eval '(ql:quickload "lem")' \
+	     --eval '(ql:quickload "lem-ncurses")' \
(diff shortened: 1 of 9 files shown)


snmsts:
I hate this.

asciian:
You can make a new package ‘ lem-installer ‘ with this request. Could you like it?
I guess it is good way to distribute for brew or pkgsrc. I’m looking forward your work! Thanks.

vindarel:
@snmsts I've seen you more helpful. Seriously, do you dislike the ability to ship binaries or the form of this PR ? (which is ok, I asked for feedback)

Building binaries is a great feature of CL. If it is in opposition with the project, I don't get it.

snmsts:
i'm a bit irritated to prepared something without Roswell
. Sorry about the attitude. But still,I don’t like how it achieved.
I’ll prepare binary in other way.Would you mind just keep it close and open a issue?

vindarel:
Ok, thanks for the clarification.

> But still,I don’t like how it achieved.
> I’ll prepare binary in other way.

+1 !

The issue: https://github.com/cxxxr/lem/issues/155

PR #170, opened 2018-03-16
add a dependency: keyboard-layout to handle different keyboard layout

picospuch:
add keyboard-layout package from atom.
the convert power comes from UCKeyTranslate on macOS.

if it works, the only disadvantage is that keyboard-layout brings native binary code dependency to this project, not that pure.
 
may the users not care.

incidentally, do `npm install` at project root directory to update the project dependency chain first before test.


picospuch:
@t-cool I have no JIS keyboard in hand, please help me test this pr, thanks

PR #258, opened 2018-08-11
simplify clipboard

cxxxr:


myaosato:
I checked ```copy-region```, ```kill-region```, ```yank``` work. 　:+1:

PR #281, opened 2018-08-15
Removed form-feed character from source code

windymelt:
I found _form-feed_(`^L`) control character in source code;
I removed them because I thought it's not necessary.

Are there any custom in Lisp to treat _form-feed_ as special character?

> The form feed character is sometimes used in plain text files of source code as a delimiter for a page break, or as marker for sections of code. Some editors, in particular emacs and vi, have built-in commands to page up/down on the form feed character. This convention is predominantly used in Lisp code, and is also seen in C and Python source code.
(https://en.wikipedia.org/wiki/Page_break)

Anyway I thought we should remove _form-feed_ or maintain it and explain about it in documents, because it's not very popular rule.

### DIFF
--- a/lem-base/buffer-insert.lisp
+++ b/lem-base/buffer-insert.lisp
@@ -228,7 +228,6 @@
                     (decf offset-line)
                 :finally (shift-markers point offset-line 0)))))))
 
-
 (declaim (inline call-before-change-functions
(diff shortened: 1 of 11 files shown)


windymelt:
I have known it's common (but little legacy) rule. Closing.

PR #360, opened 2019-01-24
add many :descriptions

pupcraft:
add descriptions to add the ASDF systems for quicklisp


pupcraft:
Progress can be tracked cxxxr/lem#359

Issue #369, opened 2019-03-28
Add a git hash to a version information, reported by M-x lem-version


svetlyak40wt:
This will be useful because we are installing Lem from the git and some users can have same version number (like 1.5), but different behaviour because they have checked out Lem from different commits.

cxxxr:
This issue was resolved by merging #370.

PR #370, opened 2019-03-28
Now lem-version command output a git hash if it is available.

svetlyak40wt:
This should close issue #369.

### DIFF
--- a/lib/core/command.lisp
+++ b/lib/core/command.lisp
@@ -589,10 +589,25 @@
                     (message "Loaded ~A." name))
         (t (message "Can't find Library ~A." name))))
 
+(defun get-git-hash (&optional (system :lem))
+  (let* ((component (asdf:find-system system))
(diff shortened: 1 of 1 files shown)


snmsts:
the git hash should be fixed when the source code is compiled,not the value of when the ``get-git-hash`` is called.

svetlyak40wt:
Is it ok to get this information at compile time in the ASD file?


чт, 28 мар. 2019 г., 9:48 SANO,Masatoshi notifications@github.com:

snmsts:
define-compiler-macro ?

svetlyak40wt:
I never used it before and would like to know more about compiler macroses.

чт, 28 мар. 2019 г., 14:33 SANO,Masatoshi notifications@github.com:

cxxxr:
Thank you.
I will fix that problem after merging.

PR #374, opened 2019-04-19
Fix disp-set-overlay for one line

pupcraft:
If you mark, then scroll down, then scroll up and have the cursor on the first line, it renders wrong. 
1. setting the mark

2. scrolling down

3. almost there

4. incorrectly colors the rest


cxxxr:
Thanks

PR #396, opened 2019-06-17
Better support for comments in paredit-mode

jessymilare:


jessymilare:
I've worked on many changes. I'll create branches so that pull requests are monothematic and better to review.

PR #399, opened 2019-06-22
Many fixes and changes for paredit-mode

jessymilare:
- Use `syntax-escape-point-p` instead of testing if char before point is `#\\`, which avoid mistakes when the backslash is itself escaped.
- Simplify `insert-doublequote`.
- `paredit-backward-delete` and `paredit-forward-delete` didn't behave correctly near an escaped paren or doublequote. Separated cases to make clearer what the function does in each case.
- `paredit-kill` didn't work as expected inside strings or block comments and also would kill too much if the sexp was followed by several lines with whitespaces and/or comments. Now the code is clearer.


cxxxr:
Thank you.

PR #400, opened 2019-06-23
Remove compile and load time dependency on Quicklisp

jessymilare:
Dynamically probe for :quicklisp package and call its symbols on
demand. References to symbols in that package are made through
`uiop:symbol-call` and `uiop:find-symbol*`. The rule doesn't apply
to '.ros' files.


contrib/version-up/version.lisp
```diff
@@ -3,7 +3,7 @@
 ;;;don't edit above
 
 (defun path (target)
-  (let ((path (ql:where-is-system target)))
+  (let ((path (uiop:symbol-call :quicklisp 'where-is-system target)))
```
phoe: Use `'#:where-is-system` instead to avoid interning a new symbol in current package. Or use a keyword.

modes/python-mode/lem-python-mode.asd
```diff
@@ -1,7 +1,18 @@
...
+;;; See: https://common-lisp.net/project/asdf/asdf/The-defsystem-grammar.html#index-_003aweakly_002ddepends_002don
+
+(defsystem "lem-python-mode/run"
+  :depends-on ("lem-pyton-mode" "lem-process")
+  :serial t
+  :components ((:file "run-python")))
+
+(defmethod perform :after (operation
+                           (system (eql (find-system "lem-python-mode"))))
+  (when (uiop:symbol-call :quicklisp :where-is-system :async-process)
```
phoe: This is going to blow up on systems that do not have Quicklisp installed (e.g. ASDF-only systems). Ensure that the QL package can be found first before executing this code.

modes/scheme-mode/lem-scheme-mode.asd
```diff
@@ -11,8 +11,18 @@
...
+
+(defsystem "lem-scheme-mode/repl"
+  :depends-on ("lem-scheme-mode" "lem-process")
+  :serial t
+  :components ((:file "eval")
                (:file "repl")))
+
+(defmethod perform :after (operation
+                           (system (eql (find-system "lem-scheme-mode"))))
+  (when (uiop:symbol-call :where-is-system :quicklisp :async-process)
```
phoe: Ditto.

jessymilare:
Made the recommended changes. Loaded using ASDF only and it seems to work.

PR #403, opened 2019-07-02
Fix buffer-package for non-standard package value

jessymilare:
If `(buffer-value buffer "package")` returned a value that was not
a string designator, and error was signaled and Lem died.
E.g. the file "iterate.lisp" from project `:iterate` contains the
following property list:
```
;;;-*- syntax:COMMON-LISP; Package: (ITERATE :use "COMMON-LISP" :colon-mode :external) -*-
```
Therefore, opening that file and calling `lisp-mode` killed Lem.
Now, wrap call with ignore-errors and use default if NIL is returned.

### DIFF
--- a/modes/lisp-mode/lisp-mode.lisp
+++ b/modes/lisp-mode/lisp-mode.lisp
@@ -138,10 +138,11 @@
     (self-connect)))
 
 (defun buffer-package (buffer &optional default)
-  (let ((package-name (buffer-value buffer "package")))
-    (or (ignore-errors (string-upcase package-name))
-        (ignore-errors (string-upcase (car package-name)))
-        default)))
+  (let ((package-name (buffer-value buffer "package" default)))
+    (if (symbolp package-name)
+        package-name
+        (or (ignore-errors (string-upcase package-name))
+            (ignore-errors (string-upcase (car package-name)))))))
 
 (defun (setf buffer-package) (package buffer)
   (setf (buffer-value buffer "package") package))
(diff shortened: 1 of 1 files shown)


modes/lisp-mode/lisp-mode.lisp
```diff
@@ -140,7 +140,9 @@
 (defun buffer-package (buffer &optional default)
   (let ((package-name (buffer-value buffer "package")))
     (if package-name
```
snmsts: This if might not needed anymore.

jessymilare: In my system, Lem still dies when opening file "iterate.lisp".

jessymilare: > This if might not needed anymore.

Excuse-me, I didn't understand what might not be needed. Do you mean the test `(if package-name [...])`?

jessymilare: Well, removed that test and it seems to work.

jessymilare: Another problem is with `*lisp-repl*` buffer in Lisp mode: the package is set to "NIL". Made one more change so that doesn't happen anymore.

snmsts:
like this?

```
(defun buffer-package (buffer &optional default)
  (let ((package (buffer-value buffer "package" default))) 
    (cond ((null package) default)
          ((or (symbolp package)
               (stringp package)) 
           (string-upcase package))
          ((consp package)
           (string-upcase (car package))))))
```

PR #406, opened 2019-07-09
Complex indent method fix

jessymilare:
Lem didn't correctly indent Lisp forms whenever the
method registered for a symbol ended in `&rest <integer>.
For instance, `multiple-value-bind` was indented like that:
```
  (multiple-value-bind (a b
                          c d
                          e f)
      g
    h)
```
However, it should be:
```
  (multiple-value-bind (a b
                        c d
                        e f)
      g
    h)
```

I also added an error-check before setting some Lisp indentation method. Tested error checking all methods both in static and dynamic tables, no error was found. Indentations now is like Emacs in all tests I made.

### DIFF
--- a/lib/lisp-syntax/indent.lisp
+++ b/lib/lisp-syntax/indent.lisp
@@ -69,7 +69,9 @@
                         (malformed "&body is not the last element")))
                      ((&rest)
                       (unless (and (consp method-rest) (null (cdr method-rest)))
-                        (malformed "&rest must be followed by only one element")))
+                        (malformed "&rest must be followed by only one element"))
+                      (when (eq '&rest (car method-rest))
+                        (malformed "&rest cannot be followed by another &rest")))
                      ((&whole)
                       (malformed "&whole can only be the first element in a submethod"))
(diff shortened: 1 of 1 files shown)


jessymilare:
Reading the documentation from Emacs, the function is treating `&whole` wrongly. It was already wrong before: `&whole` should be accepted anywhere. Also, sublists indentation count up from the 0th element (it was previously "skipped". I'm gonna change it now according to documentation.

jessymilare:
Emacs spec is not what it does actually. Changed `compute-complex-indent-method` based on Emacs source code,

By the way, I see this code is far from optimal when using modes like Emacs aggressive-indent-mode. Now I understand why Emacs have slow performance with that mode and very large sexps. I believe it would be much better if we previously parse the indent-mode before placing it in the table, transforming it in a vector, for instance. Then, for indentation, we use `(aref parsed-method n)` instead of iterating inside a list. It would be a cleaner and better code.

Plus, indenting regions could be better coded so that paths are computed more efficiently, reusing the same path prefix for an entire sexp.

PR #428, opened 2019-09-26
Refactor timer

kenoss:
I made code more clear.
I couldn't understand spec of idle timers at first glance.


lib/core/timer.lisp
```diff
@@ -70,33 +75,43 @@
...
-                    (not (timer-idle-p timer)))
-               (setf (timer-last-time timer)
-                     (get-internal-real-time)))
-              (t
-               (unless (timer-idle-p timer)
-                 (setf (timer-alive-p timer) nil))
-               (push timer promised-timers)))))
-    (setq *timer-list* (set-difference *timer-list* promised-timers))
-    (dolist (timer update-timers)
+  (let* ((tick-time (get-internal-real-time))
```
kenoss: I wonder whether this line can cause a change of behavior or not.  In old code, if the return value of `get-internal-real-time` would chaneg in this block, `update-timers` also change.

And, another view point.  In the case that `get-internal-real-time` uses system call, multiple calls of this may cause speed penalty.  It is a common technique in event loop program, e.g. API server, to cache time in the head of "some block".  I prefer that there's no ambiguity nor uncertainty, and that we cache time for each `read-key-1`.

lib/core/timer.lisp
```diff
@@ -70,33 +75,43 @@
...
-     (setf *idle-timer-list* (delete timer *idle-timer-list*))
-     (setf *timer-list* (delete timer *timer-list*)))
-    (t
-     (setf (timer-alive-p timer) nil)
-     (setf *timer-list* (delete timer *timer-list*)))))
+  (progn
+    (setf (timer-alive-p timer) nil)
+    (if (timer-idle-p timer)
+        (setf *idle-timer-list* (delete timer *idle-timer-list*))
+        (setf *timer-list* (delete timer *timer-list*)))))
```
cxxxr: `*timer-list*` はtimer-idle-pの値によらずsetfされます

cxxxr: この変更では*timer-list*にidle-timerが含まれなくなったからこれでいいみたいですね

kenoss:
Thank you for reviewing.

PR #430, opened 2019-09-27
Refactor `read-key-1`

kenoss:
This is based on https://github.com/cxxxr/lem/pull/428 . Please check that first.

Essential difference is https://github.com/kenoss/lem/compare/feature/refactor-timer...kenoss:feature/refactor-read-key-1?expand=1 .  Dependency is due to change of function name.

I have yet another proposal.  See comment below.


lib/core/input.lisp
```diff
@@ -45,17 +45,19 @@
...
+              (let ((e (read-event nil)))
+                (when (key-p e)
+                  (return-from read-key-1 e)))))
+            ((minusp ms)
+             (update-timer))
+            (t
+             (let ((e (read-event (float (/ ms 1000)))))
+               (when (key-p e)
+                 (return e)))
+             (update-timer))))))
```
kenoss: As I mentioned in slack, this `update-timer` call is essentially redundant and confusing.  I want to delete this.

PR #443, opened 2019-11-03
Fixed Nim multiline comments

FemtoEmacs:
Fixed Nim multiline comment


FemtoEmacs:
Nim multiline comment was not working, this pull request fixed it.

The pull request also added new data types and keywords to the highlight scheme.

PR #457, opened 2019-12-14
Fixed an issue when connection to swank didn't initialized propertly

svetlyak40wt:
The root of the problem was that on OSX `swank:*communication-style*` is equal to `:fd-handler`
for some reason. In this case `swank:create-server` does not start threads which
accept connections on TCP port and process incoming messages.

To make it work propertly, we need to specify that communication style is `:spawn`.

This commit closes issue https://github.com/cxxxr/lem/issues/411.

### DIFF
--- a/modes/lisp-mode/lisp-mode.lisp
+++ b/modes/lisp-mode/lisp-mode.lisp
@@ -121,7 +121,7 @@
 (defun self-connect ()
   (let ((port (random-port)))
     (let ((swank::*swank-debug-p* nil))
-      (swank:create-server :port port))
+      (swank:create-server :port port :style :spawn))
     (slime-connect *localhost* port nil)
     (update-buffer-package)
     (setf *self-connected-port* port)))


Hamayama:
Uhm, I checked the documents of slime/swank and found the following page.

https://common-lisp.net/project/slime/doc/html/Communication-style.html#Communication-style

```
The default request handling style is chosen according to the capabilities of your Lisp system.
The general order of preference is :SPAWN, then :SIGIO, then :FD-HANDLER, with NIL as a last resort.
You can check the default style by calling SWANK-BACKEND::PREFERRED-COMMUNICATION-STYLE.
You can also override the default by setting SWANK:*COMMUNICATION-STYLE* in your Swank init file.
```

So, I wonder fixing style to spawn might cause hang in other environments.

I also wonder why this problem occurs in the specified pc.

I searched PREFERRED-COMMUNICATION-STYLE by github search.

https://github.com/slime/slime/search?p=1&q=PREFERRED-COMMUNICATION-STYLE&unscoped_q=PREFERRED-COMMUNICATION-STYLE

Can you find the condition of this problem in this page?

If the specified condition is found, we can set the style only when the condition is true.

Hamayama:
Uhm, if lem needs a separate thread to process self connection to swank server,
it might be reasonable to set style to :swank .

Sorry for the noise ...

cxxxr:
Thank you.

PR #486, opened 2020-05-05
[vi-mode] Add x[it][!] support to ex-command

leegbestand:
This request adds the `:x` or `:xit` command to VI mode.
It does the same thing as `:wq` except it only writes
when changes have been made.

#### Why
Part of vi(m) and my go-to key to write and quit.

#### How to execute
Open a file, enable VI mode and execute `:x` 
Lem should write to the currently open file and quit.


gos-k:
thank you

PR #493, opened 2020-06-18
Yet another ChangeLog update.

svetlyak40wt:


svetlyak40wt:
It would be cool if commits with new features and important fixes will add a corresponding piece to the ChangeLog.md!

PR #520, opened 2020-09-06
Now if color attribute does not specify foreground or background color, the default color is used explicitly

svetlyak40wt:
This pull fixes colour glitches in case if colour theme sets the background and foreground colours and Lem's frontend is NCurses.

Two pictures to illustrate.

Before this patch:

Screenshot 2020-09-06 at 22 54 58

After this patch:

Screenshot 2020-09-06 at 22 51 16

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,15 @@
 # ChangeLog
 
+## Not released
+
+### Fixed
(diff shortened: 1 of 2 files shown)


cxxxr:
Thank you

PR #540, opened 2021-02-04
fix unknown symbol: maybe-quickload

longlene:
maybe-quickload is in lem, `(asdf:make :lem/executable)` will build failed


cxxxr:
Thank you

PR #542, opened 2021-02-15
Fix the way to enable frame-multiplexer

t-sin:
Sorry. In this case, the correct way is enabling via editor-variable.


cxxxr:
Thanks

PR #545, opened 2021-03-03
Add LSP spec for Rust's `rls`

t-sin:
This PR defines lsp-mode's language spec for Rust's [rls](https://github.com/rust-lang/rls).


cxxxr:
Thank you

PR #608, opened 2023-03-22
Add var and defun for removing trailing whitespace on save

rearman:
Changes made in src/file-command.lisp

This change adds the variable `delete-trailing-whitespace-on-writing-file` and the defun `clear-trailing-whitespace-on-write`.
I tried to keep it in the same style as `Add-Newline-at-EOF-on-Writing-File` and `add-newline-at-eof`.

I tested by setting the variable to `t`, introducing trailing whitespace in multiple buffers, then saving using `save-current-buffer` and `save-some-buffers` to ensure the whitespace was removed in the proper places.  I then did the same with the variable set to `nil`, and ensured that the whitespace was not removed.

### DIFF
--- a/src/file-command.lisp
+++ b/src/file-command.lisp
@@ -5,13 +5,13 @@
 
 (defun maybe-create-directory (directory)
   (when (prompt-for-y-or-n-p
-         (format nil "Directory does not exist: ~A. Create" directory))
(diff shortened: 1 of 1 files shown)


cxxxr:
Thanks for the PR.
I was hoping to have this feature someday.

PR #612, opened 2023-05-04
[BUG] • Fix issue with recent changes to show parens that broke vi function…

theangelperalta:
• Fix issue with recent changes to show parens that broke vi functionality


cxxxr:
Thank you!

PR #613, opened 2023-05-05
[BUG] • Killring yank function failing to return options

theangelperalta:
• address bug with current killring function implementation failing to return item-options

TODO: Add unit test


theangelperalta:
@cxxxr, feel free to provide feedback on the unit tests or the PR in general. Thank you for working on this project.

src/killring.lisp
```diff
@@ -19,5 +19,6 @@
     (copy-to-clipboard (peek-killring-item (current-killring) 0))))
 
 (defun yank-from-clipboard-or-killring ()
-  (or (and (enable-clipboard-p) (get-clipboard-data))
-      (peek-killring-item (current-killring) 0)))
+  (multiple-value-bind (str options) (peek-killring-item (current-killring) 0)
+    (or (and (enable-clipboard-p) (get-clipboard-data))
+        (values str options))))
```
cxxxr, line 24: I like that the return value is more explicit than before the change. :+1:

src/package.lisp
```diff
@@ -15,6 +15,10 @@
    :remove-command
    :find-command
    :exist-command-p)
+  ;; killring.lisp
+  (:export
+    :copy-to-clipboard-with-killring
+    :yank-from-clipboard-or-killring)
```
cxxxr, line 21: Indeed, it is better to have this exported. :+1:

tests/common/killring.lisp
```diff
@@ -72,6 +73,26 @@
...
+    
+    ;; clipboard enabled
+    (let ((lem::*enable-clipboard-p* t)
+          (lem::*killring* killring)
+          (expected-result "In LEM we trust."))
+      (lem::invoke-frontend (lambda ()))
+      (copy-to-clipboard-with-killring expected-result)
+      (ok (equal expected-result
+                 (yank-from-clipboard-or-killring))))))
+
```
cxxxr, line 95: I really like the addition of the test! :+1:

tests/common/killring.lisp
```diff
@@ -72,6 +73,26 @@
...
+(deftest external-option
+  (let ((killring (make-killring 2)))
+    (with-killring-context (:appending t :before-inserting t)
+      (push-killring-item killring "baz" :options :test))
+
+    ;; clipboard disabled
+    (let ((lem::*enable-clipboard-p* nil)
+          (lem::*killring* killring))
+      (ok (equal '("baz" (:test))
+                 (multiple-value-list (yank-from-clipboard-or-killring)))))
```
cxxxr, line 85: This might be a good addition with a good interface to encourage encapsulation.

tests/common/killring.lisp
```diff
@@ -1,7 +1,8 @@
 (defpackage :lem-tests/killring
   (:use :cl
         :rove
-        :lem/common/killring))
+        :lem/common/killring
+        :lem))
```
cxxxr, line 5: I would like to make the common package independent of lem.
How about moving the external-option test to tests/killring.lisp and testing in lem/src/killring.lisp?

tests/common/killring.lisp
```diff
@@ -72,6 +73,26 @@
...
+    (let ((lem::*enable-clipboard-p* nil)
+          (lem::*killring* killring))
+      (ok (equal '("baz" (:test))
+                 (multiple-value-list (yank-from-clipboard-or-killring)))))
+    
+    ;; clipboard enabled
+    (let ((lem::*enable-clipboard-p* t)
+          (lem::*killring* killring)
+          (expected-result "In LEM we trust."))
+      (lem::invoke-frontend (lambda ()))
```
cxxxr, line 91: Since you are using lem:invoke-frontend, it may be better to mock it with a fake interface

For example, there is this kind of macro.
https://github.com/lem-project/lem/blob/0c8af91970191d8a997fc82a81eb6d7c4624cd0b/tests/cursors.lisp#L45

cxxxr (commented):
Thanks for the great pull request.
It would be better if some of the areas I commented on could be improved :pray:

tests/common/killring.lisp
theangelperalta: Thanks, for that information. I was able to update my PR to use this mocking interface 👍.

cxxxr:
Excellent commits! Thank you very much.

PR #618, opened 2023-05-13
[VI-MODE] • Missing change from vi-mode package name migration

theangelperalta:
Also notice that `e2e` tests are currently not being run by rove. Is this intentional?


cxxxr:
Thank you for the nice PR.

> Also notice that e2e tests are currently not being run by rove. Is this intentional?

Yes. The test has become flaky, and it fails depending on the order of testing, so I had temporarily stopped running it.

PR #619, opened 2023-05-13
[VI-MODE] • Clear refactor

theangelperalta:
Some light refactoring to change "clear" to "change." , as well as address difference between `cw` vs. `dw` from standard VIM.


cxxxr:
that's good!
thank you.

PR #620, opened 2023-05-13
[VI-MODE] • Word refactor

theangelperalta:
• Excuse the other commit, my apologizes


cxxxr:
I believe it's a good change as the first step in improving vi-word-char-p.
Thank you very much!

PR #622, opened 2023-05-16
Close socket gracefully when `slime-quit'

kchanqvq:
Currently, lem leaves the connection socket open until the lem Lisp image quits. This patch close it on `slime-quit`.


cxxxr:
Thank you!

PR #624, opened 2023-05-16
[VI-MODE] • Temporary work around for macOS issue with global yank

theangelperalta:
Without this change options are always ignored if clipboard data is present, thus, breaking `yy` & `p` operations.


cxxxr:
Thank you very much!

PR #637, opened 2023-05-27
Add a launcher script for SDL2 frontend for roswell users

t-sin:
This is a  launcher script enabled when lem is installed via [roswell](https://github.com/roswell/roswell).


cxxxr:
Thanks!

Issue #644, opened 2023-05-28
First step in creating a system to maintain documentation


cxxxr:
## Context and scope

There is currently an issue with not having well-maintained user guides and developer documentation.
As a first step, I tried to consolidate commands and key bindings in d145f76c07aba62df0f2d69adcd0684bbee12627.
However, preparing and maintaining these for all functionalities is challenging, as it lacks maintainability and scalability.

One possible solution is to generate documentation using command definitions, key bindings, and docstrings, and present that documentation on Lem with clickable links.

## Goals

The initial goal is to enable automatic generation of documentation similar to d145f76c07aba62df0f2d69adcd0684bbee12627.

### Complete documentation
Only easy problems will be tackled at first.

### Rich markdown viewer
This is done by another issue

### Multilingual support
Currently, only English will be provided

## ToDo

First, the commands will be divided into packages based on their genres. This will allow writing descriptions about each command in the package's docstring.

The following files are likely to be targeted:

- primitive-command.lisp
- self-insert-command.lisp
- file-command.lisp
- window-command.lisp
- help-command.lisp
- word-command.lisp
- sexp-command.lisp
- multiple-cursors-command.lisp
- default-keymap.lisp

After that, I will extract information from the runtime and generate Markdown files. 
These Markdown files will be placed in the `docs/` directory. (The directory may be changed to a more appropriate location later.)

- [X] https://github.com/lem-project/lem/pull/650
- [x] Add docstring to each command
- [x] Add a link to the location of each command definition
- [x] Prepare two outputs for markdown and lem buffer
- [x] https://github.com/lem-project/lem/pull/669

Sasanidas:
>One possible solution is to generate documentation using command definitions, key bindings, and docstrings, and present that documentation on Lem with clickable links.

Maybe something similar to https://github.com/Wilfred/helpful , maybe extending the typeout mode (or having a major-mode in that buffer) to reflect the syntax and having the buttons.

Sasanidas:
I took the liberty to try a make a webpage for more user friendly documentation, for now is in https://sasanidas.github.io/lem-page/ 

If you (@cxxxr) think that can be suitable for the project, I am more than happy to move it to the lem project workspace
(also ping @vindarel  that I saw making some documentation)

cxxxr:
This is really wonderful!
If this document is included in the lem-project, I would be very grateful and will continue to maintain it.

Sasanidas:
[...]

PR #646, opened 2023-05-28
refactor command definitions

cxxxr:
I divided the definitions of each command into packages corresponding to their categories.
This is a refactoring to categorize commands and document them accordingly.

related issue: https://github.com/lem-project/lem/issues/644


cxxxr:
- [ ] regression: `(lem:lem <filename>)`

PR #647, opened 2023-05-29
Basic support for Elixir language

Sasanidas:
This commit has basic support for the Elixir programming language, still there are a couple of thing to improve/fix:

- The end function is not the best, there is a better way to calculate the end that have the same indentation as the function header,
this method can also be used for other languages that also end functions in "end" (like Lua).

- The lsp doesn't work, I think the configuration it's correct and at the beginning it's showing some log messages that for example GNU Emacs eglot ignores.

This is the output that eglot ignores and keep working, Lem only show this in a buffer and trigger the next error:
```
(:jsonrpc "2.0" :method "window/logMessage" :params
	  (:message "Started ElixirLS v0.14.6" :type 3))
[server-notification] Mon May 29 13:45:59 2023:
(:jsonrpc "2.0" :method "window/logMessage" :params
	  (:message "Running in /mnt/Data/Programming/atlas/remo" :type 3))
[server-notification] Mon May 29 13:45:59 2023:
(:jsonrpc "2.0" :method "window/logMessage" :params
	  (:message "ElixirLS built with elixir \"1.14.4\" on OTP \"24\"" :type 3))
[server-notification] Mon May 29 13:45:59 2023:
(:jsonrpc "2.0" :method "window/logMessage" :params
	  (:message "Running on elixir \"1.14.4 (compiled with Erlang/OTP 24)\" on OTP \"24\"" :type 3))
[server-notification] Mon May 29 13:45:59 2023:
(:jsonrpc "2.0" :method "window/logMessage" :params
	  (:message "Protocols are not consolidated" :type 3))
[server-notification] Mon May 29 13:45:59 2023:
(:jsonrpc "2.0" :method "window/showMessage" :params
	  (:message "OTP compiled without EEP48 documentation chunks" :type 2))
[server-notification] Mon May 29 13:45:59 2023:
(:jsonrpc "2.0" :method "window/logMessage" :params
	  (:message "OTP compiled without EEP48 documentation chunks. Language features for erlang modules will run in limited mode. Please reinstall or rebuild OTP with appropriate flags." :type 2))

```

 **This is the error that it triggers:**

Selection_002

- The run-elixir "works" but sometimes it returns the same thing instead of processing it (input 2+2 -> output 2+2, not performing the sum), also by default iex have a lot of colors in the terminal, Lem doesn't seem to like that.

I'm using GNU Emacs regex for inspiration, using this package [pcre2el](https://github.com/joddie/pcre2el) helped a lot converting the default Emacs regex to PCRE.

This was my first major mode, and I'm still not that familiar with the code base, so any tips or additions are more than welcome.

### DIFF
--- a/modes/elixir-mode/elixir-mode.lisp
+++ b/modes/elixir-mode/elixir-mode.lisp
@@ -1,0 +1,107 @@
+(defpackage :lem-elixir-mode
+  (:use :cl :lem :lem/language-mode :lem/language-mode-tools)
+  (:export :*elixir-mode-hook*
+           :elixir-mode))
+(in-package :lem-elixir-mode)
+
+(defvar *elixir-keywords*
+  '("case" "cond" "for" "if" "quote" "raise" "receive" "send" 
+    "super" "throw" "try" "unless" "unquote" "unquote_splicing" "with" 
+    
+    "import" "require" "use" "alias"
+    
+    "def" "defp" "defmodule" "defprotocol"
+    "defmacro" "defmacrop" "defdelegate"
+    "defexception" "defstruct" "defimpl"
+    "defguard" "defguardp" "defcallback"
+    "defoverridable" 
+    "end" "after" "else" "rescue" "catch" 
+
+    "not" "and" "or" "when" "in"
+
+    "fn" "do" "end" "after" "else" "rescue" "catch"
+
+    "->" "<-" "=>" "|>"))
+
+(defvar *elixir-boolean-literals*
+  '("true" "false"))
+
+(defvar *elixir-null-literal*
+  '("nil"))
+
+;; From:  http://elixir-lang.org/getting-started/basic-operators.html
+(defvar *elixir-operators*
+  '("+" "*" "++" "--" "<>" "||" "&&" "!"
+    "==" "!=" "===" "!==" ">=" "<="))
+
+
+(defun tokens (boundary strings)
+  (let ((alternation
+          `(:alternation ,@(sort (copy-list strings) #'> :key #'length))))
+    (if boundary
+        `(:sequence ,boundary ,alternation ,boundary)
+        alternation)))
+
+
+
+(defun make-tmlanguage-elixir ()
+  (let* ((patterns (make-tm-patterns
+                    (make-tm-match ":\\w+" 
+                                   :name 'syntax-constant-attribute) 
+                    (make-tm-match "\\w+:"
+                                   :name 'syntax-constant-attribute)
+                    (make-tm-line-comment-region "#")
+                    (make-tm-string-region "\"")
+                    (make-tm-string-region "'")
+                    (make-tm-string-region "\"\"\"")
+                    (make-tm-match (tokens :word-boundary 
+                                           (append *elixir-boolean-literals*
+                                                   *elixir-null-literal*))
+                                   :name 'syntax-constant-attribute)
+                    (make-tm-match (tokens :word-boundary *elixir-keywords*)
+                                   :name 'syntax-keyword-attribute)
+                    (make-tm-match (tokens nil *elixir-operators*)
+                                   :name 'syntax-builtin-attribute))))
+    (make-tmlanguage :patterns patterns)))
+
+(defvar *elixir-syntax-table*
+  (let ((table (make-syntax-table
+                :space-chars '(#\space #\tab #\newline)
+                :paren-pairs '((#\( . #\))
+                               (#\{ . #\})
+                               (#\[ . #\]))
+                :string-quote-chars '(#\" #\')
+                :block-string-pairs '(("\"\"\"" . "\"\"\"")
+                                      ("'''" . "'''"))
+                :line-comment-string "#"))
+        (tmlanguage (make-tmlanguage-elixir)))
+    (set-syntax-parser table tmlanguage)
+    table))
+
+(define-major-mode elixir-mode language-mode
+    (:name "elixir"
+     :keymap *elixir-mode-keymap*
+     :syntax-table *elixir-syntax-table*
+     :mode-hook *elixir-mode-hook*)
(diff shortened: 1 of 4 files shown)


modes/elixir-mode/run-elixir.lisp
```diff
@@ -0,0 +1,86 @@
...
+(in-package :lem-elixir-mode.run-elixir)
+
+(defvar *elixir-run-command* "iex")
+
+(defvar *process* nil)
+
+(define-major-mode run-elixir-mode lem-elixir-mode:elixir-mode
+    (:name "elixir"
+     :keymap *run-elixir-mode-keymap*
+     :syntax-table lem-elixir-mode::*elixir-syntax-table*)
```
cxxxr, line 15: Wouldn't this be better to export the symbol?
The same goes for `*elixir-mode-keymap*`.

modes/elixir-mode/run-elixir.lisp
```diff
@@ -0,0 +1,86 @@
...
+    buffer))
+
+(defun output-callback (string)
+  (let* ((already-exists (repl-buffer-exists-p))
+         (buffer (get-repl-buffer))
+         (p (buffer-point buffer)))
+    (buffer-end p)
+    (setf string (ppcre:regex-replace-all "\\r\\n" string (string #\newline)))
+    (insert-string p string)
+    (when (ppcre:scan "^(iex|\.\.\.)\(.+\)" (line-string p))
```
cxxxr, line 56: Doesn't the backslash inside a string literal in Common Lisp disappear if there is only one?
```
"\." => "."
```

modes/elixir-mode/lem-elixir-mode.asd
```diff
@@ -0,0 +1,9 @@
+
+(defsystem "lem-elixir-mode"
```
cxxxr, line 2: https://github.com/lem-project/lem/blob/97e322b8cd0fe31af3614516cf9cfac125587ec9/lem.asd#L122

To enable this mode, it must be added to the dependencies of this system.
Sorry it's not clear.

modes/elixir-mode/elixir-mode.lisp
```diff
@@ -0,0 +1,107 @@
...
+                :string-quote-chars '(#\" #\')
+                :block-string-pairs '(("\"\"\"" . "\"\"\"")
+                                      ("'''" . "'''"))
+                :line-comment-string "#"))
+        (tmlanguage (make-tmlanguage-elixir)))
+    (set-syntax-parser table tmlanguage)
+    table))
+
+(define-major-mode elixir-mode language-mode
+    (:name "elixir"
```
cxxxr, line 82: I would like it to be a capital-case because it is displayed on the modeline.

modes/elixir-mode/lsp-config.lisp
```diff
@@ -0,0 +1,14 @@
...
+  (:use :cl
+         :lem-lsp-mode
+         :lem-lsp-base/type))
+
+(in-package :lem-elixir-mode/lsp-config)
+
+(define-language-spec (elixir-spec lem-elixir-mode:elixir-mode)
+  :language-id "elixir"
+  :root-uri-patterns '("mix.exs")
+  :command '("sh" "language_server.sh")
```
cxxxr, line 11: Is this program being put in place where the path goes?

cxxxr (commented):
I thought it was generally good, very wonderful, thank you.
I have a few comments and questions.

modes/elixir-mode/run-elixir.lisp
Sasanidas: Maybe? I'm not an expert in regex, I converted the Emacs regex, the Emacs equivalent is:
"^\\(iex\\|\\.\\.\\.\\)(.+)>"

The otuput comes from the command rxt-elisp-to-pcre in pcre2el

modes/elixir-mode/lsp-config.lisp
Sasanidas: It's not usually in the path tho, but it also doesn't have like a default location, so I just put the string in a variable that the user can edit to make it work (this probably should me documented in some place)

modes/elixir-mode/elixir-mode.lisp
Sasanidas: Done :+1:

modes/elixir-mode/lem-elixir-mode.asd
Sasanidas: Added, I put the mode at the end of the list

modes/elixir-mode/run-elixir.lisp
Sasanidas: Done, this was because I just copy and paste from the python mode, should it also be changed there?

cxxxr: Yes, that was an oversight.

modes/elixir-mode/run-elixir.lisp
cxxxr: In ppcre, no backslash seems to be correct.

```
CL-USER> (ppcre:parse-string ".\\.")
(:SEQUENCE :EVERYTHING #\.)
```

Sasanidas: Oh, interesting, good to know :+1:  (I modify the regex, to remove the backslash)

modes/elixir-mode/lsp-config.lisp
Sasanidas: Done :+1:

vindarel:
Well done! What if you added a README in the elixir-mode directory to say, in a few bullet points, what this mode does and doesn't?

Sasanidas:
Thanks for writting some key points :raised_hands: @vindarel 

I also added the link to this merge request, so if people want to troubleshot it has some context :+1:

Issue #653, opened 2023-05-31
vi-mode break lem


Sasanidas:
When you are in the top right and press "b" to go a word back (calling the function `vi-backward-word-begin`) lem crashes with this error:
![fermin@fdebian: ~_001](https://github.com/lem-project/lem/assets/26670956/5ea6fedc-1da4-4e36-b577-435e122ba01f)

After this, it gets completely unresponsive and I have to restar the terminal

Compiled with SBCL 2.2.11 in Linuxmint 21 vanessa (AMD cpu)

Sasanidas:
Seems like the vi backward word is exhausting the stack:

![-sly-db for sbcl (thread 2)- - GNU Emacs at fdebian_001](https://github.com/lem-project/lem/assets/26670956/75d97ed8-2e7b-4265-969b-dbb9b47c5911)

Sasanidas:
Fixed https://github.com/lem-project/lem/pull/654

PR #654, opened 2023-05-31
Fix vi-backward-word-begin stack bug

Sasanidas:
I think his can fix the issue https://github.com/lem-project/lem/issues/653


cxxxr:
Thank you!

PR #655, opened 2023-05-31
fix typo in name vi-open-adove

avitkauskas:
vi mode function vi-open-above was misspelled everywhere like vi-open-adove. This PR fixes the name.


cxxxr:
Thank you.
I always get this word wrong.

PR #662, opened 2023-06-01
Add change-package REPL shorcut

Sasanidas:
- Same functionality to the SLIME/Sly in-package command


cxxxr:
Looks very useful, thanks!

PR #663, opened 2023-06-01
Add cd repl function and get/set current lisp directory

Sasanidas:
Also other command from SLIME/Sly

### DIFF
--- a/modes/lisp-mode/lisp-mode.lisp
+++ b/modes/lisp-mode/lisp-mode.lisp
@@ -320,14 +320,16 @@
     (buffer-end (buffer-point repl-buffer))))
 
 (define-command lisp-current-directory () ()
-  (message "Current directory: ~a" *default-pathname-defaults*) )
+  (message "Current directory: ~a" 
+           (lisp-eval `(swank:default-directory))))
 
 (define-command lisp-set-directory (&key directory) ()
   (unless directory
     (setf directory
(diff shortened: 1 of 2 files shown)


modes/lisp-mode/lisp-mode.lisp
```diff
@@ -319,6 +319,16 @@
...
 
+(define-command lisp-current-directory () ()
+  (message "Current directory: ~a" *default-pathname-defaults*) )
+
+(define-command lisp-set-directory (&key directory) ()
+  (unless directory
+    (setf directory
+           (prompt-for-directory "New directory: " :directory (buffer-directory))))
+  (swank:set-default-directory 
+   (swank-backend:filename-to-pathname directory)))
```
cxxxr, line 330: Perhaps this only works for self-connection?
If you are connected to another process, you may want to use lisp-eval or something similar.
https://github.com/lem-project/lem/blob/main/modes/lisp-mode/lisp-mode.lisp#LL259C5-L259C5

cxxxr (commented):
Thank you for your excellent Pull Request! 
I have left comments.

modes/lisp-mode/lisp-mode.lisp
Sasanidas: Oh, you are right indeed! 

Thanks for the catch :+1: , I update the code to reflect it

cxxxr:
Thank you!

PR #666, opened 2023-06-02
Remove elixir server-path variable and add a way to change the lsp spec

Sasanidas:
So... I just realize that the variable only get expanded at compilation time, so when I changed the variable in the config the value doesn't get updated.

The idea is to just change the value in the user configuration with a good old setf command:
```
(defvar *fer/elixir-lsp-path*
  '("sh" "path-to"))

(setf (lem-lsp-mode/spec:spec-command
       (lem-lsp-mode/spec:get-language-spec 'lem-elixir-mode:elixir-mode))
      *fer/elixir-lsp-path*)

```

### DIFF
--- a/modes/lsp-mode/spec.lisp
+++ b/modes/lsp-mode/spec.lisp
@@ -55,6 +55,10 @@
   (check-type spec spec)
   (setf (get major-mode 'spec) spec))
 
+(defmethod (setf spec-command) ((command list)
+				(spec lem-lsp-mode/spec::spec))
+  (setf (slot-value spec 'command) command))
+
 (defun get-spec-command (spec &rest args)
   (let ((command (spec-command spec)))
     (if (functionp command)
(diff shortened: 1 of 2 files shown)


Sasanidas:
If you think this makes sense, I'll update the documentation in the page later :+1:

modes/lsp-mode/spec.lisp
```diff
@@ -55,6 +55,10 @@
   (check-type spec spec)
   (setf (get major-mode 'spec) spec))
 
+(defmethod (setf spec-command) ((command list)
+				(spec lem-lsp-mode/spec::spec))
```
cxxxr, line 59: Internal prefix maybe not needed.

cxxxr (approved):
I think it is good!

modes/lsp-mode/spec.lisp
Sasanidas: Done :+1:

cxxxr:
Thank you!

PR #668, opened 2023-06-03
feat(lisp-mode): Add command to eval buffer

jcs090218:
Some improvement. Ideas are from Emacs. ;)

### DIFF
--- a/modes/lisp-mode/lisp-mode.lisp
+++ b/modes/lisp-mode/lisp-mode.lisp
@@ -411,6 +411,7 @@
           (self-interactive-eval string)))))
 
 (define-command lisp-eval-defun () ()
+  "Evaluate top-level form around point and instrument."
   (check-connection)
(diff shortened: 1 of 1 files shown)


modes/lisp-mode/lisp-mode.lisp
```diff
@@ -424,16 +425,22 @@
             (interactive-eval string))))))
 
 (define-command lisp-eval-region (start end) ("r")
+  "Execute the region as List code."
```
cxxxr, line 428: It is `Lisp code`?

cxxxr (commented):
Thanks so much for the documentation!

modes/lisp-mode/lisp-mode.lisp
jcs090218: Ah, typo! Fixed it.

PR #678, opened 2023-06-03
directory-mode: sort by name and mtime

vindarel:
It works :) and it can certainly be improved (before being extended).

### DIFF
--- a/src/base/file-utils.lisp
+++ b/src/base/file-utils.lisp
@@ -65,6 +65,19 @@
     ((ppcre:scan "^~/.*" (namestring base-dir)) (probe-file% pathspec))
     (t (probe-file pathspec))))
 
+(defun sort-files (pathnames &key (key #'namestring) (test #'string<))
+  "Sort a list of pathnames."
+  (sort (copy-list pathnames)
+        test :key key))
+
+(defun sort-files-with-method (files &key (sort-method :pathname))
+  "Sort files with a sort method, one of :pathname and :mtime."
+  (cond
+    ((eql sort-method :mtime)
+     (sort-files files :test #'> :key #'file-mtime))
+    (t
+     (sort-files files))))
+
 (defun directory-files (pathspec)
   (if (uiop:directory-pathname-p pathspec)
       (list (pathname pathspec))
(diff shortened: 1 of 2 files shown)


cxxxr:
This is a very good improvement.
Thank you very much!

Does "draft" mean there will be more changes to this PR?

vindarel:
Thank you!

Yes, I put [draft] because I plan on improving the PR a little bit. I opened the PR to collect early feedback.

Note that I used a `+sbcl` feature flag, and I threw an error for other implementations… (I didn't want to pull the osicat library as a dependency).

src/base/buffer.lisp
```diff
@@ -82,7 +82,11 @@
     :accessor buffer-encoding)
    (last-write-date
     :initform nil
-    :accessor buffer-last-write-date)))
+    :accessor buffer-last-write-date)
+   (keys
+    :initform (make-hash-table)
+    :accessor buffer-keys
+    :documentation "Store internal keys and values in a hash-table. Example: store the current sorting method of a directory buffer.")))
```
cxxxr, line 89: There is already a slot called buffer-variables, can this be used?

e.g.
```
(buffer-value buffer '<key>) => <value>
(setf (buffer-value buffer '<key>) <value>) => set <value>
```

src/base/file-utils.lisp
```diff
@@ -65,21 +65,35 @@
...
+
 (defun directory-files (pathspec)
   (if (uiop:directory-pathname-p pathspec)
       (list (pathname pathspec))
       (or (mapcar (lambda (x) (virtual-probe-file x pathspec))
                   (directory pathspec))
           (list pathspec))))
 
-(defun list-directory (directory &key directory-only)
+(defun list-directory (directory &key directory-only (sort-method 'pathname))
```
cxxxr, line 88: Can we define sort-method as a type?

```
(deftype sort-method ()
  '(member :pathname :mtime))
```

src/ext/directory-mode.lisp
```diff
@@ -1,9 +1,16 @@
 (defpackage :lem/directory-mode
   (:use :cl :lem)
   #+sbcl
-  (:lock t))
+  (:lock t)
+  (:export
+   #:*default-sort-method*))
```
cxxxr, line 6: For consistency, it might be a good idea to use the :keyword

cxxxr (commented):
Thanks for the nice PR!
I have made a few comments.

vindarel:
Nice, thanks. Should be all good! 

(I didn't rebase and force-push, I'll let you rebase while merging if you want)

src/base/package.lisp
```diff
@@ -111,6 +111,7 @@
    :buffer-undo-boundary
    :buffer-value
    :buffer-unbound
+   :buffer-keys
```
cxxxr, line 114: It appears that this export remains

cxxxr (commented):
I think it's very good.
A minor point, but I commented on it.
I am not very particular about rebase.
Thank you!

src/base/file-utils.lisp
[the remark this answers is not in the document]
vindarel: It was working because it was handled by the "otherwise" clause of `cond`, but good catch.

cxxxr:
Thank you so much!

PR #679, opened 2023-06-03
Fix abbrev duplication keywords

Sasanidas:
I realize that it checks for repeat keywords in each buffer, but not when they are combined, this solves de issue :+1:


cxxxr:
I think it's very good.
Thank you so much!

PR #681, opened 2023-06-04
chore: Document some functions in window.lisp

jcs090218:
Functions that are left undocumented are functions I'm still trying to understand. But I will add it later after I understand it!


cxxxr:
Thank you so much :pray:

PR #685, opened 2023-06-05
chore: Document functions in echo.lisp

jcs090218:
This patch also fixed some indentation issues.

### DIFF
--- a/src/echo.lisp
+++ b/src/echo.lisp
@@ -6,21 +6,40 @@
 (defgeneric clear-message ())
 
 (defun log-message (string args)
+  "Print a message.
+
+This function creates the `*Messages*` buffer before it writes.
+
+The first argument is a format control string, and the rest are data to be
+formatted under control of the string."
   (when string
     (let ((msg (apply #'format nil string args))
           (buffer (make-buffer "*Messages*")))
       (with-open-stream (stream (make-buffer-output-stream
                                  (buffer-end-point buffer)))
-        (fresh-line stream)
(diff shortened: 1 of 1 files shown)


src/echo.lisp
```diff
@@ -6,21 +6,40 @@
...
+  "Print a message.
+
+The message goes into the `*Messages*` buffer and shows besides cursor.
+Return t if success.
+
+The first argument is a format control string, and the rest are data to be
+formatted under control of the string."
   (log-message string args)
   (apply #'message-without-log string args)
   t)
```
jcs090218, line 45: In Emacs, this function returns the string instead of `t`. WDYT? 🤔

cxxxr: nice catch!
Probably a remnant of the practice of returning a T on success.
Since the purpose of this function is to display a message and not to return a value, `(values)` might be more appropriate.

jcs090218: Sorry, I am still new to the development. What do you want me to change to accomplish? 🤔 Thanks!

cxxxr: You may want to change `t' to `(values)`
The `(values)` means that no values are returned

```common-lisp
```

jcs090218: Ah! Learned new knowledge! Sorry, I was being dumb! Thank you!

Just updated the code! 🚀

cxxxr: Thank you so much!
It is very helpful that you have written the documentation.

PR #686, opened 2023-06-05
Lem with other CL implementations

Sasanidas:
So, this is the first try of making Lem compile/work with other implementations, I manage to make it compile the ncurses front-end with abcl,:
abcl

(It should of course also work with SBCL)

Compilation progress (make the :lem system compile)
- [x] ABCL
- [x] CCL
- [x] ECL
- [x] Clasp 

*Most* tests green
- [x] ABCL
- [x] CCL
- [x] ECL
- [x] Clasp 

Front-end ncurses progress (run the system)
- [x] ABCL
- [x] CCL
- [x] ECL
- [x] Clasp 

"Usable state" (it can be used like sbcl)
- [x] ABCL
- [x] CCL
- [x] ECL
- [x] Clasp


Sasanidas:
I had better luck with CCL, which works quite better, except for some errors here and there, it's usable :+1:

Sasanidas:
> So, this is the first try of making Lem compile/work with other implementations, I manage to make it compile the ncurses front-end with abcl,: abcl
> 
> Buut sadly, when launching (lem:lem), it just shows the tmp buffer and doesn't respond to any command: abcl_lem
> 
> The good news, is that it also is working with sbcl, but of course more testing is needed +1


Same thing with ECL, it compiles but it doesn't register any input

Sasanidas:
So, loading the ncurses front-end, I realize that ABCL and ECL have the same problem, no response from the user input.

I debug the *editor-event-queue* and seems like the events are indeed been receive (it shouldn't be a problem in the change of timer because SBCL-Lem does work)

Maybe a library that I missed :question:

cxxxr:
That's wonderful, thank you!

ECL should have worked before, maybe there is a problem around threads
I'll dig a little deeper tonight (at JST) and see what I can find.

Sasanidas:
Yeah, I had some problems with ECL, specially the annoying ASDF version error (not related to Lem), so I think I'm going to focus on ABCL, I think it's the one with more potential uses (open the Java ecosystem)

Sasanidas:
I did a couple of things in the last commit https://github.com/lem-project/lem/pull/686/commits/c047386225350243c48a328c1f737cc9f48b97cd :

- I adapt the library mailbox from SBCL to a CL ANSI version, all the tests seems to pass but still there is an ignore errors that I'm not sure how to fix (I'll add a TODO in it)
- Modify again timer to reflect these changes
- Also I add all the tests from the mailbox SBCL to this ANSI version to the rove framework, and all green for now.

I still didn't manage to make it work with ABCL tho, but I think I'm in the right direction here :+1:

Sasanidas:
Maybe the mailbox library can be separated in a separate repository?
I couldn't found anything similar online (there is mailbox, but it's quite limited compare to this version), and it can be used for other people/projects

cxxxr:
> Maybe the mailbox library can be separated in a separate repository?

It's might be good.

Sasanidas:
OK, I think I found THE problem:


Selection_001

So, the thing is when calling `collecting-subclasses`, in signal condition in signal handler, it breaks because the subclass cannot be coerce into a simple condition.

So, why it works in SBCL? Because is seems like it's doing some wrapping around the classes, let's see the call to the function (this functions returns a list, in this case, I'm inspecting the first element):

```
(collect-subclasses 'lem-core::before-executing-command :include-itself nil)

```
SBCL:
Selection_002

Clasp:
Selection_003

In ABCL happens the same as in Clasp.

So, in the first case, it can be signal without any problem, but in the second it cannot be coerce. I guess the odd (or clever) behaviour is in SBCL.

Not sure what to do here, so calling for some help

cxxxr:
Hmm, this seems to be a deep-seated problem.
The signal-holder mechanism is unnecessarily complicated and may be better replaced by a hook.

cxxxr:
@Sasanidas 
https://github.com/lem-project/lem/pull/716
I think I've merged the PR and this problem has been resolved, what do you think?

Sasanidas:
> @Sasanidas #716 I think I've merged the PR and this problem has been resolved, what do you think?

Oh! seems very interesting, I'll merge the branch and try it :+1:

Sasanidas:
Clasp seems lo work editing buffers and opening commands and durectory-mode, but it fails when open a file:
error_001

Sasanidas:
rove is giving wrongly triggering some tests as fault, showing this error with most of them:
```
   Raise an error while testing.
   STREAM-ERROR: java.nio.channels.ClosedByInterruptException

     0: (#<FUNCTION #<FUNCTION {42161ECE}> {42161ECE}> java.nio.channels.ClosedByInterruptException)
     1: (SIGNAL java.nio.channels.ClosedByInterruptException)
     2: (:CLASS org.armedbear.lisp.Lisp METHOD error FILE Lisp.java LINE 383)
     3: (:CLASS org.armedbear.lisp.FileStream$1 METHOD execute FILE FileStream.java LINE 333)
     4: (:CLASS org.armedbear.lisp.Symbol METHOD execute FILE Symbol.java LINE 854)
     5: (:CLASS org.armedbear.lisp.LispThread METHOD execute FILE LispThread.java LINE 992)
     6: (:CLASS org.armedbear.lisp.open_3 METHOD execute FILE open.lisp LINE 101)
     7: (:CLASS org.armedbear.lisp.CompiledClosure METHOD execute FILE CompiledClosure.java LINE 98)
     8: (:CLASS org.armedbear.lisp.Symbol METHOD execute FILE Symbol.java LINE 805)
     9: (:CLASS org.armedbear.lisp.LispThread METHOD execute FILE LispThread.java LINE 904)
     10: (:CLASS org.armedbear.lisp.local_projects_tmpTEZHAM6I_10 METHOD execute FILE local-projects.lisp LINE 90)
     11: (:CLASS org.armedbear.lisp.Symbol METHOD execute FILE Symbol.java LINE 816)
     12: (:CLASS org.armedbear.lisp.LispThread METHOD execute FILE LispThread.java LINE 924)
     13: (:CLASS org.armedbear.lisp.local_projects_tmpTEZHAM6I_11 METHOD execute FILE local-projects.lisp LINE 102)
     14: (:CLASS org.armedbear.lisp.Symbol METHOD execute FILE Symbol.java LINE 805)

```

Sasanidas:
Good news! I managed to make ABCL work very decently with Lem, there are a couple of problems here and there, but unlike CCL, it doesn't trigger any obvious visual bug :+1:

Sasanidas:
It's not related that much with this issue, but I think making Lem work with ABCL and with Clojure maybe interesting :+1:

cxxxr:
Umm, I wonder if the swank protocol is the cause.

Sasanidas:
> Umm, I wonder if the swank protocol is the cause.

Umm, it maybe, I updated the swank backend for clasp. but it trigger same error. I try the function of decode and encode integer and seem to work fine. Also, I did check the babel library for some incompatibilities but everything seems to work :thinking:

Sasanidas:
I decided to not address the support for CLISP and CMUCL in this pull request for the following reasons:

- CLISP: It cannot compile some of the dependencies, and that's maybe too much effort for an implementation is not that in development. Maybe in the future can be interesting to test it again, for now, it's not my priority.

- CUMCL: Even tho this distribution is well develop, I think it's too close to SBCL to be worth the effort for compile Lem.

For the other implementation I think each can have some interesting benefits of the project:

- ABCL: Interaction with the JVM ecosystem (this include our long distance cousin Clojure)

- Clasp: LLVM performance and C++ interoperability

- ECL: Small executables and easy interaction with C

- CCL: It's a nice alternative to SBCL, it also compile natively and really fast.

So, changing a little bit of the goal of this issue, I will focus in makes these last 4 work similar on how SBCL works. For some small bugs (visual glitch or some specific secondary packages not working I think it makes more sense to add a new pull request).

I also intend to change the implementation page in the lem-page to reflect the status of each implementation (this time I won't do it manually, I'll write a package to get that information in a nice table or something similar). So people is more aware on what to expect with each. 

Let me know what do you think @cxxxr :+1:

cxxxr:
Yes, I agree that it is a good idea to support those processors.
I think your idea of putting together a page with the status of the processing implementations support is also very good.

Sasanidas:
> That's wonderful, thank you!
> 
> ECL should have worked before, maybe there is a problem around threads I'll dig a little deeper tonight (at JST) and see what I can find.

Seems like ECL is working really good now! maybe a little slow in the ncurses front-end (it may require some optimizations)

Sasanidas:
Even tho ECL works really well (maybe the best aside from SBCL) I wasn't been able to dump the image to an executable.

Running https://ecl.common-lisp.dev/static/manual/System-building.html#Compiling-with-ASDF does creates an executable, but it  the stack crashes when I try to run it.

Sasanidas:
Using (https://github.com/lem-project/lem/pull/756) to document different implementations, some  of the results are now in https://lem-project.github.io/lem-page/development/implementations-details/

Sasanidas:
OK, seems like the CCL display bug was because in the micros backend, the function-name implementation was returning a symbol instead of a string, to ensure that this also works with other implementations, I format the function name to a string :+1:

Sasanidas:
At this point, CCL works as well as SBCL (except for lsp, but that's because of the sb-concurrency library usage, which I think it's a problem for other issue :+1: )

Sasanidas:
I think what is left in this issue is to test both ECL and Clasp, upload the results and create a few specific issues for implementation (here we can also use github labels), let me know what do you think

cxxxr:
Yes, it might be a good idea to merge at a reasonable point for now, and separate the pull requests.

Sasanidas:
OK, so I have to tweak a little bit the script to make it work with Clasp, but I think this issue is ready to merge :+1:

src/system.lisp
```diff
@@ -5,7 +5,7 @@
 (defun lem-relative-pathname (pathname)
   (when *deployed*
     (truename (merge-pathnames pathname
-                               (uiop:pathname-directory-pathname (first sb-ext:*posix-argv*))))))
+                               (uiop:pathname-directory-pathname "/home")))))
```
cxxxr, line 8: Isn't this environment-dependent?

src/version.lisp
```diff
@@ -7,9 +7,13 @@
       (uiop:with-current-directory (path)
         (string-trim
          (list #\Newline #\Space)
+	 #+sbcl
          (with-output-to-string (stream)
            (uiop:run-program "git rev-parse --short HEAD"
-                             :output stream)))))))
+                             :output stream))
```
cxxxr, line 13: Isn't the run-program portable?

src/common/timer.lisp
```diff
@@ -109,41 +111,42 @@
...
+        (bt:make-lock "timer internal mutex"))
   (start-timer-thread timer ms repeat-p)
   timer)
 
 (defmethod stop-timer ((timer timer))
   (stop-timer-thread timer))
 
 (defun start-timer-thread (timer ms repeat-p)
-  (let ((stop-mailbox (sb-concurrency:make-mailbox))
+  (let ((stop-mailbox (lem-mailbox::make-mailbox))
```
cxxxr, line 149: Don't you export symbols for mailboxes?

cxxxr (commented):
This is a very attractive change!
I am very thankful.

I just need a few more corrections :pray:

src/common/timer.lisp
cxxxr: This is a trivial issue, so a fix may not be mandatory for this PR

Sasanidas:
Let me know if there's anything more to change :+1: (@cxxxr )

cxxxr:
Thank you very much.
I think it's very good.
I appreciate your efforts over the past month.

Issue #688, opened 2023-06-05
A Roadmap for Lisp Mode Evolution 🚀 


cxxxr:
## Problem

With every version update of SLIME, incompatible changes are being made.
Up to version v2.27, it worked fine with Lem, but there seems to be quite a drastic change from around v2.28, and the behavior has changed.
For example, a significant disruptive change in the past can be found at https://github.com/slime/slime/commit/78ad57b7455be3f34a38da456183ddf8d604bdf8.
If not addressed on the client-side, the REPL will hang up.
Also, the output to the REPL has become a bottleneck, and asdf:load-system has become considerably slow.
Besides, there are other areas where behavior has changed due to minor changes in behavior.

As a policy of SLIME, it is strongly integrated with Emacs, and it seems that it does not consider other implementations much.
Continually keeping up with updates to SLIME is too costly and unrealistic, but if left unattended, it will stop working, so constant response is required.

### Fork SLIME

A conceivable solution is to fork SLIME.

#### Pros
By choosing this method, there is the advantage of not being affected by incompatible changes to SLIME.
In addition, there is already a SLIME implementation in Lem, and there is a track record of comfortable use so far, so it is easy to gradually improve.

#### Cons
Rarely, the implementation of swank may not work due to the upgrade of the sbcl version.
SLIME may make this correction, but Lem needs to deal with it without receiving the benefits.

### Moving to SLY
There is another project called SLY besides SLIME.
This is a branch from SLIME.
However, since SLY also has the potential to make incompatible changes, it is likely not to be a solution.

### Reimplement with each version upgrade
There is a method to recreate the implementation of SLIME in Lem each time the version of SLIME goes up and becomes incompatible.
This is too costly to consider.

## ToDo
I will summarize the transition procedure below.
This is one idea, so there may be a need to discuss if there are better ways.

## Phase 1 ✅ 
First, fork SLIME v2.27 and place it as a submodule in the Lem repository.

## Phase 2 ✅ 
Then, to avoid a name conflict with the external SLIME of Lem, change the name of the swank package.
Since the same was done in a repository called micros before, it will be named micros (please let me know if you have a better name).
At the same time, the swank client side of Lem also needs to change its name from swank to micros.

The downside of doing this is that swank can no longer be used in Lem.

## Phase 3
From here, I will continue developing micros.
I will incorporate the features of SLIME and SLY that I think are better, refer to the functions of symbolics genera, and make it better.

## Phase 4
It might parallel Phase 3.
[...]

PR #691, opened 2023-06-06
Replace swank with micros

cxxxr:
## Overview
This change corresponds to `phase 1` and `phase 2` of https://github.com/lem-project/lem/issues/688

I have replaced the part where I was using the swank package in lisp-mode with [micros](https://github.com/lem-project/micros/).
In addition, I had bundled micros, but have now switched it to a submodule.

## Breaking change
Due to this change, it is no longer possible to connect to Swank.
Instead, we will use micros, and there are currently no differences in the API.
```
;; before
(swank:create-server :dont-close t :port <port-number>)

;; after
(micros:create-server :dont-close t :port <port-number>)
```

In addition, since I have added micros as a submodule, updating the submodule is necessary.
```
$ git submodule update --recursive
```

### DIFF
--- a/modes/lisp-mode/swank-protocol.lisp
+++ b/modes/lisp-mode/swank-protocol.lisp
@@ -41,7 +41,7 @@
 
 (defmacro with-swank-syntax (() &body body)
   `(with-standard-io-syntax
-     (let ((*package* (find-package :swank-io-package))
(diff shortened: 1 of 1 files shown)


vindarel:
Note for devs who had cloned Lem before, I needed:

    git submodule update --init submodules/micros

thx

PR #696, opened 2023-06-08
find files recursively

vindarel:
This makes using Lem more efficient :sunglasses: 

Selection_143

The new command should work for all platforms. 
We have a choice of programs to find files: 
  - [fd](https://github.com/sharkdp/fd), which is`fdfind` in Debian, excludes .git, node_modules and the like by default
  - find
  - a lisp way

This is NOT like "find file in project", which could be done later on. This is only "find a file with a choice of all files in subdirectories, starting from this buffer's directory".

I settled on CLOS `eql` specializers to dispatch by the chosen finder program, so the user has built-in configuration options (around methods, etc).

subject to review:

- I introduced my STR library as a dependency. It is lightweight and, albeit not strictly necessary for this PR, I believe it is very handy (and popular!).
- on `#+unix` (I believe it includes Mac OS?) I call `which` to check that a program is present in the system. On non-unix, we fallback to the lispy method.
- a couple comments inline.

### DIFF
--- a/lem.asd
+++ b/lem.asd
@@ -10,6 +10,7 @@
                "yason"
                "log4cl"
                "split-sequence"
+               "str"
                "dexador"
                "lem-base"
                "lem-encodings"
--- a/src/base/buffer.lisp
+++ b/src/base/buffer.lisp
@@ -19,7 +19,8 @@
     :accessor buffer-%filename)
    (%directory
     :initform nil
-    :accessor buffer-%directory)
+    :accessor buffer-%directory
+    :documentation "The buffer's directory. See: `buffer-directory'.")
    (%modified-p
     :initform 0
     :reader buffer-modified-tick
--- a/src/commands/file.lisp
+++ b/src/commands/file.lisp
@@ -147,12 +147,16 @@
   (str:lines
    (uiop:run-program (list "find" ".") :output :string)))
 
+(defun %shorten-path (cwd path)
+  (str:replace-all cwd "" path))
+
 (defmethod get-files-recursively ((finder (eql :lisp)))
   "Find all files recursively, without external tools."
   ;; XXX: this method returns full paths, instead of paths starting at the current directory.
-  (let ((results))
+  (let ((results)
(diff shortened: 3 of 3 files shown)


cxxxr (commented):
Ooh, this is very useful!
It seems like a good idea to make it the default function of `find-file`.

I made some comments, thank you in advance.

Sasanidas:
> Ooh, this is very useful! It seems like a good idea to make it the default function of `find-file`.
> 
> I made some comments, thank you in advance.

Umm, maybe an overkill to make the default find-file version?

vindarel:
>  It seems like a good idea to make it the default function of find-file.

It might be a nice(r) default, but not in this current state:

- I am not using `prompt-file` but prompt for string, so we don't have the current directory on the prompt, so we can't go up one directory :s 
  - (any ideas?)
- you can't run this on the HOME or on a directory with too many files :s 
  - need to find workarounds, or be able to kill the underlying process, etc.

For the default, we might compare what popular editors do? Personally, I think such a feature is super handy (according it doesn't come with obvious issues, like now).

cxxxr:
Hmmm, I see.
Maybe it's similar to Command+P (Control+P) in VSCode.

cxxxr:
@vindarel Can I merge this PR already?

vindarel:
wait wait! I will improve it with your reviews.

vindarel:
```lisp
(define-command find-file-recursively (arg) ("p")
  ;; what is arg for?
```

Shall I use the arg, and how? I'm investigating this.

cxxxr:
Oh, sorry, I'm not in a hurry.
Thank you.

cxxxr:
> Shall I use the arg, and how? I'm investigating this.

Maybe we can add it when we need it.

vindarel:
It's better (maybe good?) now!

- code review
- if the process of finding finds recursively takes too long (5s by default), kill the thread and the process, fallback to find-file
- fixed the lispy way to find files 

It works for me©

cxxxr:
It is very cool!
Thank you so much 👍

PR #699, opened 2023-06-08
Add thingatp functionality

Sasanidas:
A very Emacsy approach to get the "thing at point", the idea is to extend it with more types :+1:

### DIFF
--- a/src/ext/language-mode.lisp
+++ b/src/ext/language-mode.lisp
@@ -9,6 +9,7 @@
    :line-comment
    :insertion-line-comment
    :find-definitions-function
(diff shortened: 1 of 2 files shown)


cxxxr (commented):
I think it is a very nice feature!
I've made few comments :pray:

Sasanidas:
Oh my, so many small mistakes, I guess that's happens when I copy and paste :facepalm:

Sasanidas:
I think I solve most of them :raised_hands:

cxxxr:
Thank you so much!

PR #700, opened 2023-06-09
[VI-MODE] • Enhance define-vi-state to allow expose more of the defclass CLOS functionality.

theangelperalta:
The plan is to use inheritance to reduce code redundancy and improve the extensibility of vi-states, such as `visual` state. Currently, `visual` state has three sub states (`visual-char`, `visual-line`, `visual-block`), which produce code redundancy with state changes.

TODO: Follow-up with PR to refactor redundancy `visual` state.

### DIFF
--- a/modes/vi-mode/core.lisp
+++ b/modes/vi-mode/core.lisp
@@ -6,7 +6,6 @@
   (:export :*enable-hook*
            :*disable-hook*
            :vi-state
-           :state-message
            :vi-mode
            :define-vi-state
            :current-state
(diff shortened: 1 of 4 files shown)


theangelperalta:
@cxxxr, this PR is ready for reviewing.

cxxxr:
@theangelperalta 
It's very good.
There is a behavior change in normal-mode and insert-mode where the cursor color is the same in both modes, is this intended?

theangelperalta:
> @theangelperalta It's very good. There is a behavior change in normal-mode and insert-mode where the cursor color is the same in both modes, is this intended?

Nice catch @cxxxr, it seems like I addressed the unintended behavioral change.

cxxxr:
Thank you so much!
I think it is very good.

PR #701, opened 2023-06-09
Add some tests to thingatp and hyperspec functionality

Sasanidas:


modes/lisp-mode/hyperspec.lisp
```diff
@@ -0,0 +1,1006 @@
+(in-package :lem-lisp-mode/internal)
```
cxxxr, line 1: This might be a good idea to separate the packages. e.g. `lem-lisp-mode/hyperspec`

cxxxr (commented):
Finally, hyperspec can be referenced! Thanks!

cxxxr:
There are areas where the test has failed 👀

Sasanidas:
> There are areas where the CI has failed eyes

Yeah, seems like it's the egg chicken problem, I'm using a command in thingatp and the tests load thingatp before the extensions where the command is :face_with_spiral_eyes:

cxxxr:
> Yeah, seems like it's the egg chicken problem, I'm using a command in thingatp and the tests load thingatp before the extensions where the command is 😵‍💫

What if it load it after language-mode.lisp in lem.asd?

Sasanidas:
> > Yeah, seems like it's the egg chicken problem, I'm using a command in thingatp and the tests load thingatp before the extensions where the command is face_with_spiral_eyes
> 
> What if it load it after language-mode.lisp in lem.asd?

Right! make sense :+1:

Sasanidas:
Now are the tests that are failing because of the symbol

cxxxr:
Oh, find-definitions is a symbol of the lem/language-mode package, not lem package.

cxxxr:
Thank you so much!

PR #705, opened 2023-06-09
help: add docstrings

vindarel:
so that they show up in documentation-describe-bindings and in the generated .md file.


vindarel:
I added a `help` command. The goal is really to have something called "help" under M-x, so newcomers have a starting point. I know I have searched for it.

It probably can be improved with the new documentation mode, and clickable links.

jcs090218:
> I added a help command. The goal is really to have something called "help" under M-x, so newcomers have a starting point. I know I have searched for it.

Wow, nice! 🚀

cxxxr:
Very good, Thank you!

PR #710, opened 2023-06-10
Update README.md

Sasanidas:
Fix contributors link


Sasanidas:
Or even better, use contrib.rocks (the open collective image doesn't  seem to be updated)

PR #711, opened 2023-06-10
Update README.md

Sasanidas:
Change contributors image with the one in contrib.rocks


cxxxr:
Umm, What is the intention behind this?

Sasanidas:
Seems like the contributors graphs is not updated (doesn't show new contributors) and also point out to an url that doesn't exist.

cxxxr:
Oh, I see.
nice catch! I think it's good!

PR #722, opened 2023-06-11
Add repository patch as a solution for dependency hell

Sasanidas:
So, using quick-patch as a submodule, we can use it to have certain versions of repositories loaded when compiling Lem.

For now it's only used in the patch-build script, and  it can be used the same as the build.lisp:
```
sbcl --load "patch-build.lisp"
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ GIT_IGNORE.*
 lem
 lem-ncurses
 
+build/*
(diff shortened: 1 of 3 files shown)


Sasanidas:
It would be also interesting to load the code not only in the patch-build, but also in the init file before doing any lem loading

cxxxr:
It might be good.
Let's go with this for now.

Issue #720, opened 2023-06-11
vi-mode delete region and yank region doesn't seem to work


Sasanidas:
For some reason in the main branch (in the download version of 2.0 works) the region interaction in vi-mode doesn't seem to work properly (both delete and copy region)

Sasanidas:
Solved! https://github.com/lem-project/lem/pull/724

PR #724, opened 2023-06-11
Fix visual state problem

Sasanidas:
Seems like there is no place in the code where the state is "visual", it's just one of the visual possibilities.

I think this fixes the issue https://github.com/lem-project/lem/issues/720


cxxxr:
Thank you!

PR #725, opened 2023-06-11
hyperspec: message the user if symbol not found

vindarel:
If the user right clicks on, say, "save-lisp-and-die", nothing would happen, because this symbol isn't in the index.
Display a message.


cxxxr:
Thank you!

PR #726, opened 2023-06-12
Vindarel/meta is not right alt

vindarel:
My issue was: I was not able to enter any character of my keyboard accessible with the right Alt ("AltGr" on my keyboard, MSI laptop).

Now I can. I hope to not break things! (I read that AltGr is Option (⌥) on Mac)

Now I can type an underscore (among many other useful keys < > [ ]  % ), that I was missing for vi-mode.


cxxxr:
Thanks! maybe a good idea since I don't use RAlt.

PR #730, opened 2023-06-12
new command project-find-file

vindarel:
Similar to find-file-recursively, but finds the project root first.

If no project root was found, use find-file-recursively from the current working directory.

---

It works for me on Linux. I think real tests on other platforms is required, for this reason I marked the PR as draft. Maybe the test suite can help…

### DIFF
--- a/src/commands/project.lisp
+++ b/src/commands/project.lisp
@@ -1,0 +1,127 @@
+(defpackage :lem-core/commands/project
+  (:use :cl :lem-core :lem-core/commands/file)
+  (:export :*root-directories*
+           :*root-files*
+           :root-p
+           :find-root
+           :project-find-file)
+  (:documentation "Defines utilities to find a project root directory and the command project-find-file."))
+
+(in-package :lem-core/commands/project)
+
+(defvar *root-directories*
+  (list
+   ".git"
+   ".hg"
+   "_FOSSIL_"
+   ".bzr"
+   "_darcs"
+   )
+  "A list of directory names to be seen as project roots.
+  Sans trailing slash.")
+
+(defvar *root-files*
+  (list
+   ".project"
+   ".projectile"
+   ".fslckout"
+   "TAGS"
+   "configure.ac"
+   "configure.in"
+   ".ignore"
+   "Makefile"
+   )
+  "A list of files to be seen as project roots.")
+
+(defun parent-directory (pathname)
+  "Return the parent directory (pathname)."
+  (uiop:pathname-parent-directory-pathname pathname))
+
+(defun add-directory-separator (pathname)
+  "Add a directory separator, according to the host: a slash on Linux, something else on other popular platforms."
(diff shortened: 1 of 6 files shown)


vindarel:
I added `project-root` to display the root to the user, `project-root-directory` to open the directory, and siblings `-other-window` commands.

vindarel:
I added `M-x project-delete-buffers`.

src/commands/project.lisp
```diff
@@ -0,0 +1,231 @@
...
+         (project-root (find-root cwd))
+         (root (or project-root cwd)))
+    (find-file root)))
+
+

+
+(defun %buffer-list ()
+  ;; Defined after this file in the .asd, need to defer the call.
+  (uiop:symbol-call :lem/list-buffers 'buffer-list))
```
cxxxr, line 169: I think this is fine simply with `buffer-list`.
This is because buffer-list exists in lem-base.

src/commands/project.lisp
```diff
@@ -0,0 +1,231 @@
...
+          (uiop:directory-files pathname)))
+
+(defun list-subdirectories (pathname)
+  "Return a list of subdirectories (pathnames)."
+  (uiop:subdirectories pathname))
+
+(defun root-p (pathname)
+  "Return t if this directory contains any of the stop files or directories."
+  (or (found-subdirectory-root-p pathname)
+      (found-file-root-p pathname)))
```
cxxxr, line 76: This function is not used from anywhere, is it intended?

src/commands/project.lisp
```diff
@@ -0,0 +1,231 @@
...
+  recurse: if non T, do not recurse in parent directories (testing only)."
+  ;; recursing: when calling find-root on the HOME directory, it's OK to return it as the root.
+  ;; Otherwise, return nil (to get the caller directory).
+  (check-type pathname (or pathname string))
+  (setf pathname (pathname pathname))
+  (cond
+    ;; Stop at /
+    ((or (equal (directory-namestring pathname)
+                "/")
+         ;; XXX: windows??
```
cxxxr, line 112: It was "/" even in windows.

cxxxr (commented):
It is a nice feature!

I have a few comments.
Thank you.

src/commands/project.lisp
vindarel: Oh you are right again. I forgot what lead me to this.

src/ext/filer.lisp
```diff
@@ -173,3 +174,9 @@
 
 (define-command filer-select () ()
   (select (back-to-indentation (current-point))))
+
+(define-command filer-project-root () ()
```
cxxxr, line 178: Would it be okay to overwrite the `M-x filer` command with this command?

vindarel: good point, agree for the better default.

cxxxr:
Thank you!.

PR #736, opened 2023-06-14
Add lisp-eval-print-last-sexp

Sasanidas:


cxxxr:
Isn't `lisp-eval-at-point` the equivalent function?
Sorry for being so confusing.

Sasanidas:
It's similar, but this one insert the result in the buffer

cxxxr:
Hmm, I see.
This is very helpful, thank you.

How about a command name, for example `lisp-eval-last-expression-and-insert`?

cxxxr:
I also intend to extend lisp-eval-at-point in the future,
and are considering selecting an evaluation result to launch the inspector,
or in case of an error, to launch the debugger at any time by clicking on the displayed condition.

Ideally, it would be possible to integrate the functionality of that with the insertion into the buffer, but this is difficult at present.

cxxxr:
Thank you!

PR #739, opened 2023-06-15
Change cl-charms to croatoan

Sasanidas:
So, this is the attempt to change cl-charms which is not that much develop, to a CLOS version of it (mention here https://github.com/lem-project/lem/pull/686#issuecomment-1587226396).

I will slowly but surely at first have both working at the same time, and I'll change the functions from cl-charms to croatoan and at the end I'll remove cl-charms (except for win32 build maybe).

### DIFF
--- a/frontends/ncurses/lem-ncurses.asd
+++ b/frontends/ncurses/lem-ncurses.asd
@@ -1,6 +1,7 @@
 (defsystem "lem-ncurses"
   :depends-on ("cffi"
-               "cl-charms"
+	       "croatoan"
+	       "cl-charms"
                "cl-setlocale"
                "lem"
                "lem/extensions")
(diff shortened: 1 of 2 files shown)


Sasanidas:
Now that I changed the main object, I cannot try the ncurses (it will break with the others methods that are assuming that the ncurses windows is from cl-charms).

So now, my worst enemy is going to be a hidden typo! :thought_balloon:

Sasanidas:
Ok, so after doing some testing, sadly even tho croatoan give us a really nice abstraction, it also introduce a more complex (and potentially slow) way of doing things. Also realize, that half of the ncurses calls that the front-end is using are not wrapped by croatoan.

For SBCL this maybe good enough, but I think for a terminal front-end speed should be a priority, so I'm closing this pull request so we don't have performance issues in the future with other implementations.

cxxxr:
Hmm, I see.
Thank you for trying it out.

PR #741, opened 2023-06-15
project: add keybindings (C-x p f)

vindarel:
I propose to add default keybindings.

C-x p f is modeled after Emacs' project.el, but

C-x p d (open directory) and C-x p K (upper case k) differ. There is no confirmation prompt to delete the project buffers, so I put it further away with the uppercase letter.

That being said… maybe it should with a "D", like "Delete", or rename the command to `project-kill-buffers`, like `kill-buffer`?

Emacs' C-x p d is project-find-dir, which we can do in Lem with project-find-file (it includes subdirectories).

### DIFF
--- a/src/commands/project.lisp
+++ b/src/commands/project.lisp
@@ -14,6 +14,10 @@
 
 (in-package :lem-core/commands/project)
 
+(define-key *global-keymap* "C-x p f" 'project-find-file)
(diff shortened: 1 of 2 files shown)


cxxxr:
I think it's good.

> That being said… maybe it should with a "D", like "Delete", or rename the command to project-kill-buffers, like kill-buffer?

Yes indeed, you could use the word `kill` for consistency.

Issue #743, opened 2023-06-16
A simple makefile


Sasanidas:
Instead of a autoconf, just a simple Makefile with basic commands like compile or update submodules or things like that.

If you think this makes sense, I'll add one :+1:

cxxxr:
Yes indeed!
it would be nice to have a build.lisp wrapper, etc.

PR #744, opened 2023-06-16
Add makefile and build sdl2

Sasanidas:
This solves https://github.com/lem-project/lem/issues/743 :+1:


Makefile
```diff
@@ -0,0 +1,10 @@
+LISP ?= ${shell which sbcl}
+
+build-ncurses:
+	$(LISP) --load scripts/patch-build-ncurses.lisp	
+
+build-sdl2:
+	$(LISP) --load scripts/patch-build-sdl2.lisp
```
cxxxr, line 7: Is it necessary to run `git submodule update --init --recursive`?

cxxxr: Ah, we can make it a separate rule.

cxxxr:
Thanks!

PR #747, opened 2023-06-16
vindarel/directory-sort-by-size

vindarel:
I added a third sort method on the "s" key, sorting by file size.

It's handy, but there might be a usability concern: pressing "s" twice doesn't get us back to normal and there is no visual clue about the current sort order (only the message that disappears).

One possibility could be sorting by size with another key like "S" (capital s), but it won't be discoverable by Emacs users (and by the way, sorting by size is a surprisingly difficult feature in Emacs' Dired IMO, which is to be done with a third-party package).. Also, "S" is to create a symlink in Dired. I probably should investigate a visual clue.

### DIFF
--- a/src/base/file-utils.lisp
+++ b/src/base/file-utils.lisp
@@ -75,6 +75,8 @@
   (cond
     ((eql sort-method :mtime)
      (sort-files files :test #'> :key #'file-mtime))
+    ((eql sort-method :size)
+     (sort-files files :test #'> :key #'file-size))
     (t
      (sort-files files))))
 
--- a/src/base/file-utils.lisp
+++ b/src/base/file-utils.lisp
@@ -96,6 +96,8 @@
                                                     :sort-method sort-method))))))
 
 (defun file-size (pathname)
+  #+sbcl
+  (sb-posix:stat-size (sb-posix:stat pathname))
   #+lispworks
   (system:file-size pathname)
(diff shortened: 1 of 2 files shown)


cxxxr:
How about a context menu to change the sort type?
For example, pressing a key for sorting would bring up a menu.

vindarel:
> context menu

It could be nice (but not on "s").

Indeed, it could be really nice, because we could super-charge it with sort options: atime, ctime, user, uid, group, extension…

> Or you might want to make the S key switch between mtime, size, and pathname in that order.

oops sorry, I don't understand the goal.
(do like "s" but in reverse order of sort methods?)

>  Maybe separate PRs for sort switching, what do you think?

ok, so I can try other features another day.

BTW, this compatibility library could be a candidate for inclusion: https://github.com/Shinmera/file-attributes

cxxxr:
> oops sorry, I don't understand the goal.

The "s" is used to switch between size and mtime, but I suggested that it could be switched to pathname as well.

At any rate, I think this PR is very good and I will merge it.
Thank you!

vindarel:
> The "s" is used to switch between size and mtime, but I suggested that it could be switched to pathname as well.

Currently, a directory buffer is opened and sorted by pathname. "s" then sorts by mtime, then size, then back to pathname.

PR #756, opened 2023-06-18
Extend documentation mode to document tests

Sasanidas:
The idea is to generate a markdown file with a table of each tests, this is intended to use not only with SBCL but with more implementations.


Sasanidas:
The CI error seems a little bit odd, maybe roswell is using a different rove version from quicklisp?
```
;   READ error during COMPILE-FILE:
;   
;     Symbol "TEST-PASSED-ASSERTIONS" not found in the ROVE package.
;   
;       Line: 51, Column: 33, File-Position: 1529
;   
;       Stream: #<SB-INT:FORM-TRACKING-STREAM for "file /home/runner/work/lem/lem/modes/documentation-mode/documentation-tests.lisp" {100118B893}>
Unhandled UIOP/LISP-BUILD:COMPILE-FILE-ERROR in thread #<SB-THREAD:THREAD tid=3413 "main thread" RUNNING
                                                          {1001288073}>:
  COMPILE-FILE-ERROR while
  compiling #<CL-SOURCE-FILE "lem-documentation-mode" "documentation-tests">

```

Sasanidas:
Any idea on how to solve this CI issue? Maybe don't include rove as dependency? (ping @cxxxr

cxxxr:
https://github.com/fukamachi/rove/commit/82be18ba84de173ce1032745c8e4313fd5c59c3c
Isn't the symbol test-passed-assertions lost in this commit?

cxxxr:
I think this feature is very great.

Sasanidas:
I changed the rove calls for the new version on quicklisp and it seems to work correctly :+1:

modes/documentation-mode/internal.lisp
```diff
@@ -171,16 +175,18 @@
                 (insert-string point "-|")
                 (insert-character point #\newline)))))
 
-(defun generate-markdown-file (filename)
+(defgeneric generate-markdown-file (filename type))
```
cxxxr, line 178: The method dispatch is very attractive and I love it.
But here I felt it would be better to simply separate the functions
What do you think?

modes/documentation-mode/internal.lisp
```diff
@@ -64,6 +67,7 @@
             (enough-namestring (location-file location) (asdf:system-source-directory :lem))
             (location-line-number location))))
 
+#+sbcl
```
cxxxr, line 70: Since separation by `*features*` is complicated, I think it is preferable to keep it to a minimum.
except sbcl, what about command-definition-location returning meaningless values (e.g., always set position to 0)?

cxxxr (commented):
I made a few comments, but overall I think it is very good.

Minor comments: Indentation includes tabs, but I prefer to use all spaces.
Is it because it is the default setting in emacs?

modes/documentation-mode/internal.lisp
Sasanidas: Ummm, I was thinking to use this functionality to generate other kind of markdown, like with maybe benchmark, so I think it can be reused.

But I tend to overuse generics, so if you like I can change it back to a function :+1:

modes/documentation-mode/internal.lisp
Sasanidas: You mean something like this https://github.com/lem-project/lem/pull/756/commits/52a90ebb3c0cd6e99330a6a6c116113e9b2d1789?

cxxxr: Yes, I thought it would be better to consolidate on the called side rather than the caller.
How about setting the file name to nil to indicate that it has no meaning?

modes/documentation-mode/internal.lisp
cxxxr: I was only thinking about statically determining which method the caller is.
I thought a function would be better in that case.
However, if we need extensibility in the future, it may be better to make it a method.

modes/documentation-mode/internal.lisp
Sasanidas: Done :+1:

modes/documentation-mode/internal.lisp
Sasanidas: >  However, if we need extensibility in the future, it may be better to make it a method.

Ummm, I do think it may be extensible in the future indeed, in any case I'll change it if you want :+1:

cxxxr: Well, this might be fine as it is too.

cxxxr:
Thank you so much!

PR #758, opened 2023-06-18
projects: save and switch projects

vindarel:
Continuing the projects feature:

- introduces a command to save the current project in the history, on disk
 -  => (this could be improved further with automatic project discovery later on)
- then project-switch-project prompts to choose a project, then calls project-find-file.

Keybindings: `C-x p p` is similar to Emacs' project.el, the other two on "s" and "u" are not.

In the future, it could be nice to have a setting to choose if we want automatic collecting of projects in the history, each time we visite a file in a new project (it would be the default), or not, if we prefer to have a small list of saved projects and use the "save" command ourselves (as it is the case now).

### DIFF
--- a/src/commands/project.lisp
+++ b/src/commands/project.lisp
@@ -242,17 +242,11 @@
   "Add this project path to the history file.
   Return t if the project was added, nil if it already existed."
   (let* ((history (history))
-         (data (lem/common/history:history-data history))
-         ;; project roots (pathnames) are converted to strings for the
+         ;; Project roots (pathnames) are converted to strings for the
          ;; string completion prompt.
          (input (namestring input)))
-    (cond
-      ((not (find input data :test #'equal))
-       (lem/common/history:add-history history input)
-       (lem/common/history:save-file history)
-       input)
-      (:otherwise
-       nil))))
+    (lem/common/history:add-history history input :allow-duplicates nil)
+    (lem/common/history:save-file history)))
 
 (defun forget-project (input)
   "Remove this project (string) from the projects' history file.
@@ -265,15 +259,12 @@
 (defun saved-projects ()
   "Return the saved projects.
   Return: a list (of strings), not a vector."
-  (coerce
-   (lem/common/history:history-data (history))
-   'list))
+  (lem/common/history:history-data-list (history)))
 
 (define-command project-save () ()
   "Remember the current project for later sessions."
-  (if (remember-project (find-root (buffer-directory)))
-      (message "Project saved.")
-      (message "Already saved.")))
+  (when (remember-project (find-root (buffer-directory)))
+    (message "Project saved.")))
 
 (define-command project-unsave () ()
   "Prompt for a project and remove it from the list of saved projects."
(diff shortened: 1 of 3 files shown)


src/commands/project.lisp
```diff
@@ -221,3 +224,85 @@
...
+         (data (lem/common/history:history-data history))
+         ;; project roots (pathnames) are converted to strings for the
+         ;; string completion prompt.
+         (input (namestring input)))
+    (cond
+      ((not (find input data :test #'equal))
+       (lem/common/history:add-history history input)
+       (lem/common/history:save-file history)
+       input)
+      (:otherwise
```
cxxxr, line 254: Is there a mistake with the T?

src/commands/project.lisp
```diff
@@ -221,3 +224,85 @@
...
+    (let* ((pathname (merge-pathnames "history/projects" (lem-home)))
+           (history (lem/common/history:make-history :pathname pathname)))
+      (setf *projects-history* history)))
+  *projects-history*)
+
+(defun remember-project (input)
+  "Add this project path to the history file.
+  Return t if the project was added, nil if it already existed."
+  (let* ((history (history))
+         (data (lem/common/history:history-data history))
```
cxxxr, line 245: How about defining a function that searches through the history to facilitate encapsulation?

src/commands/project.lisp
```diff
@@ -221,3 +224,85 @@
...
+
+(defun forget-project (input)
+  "Remove this project (string) from the projects' history file.
+  Return the input."
+  (let ((history (history)))
+    (lem/common/history:remove-history history input)
+    (lem/common/history:save-file history)
+    input))
+
+(defun saved-projects ()
```
cxxxr, line 265: To facilitate encapsulation, why not define a function in history.lisp that returns the history as a list?

src/commands/project.lisp
```diff
@@ -221,3 +224,85 @@
...
+(defun prompt-for-project ()
+  "Prompt for a project saved in the projects history."
+  (let ((candidates (saved-projects)))
+    (if candidates
+        (prompt-for-string
+         "Project: "
+         :completion-function (lambda (x) (completion-strings x candidates))
+         :test-function (lambda (name) (member name candidates :test #'string=)))
+        (let ((lem-core::*message-timeout* 5))
+          (message "No projects.~&~
```
cxxxr, line 295: The function show-message allows you to specify a timeout.

cxxxr (commented):
Very attractive feature, thank you very much!
I have made a few comments and would appreciate your confirmation.

src/commands/project.lisp
vindarel: You are right, I'll try to facilitate encapsulation. Thanks for the review.

src/commands/project.lisp
```diff
@@ -10,13 +10,16 @@
...
            :project-kill-buffers)
-  (:documentation "Defines utilities to find a project root directory and related user-facing commands: project-find-file, project-kill-buffers etc."))
+  (:documentation "Defines utilities to find a project root directory and related user-facing commands: project-find-file, project-kill-buffers, project-switch-project etc."))
 
 (in-package :lem-core/commands/project)
 
 (define-key *global-keymap* "C-x p f" 'project-find-file)
 (define-key *global-keymap* "C-x p d" 'project-root-directory)
 (define-key *global-keymap* "C-x p K" 'project-kill-buffers)
+(define-key *global-keymap* "C-x p p" 'project-switch-project)
```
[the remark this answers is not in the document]
vindarel: yes. This is elisp influence^^

src/commands/project.lisp
vindarel: I addressed my needs in history-add by adding a &key allow-duplicates.

cxxxr:
Thank you so much!

PR #760, opened 2023-06-19
Document some color theme functions

Neshamon:
# Documenting Lem
* Documents find-color-theme, define-color-theme, inherit-load-theme, and apply-color-theme


cxxxr:
Thank you!

PR #768, opened 2023-06-20
grep: add keys and help

vindarel:
At first I added "n" and "p", but they interfere with vi-mode.

A global "grep-help" command "clutters" a bit the global list of commands… but being a help, it isn't a bad thing.

### DIFF
--- a/src/ext/peek-source.lisp
+++ b/src/ext/peek-source.lisp
@@ -44,6 +44,9 @@
 (define-key *peek-source-keymap* "Return" 'peek-source-select)
 (define-key *peek-source-keymap* 'next-line 'peek-source-next)
 (define-key *peek-source-keymap* 'previous-line 'peek-source-previous)
+(define-key *peek-source-keymap* "Escape" 'peek-source-quit)  ;; also C-x 0 by default.
+(define-key *peek-source-keymap* "C-c C-k" 'peek-source-quit)
+(define-key *peek-source-keymap* "M-q" 'peek-source-quit)
 
 (defclass peek-window (floating-window) ())
 (defclass source-window (floating-window) ())
--- a/src/ext/peek-source.lisp
+++ b/src/ext/peek-source.lisp
@@ -47,6 +47,7 @@
 (define-key *peek-source-keymap* "Escape" 'peek-source-quit)  ;; also C-x 0 by default.
 (define-key *peek-source-keymap* "C-c C-k" 'peek-source-quit)
 (define-key *peek-source-keymap* "M-q" 'peek-source-quit)
+(define-key *peek-source-keymap* "C-x ?" 'grep-help)  ;; originally bound to describe-key.
 
 (defclass peek-window (floating-window) ())
 (defclass source-window (floating-window) ())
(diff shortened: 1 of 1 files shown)


vindarel:
oopsy^^ good catch.

Escape could work to quit

for the help I don't know which key to add, "?" must not be used too and C-h is already bound to delete-previous-char.

vindarel:
> Hmm, C-x ? is describe-key, but it might be a good idea to extend it.

maybe… it can also be useful to describe keys in this peek-source mode.

`C-x  h` is "mark set whole buffer", not very useful here I think.

> Escape could work to quit

btw my PR is due because I didn't discover if there is an easy way to quit the grep windows.

cxxxr:
> C-x  h is "mark set whole buffer", not very useful here I think.
If possible, we would like to have a consistency that can be used in other modes.
It might be a good idea to change the meaning of C-x h itself.

This mode itself can be exited with `C-x 0`

vindarel:
Done and rebased. It was simpler for me to go with C-x ?

src/ext/peek-source.lisp
```diff
@@ -43,7 +43,12 @@
 
 (define-key *peek-source-keymap* "Return" 'peek-source-select)
 (define-key *peek-source-keymap* 'next-line 'peek-source-next)
+(define-key *peek-source-keymap* "C-n" 'peek-source-next)
 (define-key *peek-source-keymap* 'previous-line 'peek-source-previous)
+(define-key *peek-source-keymap* "C-p" 'peek-source-previous)
```
cxxxr, line 48: `C-n/C-p`
Even without key bindings, I think the cursor position is displayed automatically.
Is the current behavior not enough?

src/ext/peek-source.lisp
```diff
@@ -43,7 +43,12 @@
...
+(define-key *peek-source-keymap* "Escape" 'peek-source-quit)  ;; also C-x 0 by default.
+(define-key *peek-source-keymap* "C-c C-k" 'peek-source-quit)
```
cxxxr, line 50: Well, this key binding may be good.

I came up with another one called `M-q`, what do you think?
I would like to use it in a unified operation in other places.

src/ext/peek-source.lisp
```diff
@@ -223,6 +228,24 @@
...
+    (format s "Available keybindings:~&")
+    (format s "- up/down arrows or C-p/C-n: go to the previous/next line~&")
+    (format s "- C-x o or M-o: go to the other window~&")
+    (format s "- Enter: visit the file of the result at point~&")
+    (format s "- Escape or C-x 0: quit~&")
+    (format s "- C-x ?: bring this help~&")
+    (format s "~%")
+    (format s "The results buffer on the left is editable. Any change is written to file and is reflected immediately on the right.~&")
+    (format s "You can use editing tools such as M-x query-replace in the results buffer.~&")
+    (format s "~%")))
```
cxxxr, line 247: very good 👍 
I am not very good at writing in English, so this kind of thing is very helpful.

src/ext/peek-source.lisp
vindarel: The current behavior works fine with the up and down keys.
However we don't have, by default, a movement key on the home row, that's why I was looking for more movement keys.

I realize that C-n and C-p are not appropriate, because we can edit the buffer, so they will overwrite the editor keys (same issue with "q" and "?").

I will try later to make these keys work only on the read-only text,

for now I remove them.

src/ext/peek-source.lisp
vindarel: To be honest, I added C-c C-k in the hope to ease the transition for Emacsers. I didn't document it in the help.

M-q is simple and good, I added it.

cxxxr:
Thank you so much!

Issue #757, opened 2023-06-18
Using tree-sitter for syntax highlight 


Sasanidas:
Tree sitter (https://tree-sitter.github.io/tree-sitter/) is a very popular library for parsing languages, it's more useful for non-lisp languages mainly for syntax highlight and some local refactoring. It also have a very interesting query-api(https://tree-sitter.github.io/tree-sitter/using-parsers#the-query-api).

Luckily for us, someone already did the bindings for the C API of tree-sitter https://github.com/death/cl-tree-sitter with an MIT license so we can probably use it.

cxxxr:
Thank you for sharing!
This is very interesting.

CarsonEllsworth33:
I just started working on this feature. Not much progress made yet other than basic parsing through the REPL right now (no buffer parsing yet).  

I am using death's cl-tree-sitter and it seems to work.   
There is either an issue on my end or within death's parsing code as I can't load grammars that I have built locally unless I rename them from libtree-sitter-<language>.so to tree-sitter-<language>.so

CarsonEllsworth33:
![tree-sitter](https://github.com/user-attachments/assets/5d43683e-344b-425b-9c30-8fc1ee102ed4)

CarsonEllsworth33:
![libtree-sitter](https://github.com/user-attachments/assets/c937478c-181a-446a-9dbf-38a009d122cb)

CarsonEllsworth33:
![error](https://github.com/user-attachments/assets/33ebb9a1-0b12-424c-ad99-2527320a6220)

CarsonEllsworth33:
This is the issue I mentioned in my first comment about having to rename the shared library.

cianoc:
Did you make any progress on this?

logoraz:
There is another set of CL bindings for libtree-sitter that you may want to try: https://github.com/garlic0x1/cl-treesitter

cianoc:
Given the author - that's probably a pretty solid library.

CarsonEllsworth33:
I approached trying to get tree-sitter into Lem when I was fresh to both Lem and common lisp.  I got derailed a while back and am working on a few easier things currently to get better at CL and better learn Lem's internals, but still have it in a backlog of stuff I want to work on for Lem. Last time I checked the cl-tree-sitter binding by death was breaking due to a SAP null error (or something like it). So it may be worth a look into the bindings by garlic0x1 as mentioned above.

CarsonEllsworth33:
[...]

PR #788, opened 2023-06-27
Add imenu (detective) support for Lem

Sasanidas:
The idea of this new feature is to have a version of imenu ready to work with regex but also with other external tools (like for example tree-sitter).

One of the interesting challenges of this approach is to parse the buffer efficiently with regex and organize all of them correctly. I'll focus on this the last part given that initially I don't think the overhead is that much.

### DIFF
--- a/modes/lisp-mode/lisp-mode.lisp
+++ b/modes/lisp-mode/lisp-mode.lisp
@@ -43,6 +43,12 @@
         (make-completion-spec 'completion-symbol-async :async t))
   (setf (variable-value 'idle-function) 'lisp-idle-function)
   (setf (variable-value 'root-uri-patterns) '(".asd"))
+  (setf (variable-value 'detective-search) 
+        (make-instance 'lem/detective::search-regex 
+                       :function-regex
+                       (lem/detective::make-capture-regex 
+                        :regex "^\\(defun "
+                        :function #'capture-reference)))
   (set-syntax-parser lem-lisp-syntax:*syntax-table*
                      (make-tmlanguage-lisp))
   (unless (connected-p) (self-connect))
@@ -1192,6 +1198,16 @@
 
 (add-hook (variable-value 'before-eval-functions :global)
           'highlight-evaluation-region)
+
+(defmethod capture-reference ((position lem:point) (class (eql :function-reference)))
+  (let* ((line (str:split #\Space (line-string position)))
+         (pname (second line))
+         (name (or (and (str:starts-with-p "(setf" pname)
+                        (str:concat pname " " (third line)))
+                   pname)))
+    (make-instance 'lem/detective::function-reference
+                   :reference-name name
+                   :reference-point position)))
 
 ;; workaround for windows
 #+win32
--- a/modes/lisp-mode/lisp-mode.lisp
+++ b/modes/lisp-mode/lisp-mode.lisp
@@ -44,21 +44,21 @@
   (setf (variable-value 'idle-function) 'lisp-idle-function)
   (setf (variable-value 'root-uri-patterns) '(".asd"))
   (setf (variable-value 'detective-search) 
-        (make-instance 'lem/detective::search-regex 
+        (make-instance 'lem/detective:search-regex
                        :function-regex
-                       (lem/detective::make-capture-regex 
+                       (lem/detective:make-capture-regex
                         :regex "^\\(defun "
                         :function #'capture-reference)
                        :class-regex
-                       (lem/detective::make-capture-regex
+                       (lem/detective:make-capture-regex
                         :regex "^\\(defclass "
                         :function #'capture-reference)
                        :package-regex
-                       (lem/detective::make-capture-regex
+                       (lem/detective:make-capture-regex
                         :regex "^\\(in-package "
                         :function #'capture-reference)
 		       :variable-regex
-                       (lem/detective::make-capture-regex
+                       (lem/detective:make-capture-regex
                         :regex "^\\(defvar |\\(defparameter "
                         :function #'capture-reference)))
   (set-syntax-parser lem-lisp-syntax:*syntax-table*
@@ -1217,21 +1217,21 @@
          (name (or (and (str:starts-with-p "(setf" pname)
                         (str:concat pname " " (third line)))
                    pname)))
-    (make-instance 'lem/detective::function-reference
+    (make-instance 'lem/detective:function-reference
                    :reference-name name
                    :reference-point position)))
 
 (defmethod capture-reference ((position lem:point) (class (eql :class-reference)))
   (let* ((line (str:split #\Space (line-string position)))
          (name (second line)))
-    (make-instance 'lem/detective::class-reference
+    (make-instance 'lem/detective:class-reference
                    :reference-name name
                    :reference-point position)))
 
 (defmethod capture-reference ((position lem:point) (class (eql :variable-reference)))
   (let* ((line (str:split #\Space (line-string position)))
          (name (second line)))
-    (make-instance 'lem/detective::variable-reference
+    (make-instance 'lem/detective:variable-reference
                    :reference-name name
                    :reference-point position
                    :variable-reference-value (third line))))
@@ -1239,7 +1239,7 @@
 (defmethod capture-reference ((position lem:point) (class (eql :package-reference)))
   (let* ((line (str:split #\Space (line-string position)))
          (name (second line)))
-    (make-instance 'lem/detective::package-reference
+    (make-instance 'lem/detective:package-reference
                    :reference-name name
                    :reference-point position)))
 
--- a/modes/lisp-mode/lisp-mode.lisp
+++ b/modes/lisp-mode/lisp-mode.lisp
@@ -50,8 +50,16 @@
                         :regex "^\\(defun "
                         :function #'capture-reference)
                        :class-regex
-                       (lem/detective::make-capture-regex 
+                       (lem/detective::make-capture-regex
                         :regex "^\\(defclass "
+                        :function #'capture-reference)
+                       :package-regex
+                       (lem/detective::make-capture-regex
+                        :regex "^\\(in-package "
+                        :function #'capture-reference)
+		       :variable-regex
+                       (lem/detective::make-capture-regex
+                        :regex "^\\(defvar |\\(defparameter "
                         :function #'capture-reference)))
   (set-syntax-parser lem-lisp-syntax:*syntax-table*
                      (make-tmlanguage-lisp))
@@ -1217,6 +1225,21 @@
   (let* ((line (str:split #\Space (line-string position)))
          (name (second line)))
     (make-instance 'lem/detective::class-reference
+                   :reference-name name
+                   :reference-point position)))
+
+(defmethod capture-reference ((position lem:point) (class (eql :variable-reference)))
+  (let* ((line (str:split #\Space (line-string position)))
+         (name (second line)))
+    (make-instance 'lem/detective::variable-reference
+                   :reference-name name
+                   :reference-point position
+                   :variable-reference-value (third line))))
+
+(defmethod capture-reference ((position lem:point) (class (eql :package-reference)))
+  (let* ((line (str:split #\Space (line-string position)))
+         (name (second line)))
+    (make-instance 'lem/detective::package-reference
                    :reference-name name
                    :reference-point position)))
 
--- a/modes/lisp-mode/lisp-mode.lisp
+++ b/modes/lisp-mode/lisp-mode.lisp
@@ -48,6 +48,10 @@
                        :function-regex
                        (lem/detective::make-capture-regex 
                         :regex "^\\(defun "
+                        :function #'capture-reference)
+                       :class-regex
+                       (lem/detective::make-capture-regex 
+                        :regex "^\\(defclass "
                         :function #'capture-reference)))
   (set-syntax-parser lem-lisp-syntax:*syntax-table*
                      (make-tmlanguage-lisp))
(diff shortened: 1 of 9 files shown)


src/base/buffer.lisp
```diff
@@ -30,6 +30,10 @@
     :initarg :%enable-undo-p
     :accessor buffer-%enable-undo-p
     :type boolean)
+   (references 
+    :initform (make-hash-table :test 'equal)
+    :reader buffer-references
+    :accessor buffer-references )
```
cxxxr, line 36: How about using `buffer-value`?

For example, define the following function where this is needed.
```common-lisp
(defun buffer-references (buffer)
  (buffer-value buffer 'references))
(defun (setf buffer-references) (value buffer)
  (setf (buffer-value buffer 'references) value))
```

Sasanidas: Oh I see, makes sense, I'll add it :+1:

Sasanidas:
So, I add the editor variable detective-search, the idea is to be used for the languages modes to save the object that will trigger the desired behaviour. In this case it's using the regex class, but in the future we can use the tree-sitter to get the different tags https://github.com/lem-project/lem/issues/757

Sasanidas:
Aside from the imenu-like functionality, I added two very useful command for file navigation,
 `detective-next` which moves the cursor to the closest reference location forward and `detective-prev` which does the same but backwards:


https://github.com/lem-project/lem/assets/26670956/c4f626ed-fa8e-4337-bddc-a6cec9e479b8

cxxxr:
Wow, nice!

Sasanidas:
> Wow, nice! I wish we had some useful key bindings.

Indeed, I'm not sure which keybinding would be nice by default, I'm open to ideas :raised_hands:

Sasanidas:
I think it's ready for a review :eyes:

src/ext/detective.lisp
```diff
@@ -0,0 +1,282 @@
...
+
+(defun (setf buffer-references) (value buffer)
+  (setf (buffer-value buffer 'references) value))
+
+(defgeneric search-references (search-class))
+
+(defmethod search-references ((search search-regex))
+  (labels ((find-ref (regex class)
+             (with-point ((p (buffer-start-point (point-buffer (current-point)))))
+               (loop :for position = (search-forward-regexp p (capture-regex-regex regex))
```
cxxxr, line 141: In lem, the name position is often used for an integer (e.g. `cl:position` function), and point is often used to represent a `lem:point`.
I think it is less confusing to use the variable point here.

src/ext/detective.lisp
```diff
@@ -0,0 +1,282 @@
...
+                     :while position
+                     :collect (funcall (capture-regex-function regex) 
+                                       (copy-point position)
```
cxxxr, line 144: This will probably never be released as copied.
If the buffer is never updated while using the copied point, it seems better to use `:temporary`.

src/ext/detective.lisp
```diff
@@ -0,0 +1,282 @@
...
+(defmethod move-to-reference (reference)
+  (message "Not reference available in current buffer."))
+
+(defun check-change ()
+  (cond 
+   ((null (buffer-references (current-buffer)))
+    (search-references 
+     (variable-value 'lem/language-mode:detective-search :buffer)))
+   
+   ((changed-disk-p (current-buffer))
```
cxxxr, line 203: It will maybe need to be updated if the buffer has been edited since it was last retrieved.

cxxxr: It seems that lem:buffer-modified-tick can be used to detect this.

src/ext/detective.lisp
```diff
@@ -0,0 +1,282 @@
...
+          :for rline = (line-number-at-point (reference-point ref))
+          :do (progn
+                (when flag (setf closest ref))
+                (let* ((cdiff (- (line-number-at-point (reference-point closest)) cline))
+                       (ldiff (- rline cline)))
+
+                  (when  (or (and (> rline cline)
+                                  (<  ldiff cdiff))
+                             (< cdiff 0))
+                    (setf closest ref)))))
```
cxxxr, line 223: There is a function called `lem:point<`.
For example, could it be written as follows?

```common-lisp
(loop :for (start end) :on ...points...
      :do (when (point< start (current-point) end)
            (return end)))

```

src/ext/detective.lisp
```diff
@@ -0,0 +1,282 @@
...
+
+
+(defclass misc-reference (reference)
+  ((custom-type :initarg :misc-custom-type
+                :type string
+                :reader misc-custom-type)))
+
+(defstruct capture-regex regex function)
+
+(defclass detective-search () ())
```
cxxxr, line 100: This name may be confusing as it seems to be covered with the symbol for `lem/language-mode:detective-search`.
Since it seems to be an abstract class, how about something like `<detective-search>`?

src/ext/detective.lisp
```diff
@@ -0,0 +1,282 @@
...
+               class-regex
+               variable-regex
+               misc-regex)
+      search
+    (let ((slots 
+            (list (cons (cons "functions" function-regex):function-reference)
+                  (cons (cons "classes" class-regex) :class-reference)
+                  (cons (cons "packages" package-regex) :package-reference)
+                  (cons (cons "variables" variable-regex) :variable-reference)
+                  (cons (cons "misc" misc-regex) :misc-reference))))
```
cxxxr, line 157: If it is an ID, wouldn't a keyword be preferable to a string?
This may be a matter of taste.

cxxxr (commented):
This PR is very good.
The definitions and abstractions using CLOS were very interesting and there was much to learn from reading this code.

I have commented on the improvements, thank you.

modes/lisp-mode/lisp-mode.lisp
```diff
@@ -43,6 +43,24 @@
         (make-completion-spec 'completion-symbol-async :async t))
   (setf (variable-value 'idle-function) 'lisp-idle-function)
   (setf (variable-value 'root-uri-patterns) '(".asd"))
+  (setf (variable-value 'detective-search) 
+        (make-instance 'lem/detective:search-regex
```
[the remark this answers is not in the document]
Sasanidas: What do you mean? Sorry I don't understand "def[\\w-]*"

modes/lisp-mode/lisp-mode.lisp
```diff
@@ -1193,6 +1211,38 @@
...
+                   :reference-point position
+                   :variable-reference-value (third line))))
+
+(defmethod capture-reference ((position lem:point) (class (eql :package-reference)))
+  (let* ((line (str:split #\Space (line-string position)))
+         (name (second line)))
+    (make-instance 'lem/detective:package-reference
+                   :reference-name name
+                   :reference-point position)))
+
```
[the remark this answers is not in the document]
Sasanidas: Indeed, this was just a simple example on how to extract the basic data of each reference, the idea is to build upon these and fill more object attributes (for example with the classes slots). But I think that was out of the scope of detective itself.

src/ext/detective.lisp
```diff
@@ -0,0 +1,282 @@
...
+                     :while position
+                     :collect (funcall (capture-regex-function regex) 
+                                       (copy-point position)
+                                       class)))))
+  (with-slots (function-regex
+               package-regex
+               class-regex
+               variable-regex
+               misc-regex)
+      search
```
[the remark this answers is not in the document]
Sasanidas: Done :+1:

modes/lisp-mode/lisp-mode.lisp
```diff
@@ -43,6 +43,24 @@
...
+                       (lem/detective:make-capture-regex
+                        :regex "^\\(defclass "
+                        :function #'capture-reference)
+                       :package-regex
+                       (lem/detective:make-capture-regex
+                        :regex "^\\(in-package "
+                        :function #'capture-reference)
+		       :variable-regex
+                       (lem/detective:make-capture-regex
+                        :regex "^\\(defvar |\\(defparameter "
```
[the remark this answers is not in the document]
Sasanidas: Done :+1:  (I have to improve my regex skills indeed)

src/ext/detective.lisp
```diff
@@ -0,0 +1,282 @@
...
+                             (< cdiff 0))
+                    (setf closest ref)))))
+
+    (if (< (- (+ (line-number-at-point (current-point)) 1)
+              (line-number-at-point (reference-point closest))) 0)
+        (move-to-reference closest)
+        (message "No next reference."))))
+
+
+(define-command detective-prev () ()
```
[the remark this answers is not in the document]
Sasanidas: Done :+1:

src/ext/detective.lisp
Sasanidas: Done :+1:

src/ext/detective.lisp
Sasanidas: Sorry what do you mean with `:temporary`? Like another way of having the point?

src/ext/detective.lisp
Sasanidas: Maybe something like this?
```lisp
(defun check-change ()
  (cond 
   ((null (buffer-references (current-buffer)))
    (search-references 
     (variable-value 'lem/language-mode:detective-search :buffer)))
   
   ((not (eql (buffer-value (current-buffer) 'prev-tick)
              (buffer-modified-tick (current-buffer))))
    
    (setf (buffer-value (current-buffer) 'prev-tick)
          (buffer-modified-tick (current-buffer)))
    (search-references (variable-value 'lem/language-mode:detective-search :buffer)))))

```

src/ext/detective.lisp
Sasanidas: Oh, I wasn't aware of that way of naming abstract classes, I changed it :+1:

src/ext/detective.lisp
Sasanidas: I tend to use strings for no particular reason, if you want I can change it to a keyword, I have no problem.

src/ext/detective.lisp
```diff
@@ -0,0 +1,282 @@
...
+                  (cons (cons "variables" variable-regex) :variable-reference)
+                  (cons (cons "misc" misc-regex) :misc-reference))))
+    
+    (setf (buffer-references (current-buffer))
+          (make-hash-table :test 'equal))
+
+      (loop :for ((id . regex) . class) :in slots
+            :when (and regex (capture-regex-function regex))
+            :do (setf (gethash id (buffer-references (current-buffer)))
+                      (find-ref regex class)))))))
```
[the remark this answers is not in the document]
Sasanidas: Ups, seems like I accidentally open the file with Emacs again (without sly) :facepalm:

src/ext/detective.lisp
cxxxr: Yes, it looks good.

modes/lisp-mode/lisp-mode.lisp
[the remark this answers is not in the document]
cxxxr: `[\\w-]` is equivalent to `[a-zA-Z0-9-]`.

For example, suppose the user defines a macro named define-foo or defxxx.
We would like to be able to treat this in the same way as defun and defmethod.

modes/lisp-mode/lisp-mode.lisp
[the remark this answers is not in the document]
cxxxr: I see, my comment was based on the feeling that the responsibilities of lisp-mode.lisp are too broad.

src/ext/detective.lisp
cxxxr: https://github.com/lem-project/lem/blob/dd0e5bf1b9a89aa61a5e5fc2e91cc348d3c5a805/src/base/point.lisp#L20
Would this be helpful?

modes/lisp-mode/lisp-mode.lisp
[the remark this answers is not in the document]
Sasanidas: Oooh, so something that is not strictly a "defun", so the idea of these constructs is to fill the misc reference

modes/lisp-mode/lisp-mode.lisp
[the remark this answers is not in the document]
Sasanidas: Oh now I understand, let me create a detective-lisp file :+1:

src/ext/detective.lisp
Sasanidas: Oh, so copy the point with the kind as `:temporary`?

src/ext/detective.lisp
Sasanidas: Added :+1:

src/ext/detective.lisp
cxxxr: Yes (but if you then change the buffer, that point may be broken)

modes/lisp-mode/lisp-mode.lisp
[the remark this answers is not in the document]
Sasanidas: Done :+1:

src/ext/detective.lisp
Sasanidas: Umm, so if the buffer content got deleted, that point may not exist anymore. So, the idea is that shouldn't happen because of the check if the buffer was changed. (did I understand it correctly?)

src/ext/detective.lisp
Sasanidas: So, I added a function to the point files to compare a list of points relative to the desired point and it returns the closest one down or up with optionally check if the point is in the same line :+1:  [(src/base/point.lisp)](https://github.com/lem-project/lem/pull/788/files#diff-6f18753392faf0e63518ba0d628865c3844a1a9dd9752befe8dbd9d94dc3c7afR196)

Sasanidas: This allowed me to remove most of the code from the next and previous functions, let me know what do you think

Sasanidas:
I was planning to add `which-function` functionality from GNU Emacs, but I think I'll wait for other pull request :+1:

cxxxr:
Looks good.
Thank you for the nice PR!

PR #792, opened 2023-06-28
Add trace commands to lisp-mode

cxxxr:
The following two commands have been added.
- `M-x lisp-trace-list (C-c T)`
- `M-x lisp-toggle-trace (C-c C-t)`

### DIFF
--- a/modes/lisp-mode/lem-lisp-mode.asd
+++ b/modes/lisp-mode/lem-lisp-mode.asd
@@ -38,6 +38,7 @@
                (:file "organize-imports")
                (:file "connection-list")
                (:file "self-insert-hook")
+               (:file "trace")
                (:file "class-browser")
                (:file "package")))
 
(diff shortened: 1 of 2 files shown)


vindarel:
Hello, when I use `C-c C-t` on a function, the feedback is NIL. It would help to know if the function just got traced or untraced.

C-c T when no function is traced only shows a blank box.

cxxxr:
@vindarel
Thanks for the feedback!
I've changed the behavior.
If you have a better idea, please let me know.

vindarel:
Thanks, it's clearer for me.

but: call C-c C-t, press "Enter" on the empty prompt without choosing anything -> "end of file on #<dynamic-extent STRING-INPUT-STREAM (unavailable) from "" : ]

cxxxr:
Thanks!
This problem has been fixed.

PR #793, opened 2023-06-28
Add command prompt navigation forward and backwards

Sasanidas:


cxxxr (commented):
It is a good feature.
How about looking for the prompt as follows?

```
(lem:previous-single-property-change (lem:current-point) :field)
(lem:next-single-property-change (lem:current-point) :field)
```

cxxxr:
Indeed, you may need to shift the starting position when searching

PR #799, opened 2023-06-30
fix blurry text on Apple retina display

VitoVan:
imagen

This patch handles the text rendering on high-DPI screen for macOS.

I mark this PR Work-in-Progress, because I haven't fully tested this patch, here is the check list:

For MacOS

- [x] no more blurry text on macOS with retina screen
- [x] automatic font resizing when moving window between high-DPI screen and low-DPI screen
- [x] don't break low-DPI screen rendering (when start from the low-DPI screen)

For other platforms 

- [x] does it break the rendering on Windows?
    It was broken by itself, this patch didn't make it worse.
    #801
- [x] does it break the rendering on Linux?
    It was broken by itself, this patch didn't make it worse.
    #802

### DIFF
--- a/frontends/sdl2/font.lisp
+++ b/frontends/sdl2/font.lisp
@@ -42,8 +42,9 @@
   char-width
   char-height)
 
-(defun save-font-size (font-config)
-  (setf (lem:config :sdl2-font-size) (font-config-size font-config)))
+(defun save-font-size (font-config &optional (ratio 1))
+  (setf (lem:config :sdl2-font-size)
+        (round (/ (font-config-size font-config) ratio))))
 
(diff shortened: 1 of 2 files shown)


VitoVan:
Current problem:

If `:accelerated` is not allowed in renderer flags, then the high-DPI problem could be very hard to solve.

Without that flag, SDL2 won't be high-DPI awareness, so we can't handle it.

But, it seems there are also some problems when we set that flag: #787 

So, what's your thought?

VitoVan:
This is my first PR, I must have done many things inappropriate, so if you saw any, please don't hesitate to tell me.

cxxxr:
Thank you very much for the great PR.

PR #800, opened 2023-06-30
Implement get-font-list for windows.

jsparkes:
Returns a list of filenames as strings containing TTF and OTF fonts installed on the windows system.

(length (lem-sdl2/font:get-font-list (lem-sdl2/platform:get-platform))) => 2132 on my particular machine


cxxxr:
Thank you!

PR #810, opened 2023-07-03
Add find defintions for elisp-mode

Sasanidas:


cxxxr:
@Sasanidas 
Is this still in process?

Sasanidas:
> @Sasanidas Is this still in process?

Oh sorry, I think this is ready for review :+1:

cxxxr:
Looks good.
Thank you!

modes/elisp-mode/lem-elisp-mode.asd
```diff
@@ -2,6 +2,7 @@
   :depends-on ("lem"
                "jsonrpc"
                "lem-lisp-mode"
+               "jsonrpc/transport/http"
```
cxxxr, line 5: @Sasanidas 
I'm sorry.
After all, this dependence on ironclad, crack, etc. seems to be quite large.
Is there any way we can delay it until we use a specific feature of elisp-mode?

Sasanidas: Umm I see, I can separate the elisp-mode form the rpc interface, so elisp-mode will only provide basic major-mode

Sasanidas: I think this should solve the issue https://github.com/lem-project/lem/pull/825 :+1:

PR #815, opened 2023-07-04
Fixed overlays not being considered when finding the x position in a window

BierLiebHaber:
As the title says this should fix the x position of a mouse click when the line numbers overlay is active.


cxxxr:
Thank you so much!

PR #817, opened 2023-07-05
detective: export user-facing functions, add docstrings

vindarel:
@Sasanidas Allow me to chime in :D 

Exporting functions was necessary to discover them while defining keybindings.


cxxxr:
It looks good.

PR #818, opened 2023-07-05
add time-stamp command

vindarel:
I often use `org-time-stamp` when taking notes. I replicate the behaviour and the file format: 

     <2023-07-05 Wed>

(although org-mode has an interactive date picker)

### DIFF
--- a/lem.asd
+++ b/lem.asd
@@ -126,7 +126,8 @@
                              (:file "filer")
                              (:file "deepl")
                              (:file "themes")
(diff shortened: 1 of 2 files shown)


cxxxr:
It looks interesting.
This seems like too little module granularity to me.
Will you be expanding this module in the future?

vindarel:
You are right, I was thinking the same. 

I think I will… but not now.

PR #822, opened 2023-07-07
sdl2: fix retina mouse coordinates

seanfarley:
Now that lem has nice retina fonts, let's also add the same for mouse events.


cxxxr:
Thank you so much!

PR #825, opened 2023-07-09
Remove jsonrcp/http dependency from elisp-mode

Sasanidas:
Seems like there are a few unwanted dependencies form the jsonrpc/http transport class, and in this case I'm just calling it for a simple information prompt. So for now, it will load only when the user setup the rpc server as mention here https://github.com/lem-project/lem/pull/810#discussion_r1257303942


cxxxr:
Thank you

PR #831, opened 2023-07-12
Fix some minor SDL2 paper cuts

seanfarley:
These patches add the ability to get/set the gui window's title along with toggling fullscreen.


cxxxr (commented):
Thank you for the nice PR.
I have checked it on windows and mac and it seems to be working properly.

PR #844, opened 2023-07-17
Fix closest-point function

Sasanidas:
So, this functions was not working correctly, assuming that the reference were going always in descendant order, when is not the case.

I can arrange them and then correct the function, but I think it may cost too much, so I with a few more checks, I can be sure to get the correct reference.

Ping @cxxxr , because this should be included in the 2.1v release (sorry :raised_hands: )


cxxxr:
Sorry, I've already tagged it, put together the release notes, and created the application.
I would like to include this in the next version :pray:

Sasanidas:
No problem, thanks for the merge :)

PR #851, opened 2023-07-19
Add modified buffers information when exit Lem

Sasanidas:
Yesterday I was working on Lem with 30+ buffers open, and when closing the editor, it told me that I had some that were modified but not saved, so I spent a few minutes going around trying to realize which one were the ones that were open.

This patch aims to solved this issue providing information on which buffers are modified but not saved.


cxxxr:
Thank you!

PR #855, opened 2023-07-20
Vindarel/legit rebase interactive

vindarel:
Works for the base case, we can pickup/fixup commits, other verbs that need another editing action (edit, reword) don't work.

I wrote up todos and other obvious missing features.

To start a rebase: put the point on a commit, press `r i`.

Awesome though of course :sunglasses:


src/ext/legit/legit.lisp
```diff
@@ -291,14 +295,21 @@ Next:
...
   Typicaly used to run an external process in the context of a diff buffer command."
   (multiple-value-bind (output error-output exit-code)
       (funcall fn)
-    (declare (ignorable output))
     (cond
       ((zerop exit-code)
-       (message (str:join #\newline (remove-if #'null (list message output)))))
+       (let ((msg (str:join #\newline (remove-if #'null (list message output)))))
+         (when (str:non-blank-string-p msg)
+           (message msg))))
```
cxxxr, line 312: It probably malfunctions if a `~` is included.

cxxxr:
Thank you for the very nice PR!

src/ext/legit/legit.lisp
vindarel: mmh right, but the messages come from porcelain, so it should be fine?

cxxxr: Hmm, I think it's better to write in a safe way to avoid unnecessary worry.

cxxxr: That said, it's not enough of a problem to hinder development.
I am very grateful to you for doing such a great job.

PR #865, opened 2023-07-21
image viewer: add +, -, 0 keys and help

vindarel:
\+ and - are much more convenient, and they also work on my keyboard when `C-+` doesn't :D


cxxxr:
Thanks you!

PR #867, opened 2023-07-22
Vindarel/image fit to screen

vindarel:
When opening an image, if it is bigger than the screen, fit it to the screen. We still have the shortcut `0` / C-0 to see it in its original size.


cxxxr:
This is a very nice feature.
Thank yo very much!

PR #868, opened 2023-07-22
file: add find-file-directory on C-x C-j

vindarel:


vindarel:
mmh I am thinking of a better strategy to follow the file at point.

src/commands/file.lisp
```diff
@@ -103,6 +104,20 @@
...
+    (cond
+      ((null fullpath)
+       (message "No file at point"))
+      (t
+       (switch-to-buffer
+        (find-file-buffer (lem-core/commands/file::directory-for-file-or-lose (buffer-directory))))
+       (let ((filename (file-namestring fullpath)))
+         (search-forward (current-point) filename)
+         (window-recenter (current-window))
+         (character-offset (current-point) (* -1 (length filename))))))))
```
cxxxr, line 119: How about making this sequence of operations a function of the directory-mode package?

lem.asd
```diff
@@ -1,5 +1,5 @@
 (defsystem "lem"
-  :version "2.0.0"
+  :version "2.1.0"
```
cxxxr, line 2: Oops!
Seems I forgot to make this change in release.

cxxxr:
Thank you very much for always contributing to lem.
By the way, but I am wondering if it is necessary to add this command when C-x C-f does the same thing.
I believe that doing more with less is preferable to adding more.

vindarel:
The commands do not "do the same thing", but maybe could. They are subtly different. And (as we will hear a lot here): "as an Emacs user" I was glad to find out C-x C-j.

- C-x C-f: find-file, allows to open a repository. 
- C-x C-j: when we are visiting a file, we can press this and voilà, we opened the repository of the file AND the point is on the file line. Doing the same with C-x C-f would require, let's see… one more keystroke, according we also put the point on the file line.

I can have a look and see if it's easy/clean to add the functionality to C-x C-f.

However sometimes, for commands we use many times a day, for a software at the heart of our work, it is important to remove unnecessary clicks/keystrokes.

cxxxr:
Indeed, one less keystroke is an advantage, there is no doubt about that.
But it also means consuming valuable key bindings for similar commands.

Is there any other way?
For example, you can use C-x d to display the filer, and it is conceivable that the same behavior would occur at that time.

vindarel:
I will try some things.

I also invite you to add this function to your Lem init and see if you fall in love with it :D

cxxxr:
I think this feature in itself is very useful.
And in the short term, this feature is definitely useful.

Also, I have a conflict about whether it would be better to differentiate rather than implementing exactly the same functionality as Emacs

vindarel:
I close because it's on my `main` branch, adding difficulty.

By the way, what about adding the function without a keybinding? (maybe not the best)

cxxxr:
Hmmm, I seem to be resistant to the fact that lem is getting closer and closer to GNU Emacs.
Since the two are different, they do not need to be compatible, and there is no need to drag out the old history, there must be a better alternative.

cxxxr:
However, this appears to be a useful feature.

cxxxr:
Hmmm, maybe I should merge this feature.

PR #876, opened 2023-07-23
Vindarel/directory sort follow filename

vindarel:
Commits seen in https://github.com/lem-project/lem/pull/868


cxxxr:
Thank you

cxxxr:
Sorting on unexpected curosr postiion seems to cause an internal error.
image

PR #881, opened 2023-07-25
Set last focused window in switch-to-window.

Homo-carbonis:
I would like to be able to switch to the repl and then use `switch-to-last-focused-window` to return to the buffer I was last editing. Currently the use of `switch-to-last-focused-window` is limited because commands which switch windows are left to set `*last-focused-window*` themselves and most of them don't.

To resolve this I added a line to set `*last-focused-window*` in the `switch-to-window` function.

The next step would be to make all the commands which directly call (setf (current window)) use `switch-to-window` so that `*last-focused-window*` is set and `window-leave-hook` is run.

### DIFF
--- a/src/internal-packages.lisp
+++ b/src/internal-packages.lisp
@@ -203,6 +203,7 @@
    :*window-scroll-functions*
    :*window-size-change-functions*
    :*window-show-buffer-functions*
+   :*last-focused-window*
    :window-parent
    :scroll
    :window-view-point
--- a/src/internal-packages.lisp
+++ b/src/internal-packages.lisp
@@ -203,7 +203,6 @@
    :*window-scroll-functions*
    :*window-size-change-functions*
    :*window-show-buffer-functions*
-   :*last-focused-window*
    :window-parent
    :scroll
    :window-view-point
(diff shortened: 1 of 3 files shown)


src/internal-packages.lisp
```diff
@@ -203,6 +203,7 @@
    :*window-scroll-functions*
    :*window-size-change-functions*
    :*window-show-buffer-functions*
+   :*last-focused-window*
```
cxxxr, line 206: Maybe this variable should be encapsulated, what do you think?

cxxxr: Or if you need to reference it from the outside package, maybe use the `(last-focusd-window)` function?

cxxxr (commented):
Thanks for the PR.
I think it is a very nice improvement.

src/internal-packages.lisp
Homo-carbonis: Yes I think you are right. I'll export a `(last-focused-window)` function instead. It is only `switch-to-last-focused-window` which needs to access externally but I think it is best to keep the logic with the command.

cxxxr:
Thank you very much!

PR #884, opened 2023-07-27
Use (switch-to-window) where appropriate.

Homo-carbonis:
This replaces all the uses of `(setf (current-window))`to switch windows (except pop-up windows) with `(switch-to-window)`.


cxxxr:
Thank you so much!

PR #885, opened 2023-07-28
Set WM_CLASS in X11

t-sin:
Before this PR Lem SDL2 has various WM_CLASS like `sbcl`, `lem-sdl2`, etc... It caused X11's WM_CLASS is not set explicitly. Because of it we cannot configure some software detect Lem process with WM_CLASS e.g. [xkeysnail](https://github.com/mooz/xkeysnail).

This PR sets WM_CLASS as `Lem SDL2` so now users can be find Lem SDL2 process via windos class name.

I checked WM_CLASS is set after this commit like this:

```
;; clicked lem window after launching `lem-sdl2` or `lem -f sdl2` or something
$ xprop | grep WM_CLASS
WM_CLASS(STRING) = "Lem SDL2", "Lem SDL2"
```

### DIFF
--- a/frontends/sdl2/main.lisp
+++ b/frontends/sdl2/main.lisp
@@ -27,6 +27,8 @@
 (defconstant +display-width+ 100)
 (defconstant +display-height+ 40)
 
+(defvar +lem-x11-wm-class+ "Lem SDL2")
+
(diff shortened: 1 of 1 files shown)


cxxxr:
Thank you so much!

PR #889, opened 2023-08-05
Add lisp-search-symbol command

cxxxr:
Added `M-x lisp-search-symbol (C-c C-d s)`.

Use apropos to search for a symbol and jump to the location of the definition.
Inspired by lsp workspace/symbol.


vindarel:
Great!

(Similar to edit-definition-of (M-,) when the shortcut is used on a white line, not on a symbol, as it makes us choose a symbol)

PR #898, opened 2023-08-09
Add :b and :bd command for vi-mode.

fukamachi:
Add 2 commands for vi-mode:

* `:b`, equivalent to `C-x b` (`M-x select-buffer`)
* `:bd`, equivalent to `C-x k` (`M-x kill-buffer`)


cxxxr:
Thank you!

PR #899, opened 2023-08-09
Add lem-lisp-mode:*lisp-sldb-mode-hook*.

fukamachi:
Similarly to #896, I want to change the vi-mode state to NORMAL in SLDB mode.

ex. https://github.com/fukamachi/.lem/commit/ba23d4569c8d226d6378fb2ac64a03f7a49e310b


cxxxr:
Thank you.

PR #903, opened 2023-08-11
Initial idea of a general translation service

Sasanidas:
So, I saw that there is a quite handy file that handle the deepdl translation service, so I think into generalize that idea into a package, so we can define other kind of translation services.

I made a package for stumpwm called [stumpwm-translate](https://codeberg.org/sasanidas/stumpwm-translate) , which uses a https://github.com/TheDavidDelta/lingva-translate instance to translate from one language to other, using the stumpwm interface.

I also liked the idea of translate depending on the detected language (tho this can probably be more lengthy to implement so for it would be out of the scope of this issue), I'll keep it for japanese, given that it is already written so.

Also, the decode url code is taken from https://github.com/drdo/do-urlencode , mainly because I didn't wanted to add another dependecy and I found better just to borrow a few lines from there.

This is still a draft, I want to create a proper extension (maybe in the extension directory, now that it does have a more general name) and separate the functionality, I'll open this PR soon in the process to maybe get some feedback.

Thanks! :+1:


seanfarley:
Would features like this but better suited for the new package manager?

Sasanidas:
Oh you mean as a separate pacakge?
It maybe, but I think we are still thinking about how to organize the external packages, so for now I think the extension directory works :)

PR #904, opened 2023-08-11
Attempt #2: config-plist errors out when ~/.lem/config.lisp is empty

kamysheblid:
Sorry, I made a mistake in the altered config function. 

I should not have included the if statement otherwise config returns nil, not even the default value. Instead we just needed to add (car plist) otherwise everything else stays the way you wrote it.

My previous commit leaves all the settings unset.


cxxxr:
Oh, thank you.

PR #906, opened 2023-08-12
Add ';' for vi-mode to find a char in the same line repeatedly.

fukamachi:
`;` is a key command for finding a char in the same line (`f`, `F`, `t`, `T`) by the last searched query.


cxxxr:
Thank you!

PR #908, opened 2023-08-13
Introduce Vim's options feature, and add `:set` & `:cd`

fukamachi:
### Q. Why does the `:e` command no longer complete from the directory of the file being edited?

It is to make Lem's vi-mode closer to Vim's behavior. If you prefer the previous behavior, there's 2 options to achieve:

### A1. Turn `autochdir` option on

Set `autochdir` to `t` in your `init.lisp`:

```common-lisp
(setf (lem-vi-mode:autochdir) t)
```

To change the option temporarily, just execute `:set autochdir` (and `:set noautochdir` to turn off).

### A2. Change the current directory manually

Use `:cd` command to change the directory manually:

```vim
" Move to "extensions/vi-mode"
:cd extensions/vi-mode
" `%:h` expands to the directory of the current buffer's file
:cd %:h
" `-` means the previous directory
:cd -
```

## Changes

* Add Vim's options feature
  * Currently only `autochdir` is implemented
    * Works when running `:e` (`find-file`) or switching the buffer/window
* **[Incompatible changes]** Change the base directory to find with `:e` to the **current directory**
  * You need to change the directory manually with `:cd`
    * `:cd %:h` moves to the current buffer's directory (See "filename modifiers" below)
  * Or, turn `autochdir` option on
* Add "filename modifiers" which would be expanded in `:e`, `:vs`, `:sp`, and `:cd`
  * Examples:
    * `%`: The current buffer's file path
    * `%:p`: The absolute pathname of the current buffer's file path
    * `%:h`: The current buffer's directory
    * `%:e`: The file extension of the current buffer's file
    * `%:t`: The filename of the current buffer's file
    * `%:r`: The root of the filename
* Add `:cd` for changing the current directory
  * `:cd <dir>` moves to the `<dir>`
  * `:cd` moves to the home directory
  * `:cd -` moves to the previous directory
* Add `:set` (also `:se` works)
  * Examples:
    * Turn on: `:set autochdir`, `:set acd`, `:set acd=t`
    * Turn off: `:set noautochdir`, `:set noacd`, `:set acd=nil`
    * Toggle: `:set invautochdir`, `:set autochdir!`, `:set invacd`, `:set acd!`
    * Get the current value: `:set autochdir`, `:set acd?`
    * Reset to the default: `:set autochdir&`, `:set acd&`
  * In `init.lisp`
    * `(setf (lem-vi-mode:autochdir) t)`

### DIFF
--- a/extensions/vi-mode/core.lisp
+++ b/extensions/vi-mode/core.lisp
@@ -3,7 +3,7 @@
         :lem
         :lem/universal-argument)
   (:import-from :cl-package-locks)
-  (:import-from #:cl-ppcre)
+  (:import-from :cl-ppcre)
   (:export :*enable-hook*
            :*disable-hook*
            :vi-state
--- a/extensions/vi-mode/core.lisp
+++ b/extensions/vi-mode/core.lisp
@@ -18,7 +18,8 @@
            :state-enabled-hook
            :state-disabled-hook
            :normal
-           :insert))
+           :insert
+           :change-directory))
 (in-package :lem-vi-mode/core)
 
 (defvar *default-cursor-color* nil)
@@ -180,3 +181,22 @@
             (finalize-vi-modeline)
             (remove-hook *prompt-activate-hook* 'prompt-activate-hook)
             (remove-hook *prompt-deactivate-hook* 'prompt-deactivate-hook)))
+
+(defvar *previous-cwd* nil)
+
+(defun change-directory (new-directory)
+  (check-type new-directory (or string pathname))
+  (let* ((previous-directory (uiop:getcwd))
+         (new-directory (cond
+                          ((equal new-directory "")
+                           (user-homedir-pathname))
+                          ((equal new-directory "-")
+                           (or *previous-cwd* previous-directory))
+                          (t
+                           (truename
+                            (merge-pathnames (uiop:ensure-directory-pathname new-directory) previous-directory))))))
+    (assert (uiop:absolute-pathname-p new-directory))
+    (uiop:chdir new-directory)
+    (unless (uiop:pathname-equal *previous-cwd* previous-directory)
+      (setf *previous-cwd* previous-directory))
+    new-directory))
--- a/extensions/vi-mode/core.lisp
+++ b/extensions/vi-mode/core.lisp
@@ -3,6 +3,7 @@
         :lem
         :lem/universal-argument)
   (:import-from :cl-package-locks)
+  (:import-from #:cl-ppcre)
   (:export :*enable-hook*
            :*disable-hook*
            :vi-state
@@ -19,7 +20,8 @@
            :state-disabled-hook
            :normal
            :insert
-           :change-directory))
+           :change-directory
+           :expand-filename-modifiers))
 (in-package :lem-vi-mode/core)
 
 (defvar *default-cursor-color* nil)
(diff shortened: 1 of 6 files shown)


fukamachi:
We may need file modifiers too, like `:cd %:h` to move to the buffer's directory.

extensions/vi-mode/ex-command.lisp
```diff
@@ -147,3 +140,28 @@
...
+
+(define-ex-command "^set?$" (range option-string)
+  (declare (ignore range))
+  (multiple-value-bind (option-value option-name old-value isset)
+      (execute-set-command option-string)
+    (let ((lem-core::*message-timeout* 10))
+      (if (and isset
+               (not (equal option-value old-value)))
+          (lem:message "~A: ~S => ~S" option-name old-value option-value)
+          (lem:message "~A: ~S" option-name option-value)))))
```
cxxxr, line 162: What about using show-message instead of message and *message-timeout*?
```
(lem:show-message (format ...) :timeout ...)
```

extensions/vi-mode/core.lisp
```diff
@@ -3,6 +3,7 @@
         :lem
         :lem/universal-argument)
   (:import-from :cl-package-locks)
+  (:import-from #:cl-ppcre)
```
cxxxr, line 6: we use keywords for package-designator in this project, so it would be nice to see it unified.

extensions/vi-mode/options.lisp
```diff
@@ -0,0 +1,178 @@
...
+       (progn
+         (lem:add-hook lem:*find-file-hook* 'auto-change-directory)
+         (dolist (window (lem:window-list))
+           (lem:add-hook (lem-core::window-switch-to-buffer-hook window) 'auto-change-directory)
+           (lem:add-hook (lem-core:window-leave-hook window) 'auto-change-directory)))
+       (progn
+         (lem:remove-hook lem:*find-file-hook* 'auto-change-directory)
+         (dolist (window (lem:window-list))
+           (lem:remove-hook (lem-core::window-switch-to-buffer-hook window) 'auto-change-directory)
+           (lem:remove-hook (lem-core:window-leave-hook window) 'auto-change-directory))))))
```
cxxxr, line 178: Perhaps there is an issue that does not apply to windows that are split after this.
It might be a good idea to have the issue.

extensions/vi-mode/options.lisp
```diff
@@ -0,0 +1,178 @@
...
+      name))
+
+(defun get-option (name &optional (error-if-not-exists t))
+  (check-type name string)
+  (let ((name (canonical-option-name name)))
+    (multiple-value-bind (option exists)
+        (gethash name *options*)
+      (when (and (null exists)
+                 error-if-not-exists)
+        (error "Unknown option: ~A" name))
```
cxxxr, line 57: error is handled as an internal error.
Instead of treating user input as an internal error, it may be better to use editor-error or a condition that inherits from it.

cxxxr (commented):
Thank you for the PR.
I made a few petty comments, but it seems like a very good change.

extensions/vi-mode/options.lisp
fukamachi: I tried some cases on my machine, but it seems to be working. I don't know why.

cxxxr:
Please let me know when you are done making changes.

cxxxr:
I think it's good.
Thanks for the interesting PR.

PR #942, opened 2023-08-13
Add lem-if:update-cursor-shape to change the cursor's shape for lem-ncurses

fukamachi:
**This also affects other than vi-mode.**. 
Please review changes if this doesn't break other Lem's features or frontends.

Adds `lem-if:update-cursor-shape` to change the cursor's shape, like a box, a vertical bar, or an underline.  
It works only on ncurses frontend for now.

In vi-mode, the cursor will be changed to 'vertial bar' on INSERT mode.

### DIFF
--- a/frontends/ncurses/ncurses.lisp
+++ b/frontends/ncurses/ncurses.lisp
@@ -273,7 +273,7 @@
   (uiop:run-program `("printf"
                       ,(format nil "~C[~D q"
                                #\Esc
-                               (typecase cursor-type
+                               (case cursor-type
                                  (:box 2)
                                  (:bar 5)
                                  (:underline 4)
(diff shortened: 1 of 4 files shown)


frontends/ncurses/ncurses.lisp
```diff
@@ -269,6 +269,18 @@
 (defmethod lem-if:update-foreground ((implementation ncurses) color-name)
   (lem.term:term-set-foreground color-name))
 
+(defmethod lem-if:update-cursor-shape ((implementation ncurses) cursor-type)
+  (uiop:run-program `("printf"
+                      ,(format nil "~C[~D q"
+                               #\Esc
+                               (typecase cursor-type
```
cxxxr, line 276: Did you mean `case`?

cxxxr (commented):
This PR is very good.

frontends/ncurses/ncurses.lisp
fukamachi: Right! Good catch.

cxxxr:
Thank you!

PR #947, opened 2023-08-13
Don't highlight trailing spaces in popup windows.

fukamachi:
It's annoying that trailing-spaces mode highlights spaces also in popup windows like completion.

This fixes it by disabling the scanning on `not-switchable-buffer`.

## Before
Screen Shot 2023-08-13 at 23 48 31

## After
Screen Shot 2023-08-13 at 23 49 48


cxxxr:
Thank you!

PR #948, opened 2023-08-14
Add ',' to search in the line repeatedly but backward

fukamachi:
It's a pair of `;`, but repeat the last search in the opposite direction.


cxxxr:
Thank you!

PR #949, opened 2023-08-14
Small changes to paredit-mode

fukamachi:
## Changes

* Escape a double quote when inserting it into a string (`paredit-insert-doublequote`)
* Add a space between the previous element when wrapping an element (`paredit-wrap`)
* Fix `paredit-raise` to work when the cursor is at the beginning of a form
* Bind `paredit-splice-backward` to `"M-Up"`
* Add `paredit-meta-doublequote` to wrap a form with double quotes
* Rename `paredit-wrap` -> `paredit-wrap-round`, and export it
* Fix `paredit-forward` and `paredit-backward` to go up to the parent form at the beginning/end of the form


cxxxr:
Thanks a lot for the improvements

PR #950, opened 2023-08-14
Add "number" option to show line numbers in vi-mode.

fukamachi:
Also, stop defining functions for each option, because `number` conflicts with the common-lisp package. (ref #908)


cxxxr:
Looks good!

PR #952, opened 2023-08-15
Refactor definitions of vi-mode operators.

fukamachi:
Add `define-vi-motion` and `define-vi-operator` macros, and rewrite vi-mode's operator commands with them.

This aims to make them much cleaner and behave more consistently.

## Bugs fixed

* Fix `dh`, `ch`, `d^` and `c^` not to delete the char under the cursor
* Fix `1000dl` to delete the last char of the line
* Fix `df<char>`, `cf<char>`, `dt<char>`, and `ct<char>` to delete the char under the cursor after the movement
* Fix `dF<char>`, `dT<char>` not to delete the char under the cursor after the movement
* Fix `cj` to remain a single line
* Fix `yy` to allow to take a universal argument
* Fix `J` not to skip empty lines, and don't allow to concatenate the following ones
* Fix `J` and `gJ` to allow to take a universal argument, like `3J` to join 3 lines
* Fix `J` not to add a space in between if the next line starts with a ')' not only when the current line ends with ')'
* Fix `J` in visual mode to join lines between pointers
* Fix `D` in visual mode to delete the whole lines between pointers
* Fix `r` to work in visual mode
* Fix `r` to allow to take a universal argument, like `2r<char>`
* Fix the cursor position when `s` is pressed on the last char of the line
* Fix `dG` to delete to the end of the buffer instead to the begin
* Fix `dM` and `dL` to delete the line if the pointer is the end of the buffer after the movement
* Add `gU` and `gu` to change cases in normal mode
* ...and more, fix a lot of operations that move the cursor to the wrong position after they run

I also found strange behavior of `e` and `E`, but I decided to leave them not to make changes huge.


src/defcommand.lisp
fukamachi, line 1: I changed `define-command` to allow to take `:initargs` to specify `:default-initargs` to the class.

I don't believe this breaks any other parts, but please keep in mind.

src/base/package.lisp
```diff
@@ -148,6 +148,7 @@
    :delete-point
    :point-buffer
    :point-charpos
+   :point-linum
```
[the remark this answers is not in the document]
fukamachi: Thanks!
I rewrote it with the function and stopped exporting.

cxxxr (commented):
I think it's a very nice change.
One thing that concerns me is that there are several compile-time warnings, could you please fix them?

extensions/vi-mode/commands/utils.lisp
```diff
@@ -0,0 +1,196 @@
...
+  (let ((len (length (line-string point))))
+    (or (zerop len)
+        (>= (point-charpos point)
+            (1- len)))))
+
+(defun goto-eol (point)
+  "Goto end of a line."
+  (line-end point)
+  (unless (bolp point)
+    (character-offset point *cursor-offset*)))
```
cxxxr, line 45: It may be better to move before using definitions.

cxxxr:
Thank you for so many improvements!

PR #957, opened 2023-08-15
Add test detection with detective and some handy lisp commands to run the test on the REPL

Sasanidas:
- Add the detective ability to get misc references (only tests for now)
- Add a few convenience options to detective, so it's more flexible
- Add two commands to run tests:  `lisp-test-run-buffer` and `lisp-test-run-current`

### DIFF
--- a/extensions/lisp-mode/detective.lisp
+++ b/extensions/lisp-mode/detective.lisp
@@ -6,7 +6,7 @@
 
 (defun %default-capture (class position)
   (let* ((line (str:split #\Space (line-string position)))
-         (name (second line)))
+         (name (str:replace-all ")" ""  (second line))))
     (make-instance class
                    :reference-name name
                    :reference-point position)))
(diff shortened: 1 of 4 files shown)


Sasanidas:
Just when I finished doing this feature, I realize that @cxxxr had a similar idea about improving testing (https://github.com/lem-project/lem/pull/956/files), but I guess it has most of the job locally. So maybe we can combine both approaches.

This one is clearly more static, using the available buffer information. I think it can be combine with a more package-base one so it's possible to run any test define on any package (with a simple selections prompts).

Let me know what do you think :raised_hands:

src/ext/detective.lisp
```diff
@@ -214,16 +217,37 @@
           (buffer-modified-tick (current-buffer)))
     (search-references (variable-value 'lem/language-mode:detective-search :buffer)))))
 
-(defun %detective-move (&key (direction :up))
+(defun current-reference (&key (point (current-point)))
+  (let ((closest (%closest-reference))
+        (current-expression
+          (lem:with-point ((p point))
+            (funcall
+             (variable-value 'lem/language-mode::beginning-of-defun-function :buffer)
```
cxxxr, line 225: I see that beginning-of-defunctions are exported.

extensions/lisp-mode/detective.lisp
```diff
@@ -6,7 +6,7 @@
 
 (defun %default-capture (class position)
   (let* ((line (str:split #\Space (line-string position)))
-         (name (second line)))
+         (name (str:replace-all ")" ""  (second line))))
```
cxxxr, line 9: I have learned from this that you don't necessarily have to write a serious parser.

cxxxr (approved):
Thank you for the wonderful and PR!
I commented on this, but it is trivial and I would like to merge it.

PR #958, opened 2023-08-16
Fix test name when starting with # and small improvements

Sasanidas:
I was testing the commands with some Lem tests, and forgot about the `#` symbol, this should fix it.

I'm also addressing some of the comments https://github.com/lem-project/lem/pull/957#pullrequestreview-1580460203 tho I don't fully understand this one https://github.com/lem-project/lem/pull/957#discussion_r1295809648 :sweat_smile: 

Thanks!

### DIFF
--- a/extensions/lisp-mode/detective.lisp
+++ b/extensions/lisp-mode/detective.lisp
@@ -6,7 +6,7 @@
 
 (defun %default-capture (class position)
   (let* ((line (str:split #\Space (line-string position)))
-         (name (str:replace-all ")" ""  (second line))))
+         (name (remove #\) (second line))))
     (make-instance class
                    :reference-name name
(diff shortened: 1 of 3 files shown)


extensions/lisp-mode/detective.lisp
```diff
@@ -6,7 +6,7 @@
 
 (defun %default-capture (class position)
   (let* ((line (str:split #\Space (line-string position)))
-         (name (str:replace-all ")" ""  (second line))))
+         (name (remove #\) (second line))))
```
cxxxr, line 9: I realized later that it just need to fix this https://github.com/lem-project/lem/blob/main/extensions/lisp-mode/detective.lisp#L39.

Issue #434, opened 2019-09-29
Conform to XDG Base Directory Specification


kenoss:
https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html
https://qiita.com/lighttiger2505/items/440c32e40082dc310c1e#nvim-configuration

I prefer that lem forces users to place config files into `$XDG_CONFIG_HOME/lem`.

kenoss:
I mean "not selectable".  Config dir should be `$XDG_CONFIG_HOME/lem`, not `$XDG_CONFIG_HOME/lem` "and" `$HOME/.lem`.

appetrosyan:
I second this.

PR #960, opened 2023-08-17
Use XDG_CONFIG_HOME to find the lem configuration directory.

timmydo:
#434

### DIFF
--- a/src/config.lisp
+++ b/src/config.lisp
@@ -1,8 +1,11 @@
 (in-package :lem-core)
 
 (defun lem-home ()
-  (or (uiop:getenv "LEM_HOME")
-      (merge-pathnames ".lem/" (user-homedir-pathname))))
+  (let ((xdg-lem (uiop:xdg-config-home "lem/"))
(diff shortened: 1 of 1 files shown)


src/config.lisp
```diff
@@ -2,7 +2,7 @@
 
 (defun lem-home ()
   (or (uiop:getenv "LEM_HOME")
-      (merge-pathnames ".lem/" (user-homedir-pathname))))
+      (uiop:xdg-config-home "lem/")))
```
seanfarley, line 5: This should fall back to using `~/.lem` so that current users aren't caught off-guard, right?

Sasanidas:
Maybe we can support both? Like in GNU Emacs https://www.gnu.org/software/emacs/manual/html_node/emacs/Find-Init.html

src/config.lisp
timmydo: I changed the logic to take .lem if it exists first and fallback to xdg-config + 'lem/' if that doesn't already exist. Thoughts?

cxxxr:
Thank you!

PR #964, opened 2023-08-17
Fix new bugs of vi-mode introduced by #952

fukamachi:
## Bug fixed

* Fix `d%` to work regardless of whether the direction is forward or backward
* Fix `df<char>` and `dt<char>` not to delete if the search failed
* Support visual block mode to make `d` & `y` work with a rectangle region

ref #952


cxxxr:
Thank you!

PR #965, opened 2023-08-18
Add the base of tests for lem-vi-mode

fukamachi:
## Usage

* `with-test-buffer` creates a buffer object for testing
* `with-vi-buffer` to set the initial vi state and bind the given buffer (or string) to be the current buffer
  * Following functions are good to work with it:
    * `cmd`: Executes vi commands by key types in a string
      * ex) `h`, `G`, `<Esc>`, `3h`, `5j^D`, `<C-x>o`, `<H-Shift-x>`, `<M-Space>`.
    * `buf=`: Check if the current buffer text equals the given buffer's text, and the cursor position is right
    * `text=`: Check if the current buffer text equals the given buffer's text
    * `pos=`: Check if the current point is at the same position as the given point
    * `state=`: Check if the vi state is the given one.

#### Test definitions

`[x]` means a cursor that is on the character `x`. `<` and `>` means a region of vi's visual mode.

```common-lisp
(deftest vi-forward-char
  (with-fake-interface ()
    (with-vi-buffer (#?"[a]bcdef\n")
      ;; `l`: vi-forward-char
      (cmd "l")
      (ok (buf= #?"a[b]cdef\n"))
      ;; `l` accepts a universal argument
      (cmd "3l")
      (ok (buf= #?"abcd[e]f\n"))
      ;; `l` doesn't exceed the end of the line
      (cmd "10l")
      (ok (buf= #?"abcde[f]\n"))
      (cmd "hv^")
      (ok (buf= #?"<[a]bcde>f\n"))
      (cmd "vlV")
      (ok (buf= #?"<a[b]cdef\n>")))))))
```

#### Test outputs

```
Testing System lem-vi-mode/tests

;; testing 'lem-vi-mode/tests/motion'
vi-forward-char
  [buf] "[a]bcdef\n"
    [cmd] l
    ✓ Expect the buffer to be "a[b]cdef\n" (actual: "a[b]cdef\n")
    [cmd] 3l
    ✓ Expect the buffer to be "abcd[e]f\n" (actual: "abcd[e]f\n")
    [cmd] 10l
    ✓ Expect the buffer to be "abcde[f]\n" (actual: "abcde[f]\n")
    [cmd] hv^
    ✓ Expect the buffer to be "<[a]bcde>f\n" (actual: "<[a]bcde>f\n")
    [cmd] vlV
    ✓ Expect the buffer to be "<a[b]cdef\n>" (actual: "<a[b]cdef\n>")

✓ 1 test completed

Summary:
  All 1 test passed.
```


extensions/vi-mode/tests/utils.lisp
```diff
@@ -0,0 +1,298 @@
...
+                :apply-visual-range)
+  (:import-from :cl-ppcre
+                :scan
+                :regex-replace)
+  (:import-from :alexandria
+                :remove-from-plistf
+                :with-gensyms
+                :once-only
+                :appendf
+                :if-let)
```
cxxxr, line 34: It seems that there are symbols that have been imported but not used.
Can you check with `M-x lisp-organize-import`?

extensions/vi-mode/tests/utils.lisp
```diff
@@ -0,0 +1,298 @@
...
+                             &rest buffer-args
+                             &key name (temporary t temporary-specified-p)
+                             &allow-other-keys)
+                            &body body)
+  (declare (ignore temporary))
+  (unless temporary-specified-p
+    (setf (getf buffer-args :temporary) t))
+  (remove-from-plistf buffer-args :name)
+  (with-gensyms (point buffer-content position visual-regions)
+    `(with-fake-interface ()
```
cxxxr, line 186: I am a little concerned about the with-fake-interface being hidden inside this macro.
The person reading the test needs to dig into the macro to understand that this is using a fake-interface.
How about moving the use of with-fake-interface to the top level of testing?

cxxxr (commented):
The time spent reading this PR was very meaningful and educational.
I made some comments, but it is not MUST.

fukamachi:
@cxxxr
Thanks for your comments!
They're really helpful, and I applied all your advice :)

extensions/vi-mode/tests/utils.lisp
fukamachi: Why haven't I known this feature?
This must be one of the good reasons to recommend people to use Lem.

cxxxr:
Thank you so much!

PR #966, opened 2023-08-18
Add "iskeyword" option

fukamachi:
## Description

Add "iskeyword" of vi-mode option, which is used to specify a character set to be treated as a `word` by vi-mode.

* Type: a list of strings
* Default: `@,48-57,_,192-255`
* Alias: `isk`

## Usage

It can be set in ex mode, like `:set iskeyword=@,_`. Be sure it must be a comma-separated string in ex mode.

In this PR, several set operations like `+=`, `-=`, `^=` are also added, which can be used like `:set iskeyword+=@-@` to add a new value.

To set in `init.lisp`, a function `vi-option-value` is available:

```common-lisp
(import 'lem-vi-mode:vi-option-value)
(setf (vi-option-value "iskeyword") '("@" "_"))
```

### DIFF
--- a/extensions/vi-mode/ex-command.lisp
+++ b/extensions/vi-mode/ex-command.lisp
@@ -153,12 +153,22 @@
 
 (define-ex-command "^set?$" (range option-string)
   (declare (ignore range))
-  (multiple-value-bind (option-value option-name old-value isset)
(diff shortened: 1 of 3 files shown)


cxxxr:
Thank you!

PR #969, opened 2023-08-19
Fix vi-mode's word-related motions

fukamachi:
Fix vi-mode's word-related motions, like `e`, `E`, `w`, `W`, `b` and `B`.

The characters consist of `word` is determined by the `iskeyword` option, added by #966.


extensions/vi-mode/word.lisp
```diff
@@ -1,57 +1,75 @@
...
+      (t :non-word))))
+
+(defun broad-char-type (char)
+  (when char
+    (cond
+      ((blank-char-p char) :blank)
+      (t :word))))
+
+(defun forward-word-begin (char-type-fn &optional (point (current-point)))
+  (symbol-macrolet ((point-char (character-at point)))
```
cxxxr, line 40: This is a very cool way to write, but I am concerned that others will see this and copy it elsewhere.
What about using labels/flet?

cxxxr (commented):
Thanks for the great PR!
It's really good that they added the test.

extensions/vi-mode/word.lisp
fukamachi: Okay, I replaced them with simple `flet`s.
Actually, they didn't need symbol macros, I don't know I wrote them with it.

cxxxr:
Thank you so much!

PR #971, opened 2023-08-19
Rgrep command

Sasanidas:
So, grep it's an awesome command, but most of the time I just want a simple "search this thing at point", so this mimic in simplicty the GNU Emacs command https://www.gnu.org/software/emacs/manual/html_node/emacs/Grep-Searching.html

It can also be interesting to emulate the others grep options :+1:  

This command also uses a variable to store the default command (similar to GNU Emacs) so it can be customized by the user


cxxxr:
Hmm, If I want to search for a symbol, it seems cleaner to add project-search-symbol, etc. than to add an ad hoc grep command, but what do you think?

Sasanidas:
Ummm, the thing is, my idea is not to bound this command (or the grep command) to a project search, so I can search in whatever place, so in this case the naming wouldn't make that much sense. 

Also even if the function to get the "thing-at-point" is called symbol, the idea is to take whatever is at point (in every kind of buffer) so it doesn't necessarily have to be a symbol (as CL symbol I mean).

I don't think that rgrep is a great name either, because grep is already recursive, but I'm not great at naming :thinking:

cxxxr:
https://github.com/lem-project/lem/wiki/lem's-draft-vision%2C-goals%2C-non%E2%80%90goals
> Non goals
  Compatibility with GNU Emacs

so I need to be cautious about imitating emacs in the dark.

Sasanidas:
I see, then this maybe more suited for a user configuration, I'll close the issue :+1:

cxxxr:
I understand that there is a demand.
I hope to achieve this with a more comprehensive UI.

Sasanidas:
I mean the grep command works really well and the interface is awesome (way better than GNU Emacs by default) , so maybe we can come up with a more variety of commands :100:

PR #972, opened 2023-08-19
NeoVim-like status indicator for vi-mode

fukamachi:
I don't know if this is the appropriate way (and works in other environments) to set it up in Lem.

Screen Shot 2023-08-20 at 1 16 05
Screen Shot 2023-08-20 at 1 16 12
Screen Shot 2023-08-20 at 1 16 29
Screen Shot 2023-08-20 at 1 16 49


cxxxr:
That's so really cool!

PR #976, opened 2023-08-20
Switch between visual modes without resetting the starting point

fukamachi:
The vi-mode has several visual modes, `visual-char`, `visual-line` and `visual-block`. When one of those modes is enabled while the other one is one, the starting point should be taken over.


cxxxr:
Thank you!

PR #980, opened 2023-08-20
site-init: use lem-home instead of user's home directory

timmydo:
Currently this drops lem-site-init.asd into ~/.lem which causes lem-home to always use ~/.lem instead of XDG_CONFIG_HOME + lem

small whitespace fix in config.lisp #434


cxxxr:
Thank you!

Issue #630, opened 2023-05-23
Lem 2.0 crashes on SDL event (?), Linux Mint 21, 64 bit


maximinus:
Download Lem 2.0 from https://github.com/lem-project/lem/releases/download/v2.0.0/lem-linux-v2.0.0.tar.gz

Extracting and running ./lem opens a blank black window. Any keypress or mouse move will close the window and exit Lem.

The error.log in ./lem reports:

```
The value
  NIL
is not of type
  AUTOWRAP:WRAPPER
Backtrace for: #<SB-THREAD:THREAD tid=7276 "editor" RUNNING {1004CD9BB3}>
0: (AUTOWRAP:INVALIDATE NIL)
1: ((:METHOD LEM-SDL2::RESIZE (LEM-SDL2::VIEW T T)) #<LEM-SDL2::VIEW {10067E8DA3}> 50 1) [fast-method]
2: ((FLET "WITHOUT-INTERRUPTS-BODY-11" :IN SB-THREAD::CALL-WITH-RECURSIVE-LOCK))
3: (SB-THREAD::CALL-WITH-RECURSIVE-LOCK #<FUNCTION (FLET SB-THREAD::WITH-RECURSIVE-LOCK-THUNK :IN LEM-SDL2::CALL-WITH-RENDERER) {7FA949ACE1EB}> #<SB-THREAD:MUTEX "lem-sdl2 display mutex" taken owner=editor> T NIL)
4: (LEM-SDL2::CALL-WITH-RENDERER #<FUNCTION (LAMBDA NIL :IN LEM-INTERFACE:SET-VIEW-SIZE) {10067F2BFB}>)
5: (LEM-SDL2::CALL-WITH-DEBUG #<FUNCTION (LAMBDA NIL :IN LEM-INTERFACE:SET-VIEW-SIZE) {10067F2B9B}> #<FUNCTION (LAMBDA NIL :IN LEM-INTERFACE:SET-VIEW-SIZE) {10067F2BCB}>)
[...]

PR #981, opened 2023-08-21
sdl2: Run paint operations on SDL2 main thread

timmydo:
This seems to solve my issues with SDL2 #630. Could someone else test it out? I tested running from sly/emacs and from sbcl on the command line. GNU Guix with Sway (no X support).


cxxxr:
Seems to work on mac.

cxxxr:
Thank you so much.
We'll merge them once and think about it if we run into problems.

PR #984, opened 2023-08-21
Fix legit select path find

Sasanidas:


src/ext/legit/peek-legit.lisp
```diff
@@ -320,7 +320,13 @@ Notes:
...
   (alexandria:when-let ((file (get-matched-file)))
     (quit)
-    (alexandria:when-let ((buffer (find-file-buffer file)))
+    (alexandria:when-let
+        ((buffer (or (and (uiop:file-exists-p file)
+                          (find-file-buffer file))
+                     (find-file-buffer
+                      (format nil "~a~a"
+                              (lem-core/commands/project:find-root (buffer-filename))
+                              file)))))
```
cxxxr, line 329: Using merge-pathnames seems to be a good idea.

src/ext/legit/peek-legit.lisp
```diff
@@ -320,7 +320,13 @@ Notes:
 (define-command peek-legit-select () ()
   (alexandria:when-let ((file (get-matched-file)))
     (quit)
-    (alexandria:when-let ((buffer (find-file-buffer file)))
+    (alexandria:when-let
+        ((buffer (or (and (uiop:file-exists-p file)
```
vindarel, line 324: (I don't really know the context for this fix)

In legit, pressing Enter in the legit window calls this function to visit the file at point. There was a file listing, the file should exist. So if the file doesn't exist anymore, we maybe should display a message to the user (instead of doing nothing)? (and refresh the legit window?)

Also I thought the `%filename` slot of a buffer had the full path, not requiring to get the project root?

exple:

    [ ]  %FILENAME       = "/home/vince/lisp-projects/lem/legit.lisp"

Sasanidas: I'm not sure, but seems like the path is indeed not correct:

https://github.com/lem-project/lem/assets/26670956/837283d4-a1a2-4388-aebc-03184462c592

Maybe the buffer doesn't have the information about the absolute path? :thinking:

src/ext/legit/peek-legit.lisp
Sasanidas: Added :+1:

src/ext/legit/peek-legit.lisp
vindarel: I can't reproduce. Do you try to visit an unstaged file?

- start from directory mode
- move cursor on an "unstaged changes" line
- press Enter

? this works for me, I visit the file at point. I can stage it with "s" too.

Sasanidas: Yes I did just that :thinking:

cxxxr:
What is the state of this?

vindarel:
> What is the state of this?

I myself can't reproduce, I have yet to try more, 

and I have an open improvement suggestion in my comment:

> if the file doesn't exist anymore, we maybe should display a message to the user (instead of doing nothing)?

Sasanidas:
I added a message when the file cannot be found :+1:

src/ext/legit/peek-legit.lisp
```diff
@@ -320,14 +320,15 @@ Notes:
...
+    (alexandria:if-let
         ((buffer (or (and (uiop:file-exists-p file)
                           (find-file-buffer file))
                      (find-file-buffer
                       (merge-pathnames
                        (lem-core/commands/project:find-root (buffer-filename))
                        file)))))
-      (switch-to-buffer buffer))))
+      (switch-to-buffer buffer)
+      (message "File ~a doesn't exist." file))))
```
cxxxr, line 331: Elsewhere, we use editor-error to display errors.
Is it possible to do that here as well?

Sasanidas: Done :+1:

cxxxr:
Thank you!

Issue #987, opened 2023-08-21
Break error when evaluating a line using `lisp-eval-at-point`


Sasanidas:
When using the function `lisp-eval-at-point` (or the key combination of it) it breaks the editor entering an eternal loop:

![fermin@fdebian: ~-quicklisp-local-projects-lem_001](https://github.com/lem-project/lem/assets/26670956/39113348-4b8b-484a-af92-69b326056687)

It seems like it has something to do with the spinner (maybe related to the new refactor overlay? https://github.com/lem-project/lem/pull/983)

Ping @cxxxr

masatoi:
It seems trivial bug. I made PR #988 for this.

cxxxr:
I think I broke it in https://github.com/lem-project/lem/pull/983

PR #988, opened 2023-08-21
Fix error in start-loading-spinner

masatoi:
Fix for https://github.com/lem-project/lem/issues/987

- nil is passed to `update-line-spinner`
- The `temporary` slot of `overlay-line-endings` is unbound, causing an error in initialize-instance


cxxxr:
Thank you.
I think I broke it in https://github.com/lem-project/lem/pull/983

PR #990, opened 2023-08-23
Fix vi-join-line to work for a line longer than the window width

fukamachi:


cxxxr:
It seems to be conflicted.

PR #991, opened 2023-08-23
Treat modifications while in INSERT mode as a single undo stack

fukamachi:


extensions/vi-mode/tests/utils.lisp
```diff
@@ -187,7 +189,9 @@
       (multiple-value-bind (buffer-text position visual-regions)
           (parse-buffer-string content)
         (let ((point (buffer-point buffer)))
-          (insert-string point buffer-text)
+          (lem:insert-string point buffer-text)
+          (setf (lem-base::buffer-edit-history buffer)
+                (make-array 0 :adjustable t :fill-pointer 0))
```
cxxxr, line 194: It may be a good idea to prepare an API for buffer-edit-history and use that.

cxxxr:
Thank you!

PR #992, opened 2023-08-23
Change vi-mode to a global minor mode to make its keybinds' priority high

fukamachi:
This fixes that `Return` adds a newline even in NORMAL mode with lisp-mode.


fukamachi:
Buffer-local keymap and buffer-created hook are required to change keybindings for each mode flexibly.

fukamachi:
Without them, it can be achieved by overwriting `execute` method, like:

```common-lisp
(defmethod execute ((mode directory-mode) (command vi-return) uarg)
  (declare (ignore uarg))
  (directory-mode-find-file))
```

However, `M-x describe-key <Return>` still says it's bound to `vi-return`, and all commands have to be overwritten.

fukamachi:
Will be fixed in a different way at #1008.

PR #994, opened 2023-08-23
Fix bugs of vi-mode's line search commands

fukamachi:
Fix bugs of vi-mode's line search commands such as `f`, `F`, `t` and `T`.

* Fix them to search by case-sensitive
* Make them allow to take a universal argument


cxxxr:
There are tests and they are very good.

PR #996, opened 2023-08-23
Fix vi-mode search feature to work similar to Vim

### LINKED ISSUE #314: [vi-mode] The cursor position after `n` is wrong
I expect `n` moves the cursor to the start of the next match after the each search commands, but to the end of it.
`N` works all right.

fukamachi:
* Fix the cursor position when searching forward (`/`), moving to the next match (`n`)
* Fix the searching direction of `n` and `N` when searching backward (`?`)
  * `n` should search backward if the last search was backward

fixes #314


src/ext/isearch.lisp
```diff
@@ -383,9 +395,8 @@
         (concatenate 'string
                      *isearch-string*
                      (string c)))
-  (with-point ((start-point (current-point)))
-    (unless (funcall *isearch-search-function* (current-point) *isearch-string*)
-      (move-point (current-point) start-point)))
+  (unless (funcall *isearch-search-function* (current-point) *isearch-string*)
+    (move-point (current-point) *isearch-start-point*))
```
fukamachi, line 399: I think the current behavior is not intended that the cursor point doesn't go back when the search fails.

src/ext/isearch.lisp
```diff
@@ -347,30 +347,42 @@
...
+                 (if (funcall *isearch-search-forward-function* p *isearch-string*)
+                     (progn
+                       (move-point (current-point) p)
+                       t)
+                     nil)))
+              ((not (funcall *isearch-search-forward-function* (current-point) *isearch-string*))
+               (setf (variable-value 'isearch-next-last :buffer) t)
+               nil)
+              (t))
+      (isearch-update-display))))
```
fukamachi, line 364: Changed to return a boolean value to tell whether the search is successful.

src/ext/isearch.lisp
```diff
@@ -347,30 +347,42 @@
...
+                 (buffer-end p)
+                 (if (funcall *isearch-search-backward-function* p *isearch-string*)
+                     (progn
+                       (move-point (current-point) p)
+                       t)
+                     nil)))
+              ((not (funcall *isearch-search-backward-function* (current-point) *isearch-string*))
+               (setf (variable-value 'isearch-prev-last :buffer) t)
+               nil)
+              (t))
```
cxxxr, line 384: This returns T. Is this intended?

fukamachi: Yep. I changed this to return a boolean value as I did to `isearch-next`.

src/ext/isearch.lisp
cxxxr: Ah, I see.
if you write the following, the current-point might not be broken when interrupted.

```common-lisp
(with-point ((point (current-point)))
  (when (funcall *isearch-search-function* point *isearch-string*)
    (move-point (current-point) point)))
```

src/ext/isearch.lisp
cxxxr: Oh, I understand now 👍

src/ext/isearch.lisp
fukamachi: Oh, that looks better. I'll rewrite it as you suggested. Thanks!

fukamachi: Sorry, this has to be the code I originally changed.

For example, when I have a buffer like the following:

```
[a]pple orange grape
``` 

When starting ISearch with `C-s lisp`, then the cursor moves to `app[l]e` even though the search is failed.

It's better to go back to the starting point when the search has failed. How do you think?

fukamachi: PR #1002

PR #1004, opened 2023-08-25
Add a feature to set high priority keymaps for emulation modes

fukamachi:
Add a new variable `*emulation-mode-keymaps-functions*` to set keymaps that precede other major/minor modes. This is equivalent to Emacs's `emulation-mode-map-alists` for emulation mode.

ref. [Controlling Active Maps | Emacs manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Controlling-Active-Maps.html#index-emulation_002dmode_002dmap_002dalists)

For the time being, `vi-mode` is in mind. It requires switching multiple keymaps by its "state" and has to _merge_ with others to work better in other modes.

Talking about preceding lisp-mode's keybinds, and problems with directory-mode at #992 .

### DIFF
--- a/src/internal-packages.lisp
+++ b/src/internal-packages.lisp
@@ -365,6 +365,7 @@
    :abort-key-p
    :with-special-keymap
    :traverse-keymap
+   :compute-keymaps
    :collect-command-keybindings)
   ;; reexport common/timer
   (:export
--- a/src/keymap.lisp
+++ b/src/keymap.lisp
@@ -143,11 +143,16 @@
                 (match-key key :sym sym))
            (char sym 0)))))
 
+(defgeneric compute-keymaps (global-mode)
+  (:method ((mode global-mode)) nil))
+
 (defun all-keymaps ()
-  (let ((keymaps
-          (loop :for mode :in (all-active-modes (current-buffer))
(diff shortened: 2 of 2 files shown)


fukamachi:
The other option to achieve the demand for vi-mode is allowing multiple keymaps from modes, something like `mode-keymaps`.
However, it's rare to need this kind of feature, so adding this to all modes is overkill and will negatively influence Lem's performance.

cxxxr:
Hmm, I would prefer not to add more special variables.
How about separating the methods as follows?

```common-lisp
(defmethod compute-keymaps ((global-mode emacs-mode)) ...)
(defmethod compute-keymaps ((global-mode vi-mode)) ...)

(defun all-keymaps ()
  ...
  (let ((keymaps (compute-keymaps (current-global-mode))))
    ...
  ))
```

This way, I feel it will be easier to expand in the future.

fukamachi:
Thanks for your comment.
It'd be fine for me if it can return multiple keymaps that precede others.
I think `emulation-mode-map-alists` is Emacs's way, as Emacs Lisp doesn't have generic methods.

cxxxr:
Thank you.
Personally, I feel this is more Common Lisp like.

PR #1008, opened 2023-08-26
Precede vi-mode keymaps over major-mode.

fukamachi:
Not allowed to insert newlines while in the normal state on lisp files.

ref #992, #1004


cxxxr:
Thank you!

PR #1012, opened 2023-08-27
Add `diw` and `daw` to vi-mode

fukamachi:
Add operator-pending state and text-objects to vi-mode, and add `diw` and `daw` as the first step.

`diw` and `daw` are commands to delete the word under the cursor. Also works with other operators, like `ciw` or `yaw`.

It does not work in visual state yet to expand the visual region. It will be added in the later changes.


cxxxr:
Thank you!

PR #1029, opened 2023-08-30
Fix vi-mode's visual rectangle region 

fukamachi:
Fix the vi-mode's visual block region when the cursor moves to right-top or left-bottom.

### DIFF
--- a/extensions/vi-mode/visual.lisp
+++ b/extensions/vi-mode/visual.lisp
@@ -95,13 +95,17 @@
 (defmethod state-setup ((state visual-block))
   (with-point ((start *start-point*)
                (end (current-point)))
-    (when (point< end start)
-      (rotatef start end))
-    (character-offset end 1)
     (let ((start-column (point-column start))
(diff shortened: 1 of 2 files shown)


extensions/vi-mode/visual.lisp
```diff
@@ -95,13 +95,17 @@
...
-        (rotatef start-column end-column))
+      (if (< end-column start-column)
+          ;; left-top or left-bottom
+          (progn
+            (character-offset start 1)
+            (incf start-column))
+          ;; right-top or right-bottom
+          (progn
+            (character-offset end 1)
+            (incf end-column)))
```
cxxxr, line 108: If possible, `cond` would be better in these cases.

https://google.github.io/styleguide/lispguide.xml?showone=Conditional_Expressions#Conditional_Expressions

>  However, don't use PROGN for an IF clause — use COND, WHEN, or UNLESS.

cxxxr:
Thanks!

PR #1033, opened 2023-08-31
Add replace-state to vi-mode

fukamachi:
Add replace-state to vi-mode that is on when `r` is pressed. It allows quitting the replace-state by `Esc` or `C-g`.

As Vim's replace state provides other special commands like `C-e` and `C-y`, this flexibility would be required for the future extension.

### DIFF
--- a/extensions/vi-mode/states.lisp
+++ b/extensions/vi-mode/states.lisp
@@ -17,11 +17,13 @@
            :*insert-keymap*
            :*inactive-keymap*
            :*operator-keymap*
+           :*replace-state-keymap*
            :*outer-text-objects-keymap*
            :*inner-text-objects-keymap*
            :normal
            :insert
-           :operator))
+           :operator
+           :replace-state))
 (in-package :lem-vi-mode/states)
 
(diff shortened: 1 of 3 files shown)


extensions/vi-mode/states.lisp
```diff
@@ -37,11 +39,16 @@
...
 (defvar *operator-keymap* (make-keymap :name '*operator-keymap*))
+(defvar *replace-state-keymap* (make-keymap :name '*replace-state-keymap*
+                                            :undef-hook 'return-last-read-char))
 (defvar *outer-text-objects-keymap* (make-keymap :name '*outer-text-objects-keymap*))
 (defvar *inner-text-objects-keymap* (make-keymap :name '*inner-text-objects-keymap*))
 
 (defvar *inactive-keymap* (make-keymap))
 
+(define-command return-last-read-char () ()
+  (key-to-char (first (lem-core:last-read-key-sequence))))
```
cxxxr, line 50: If a sequence has more than one length, is it not necessary to get the last?

cxxxr:
Thank you!

PR #1034, opened 2023-08-31
Add a feature to disable undo boundary temporarily

fukamachi:
This PR allows users to enable/disable auto undo boundary addition temporarily.

This feature is required in vi-mode to take all modifications done while in INSERT mode as a single undo stack.

I made it in a *hacky* way in vi-mode (ref #991), but it's not perfect since Lem automatically adds an undo boundary after every command invocation. For example, Lem adds a boundary even after completion.

Take this as a feature request with a code. I'm fine if the feature is done in a different way. Any advice is welcome.


cxxxr:
Thank you for the nice PR.
In the future, undo management may need to be improved, but for now, this is a good idea.

PR #1039, opened 2023-08-31
Make `gg` allow to take a uarg

fukamachi:


extensions/vi-mode/commands.lisp
```diff
@@ -579,21 +579,25 @@
...
     (:type :line
      :jump t)
-  (if (null n)
-      (progn
-        (move-to-end-of-buffer)
-        (line-start (current-point)))
-      (goto-line n))
-  (skip-whitespace-forward (current-point) t))
+  (let ((col (point-charpos (current-point))))
+    (if (null n)
```
cxxxr, line 593: If possible, please use `cond`.

fukamachi: I can't entirely agree that it improves readability in this context, but that's fine if you like it.

Issue #997, opened 2023-08-24
notmuch-mode


timmydo:
I'm currently using Emacs and notmuch (https://notmuchmail.org/) to read my email. I was interested in trying to port this UI from emacs to Common Lisp/Lem. 

Notmuch is basically a command line tool that indexes a maildir into a xapian DB and lets you run commands on it. It comes with an emacs mode which provides a UI. Code: https://git.notmuchmail.org/git?p=notmuch;a=tree;f=emacs;h=ae8088e5ef4543c53edb6e2e24bf368a64097993;hb=HEAD.

If I was to make a notmuch package for Lem, what's the right way to host it? Would it be an external package or would it be in the Lem git repository?

Does lem have the right primitives for creating a UI for reading/writing mail?

vindarel:
> Does lem have the right primitives for creating a UI for reading/writing mail?

Hello, I'd say yes, because Lem has many primitives to display and write text. You can use plain buffers or windows (like the grep interface), you can associate attributes to text and run a command when a key is pressed on this text depending on the attribute (buttons, essentially) (grep, legit interface), you have completion windows, window popups, etc.

Lem has no text folding yet (it's in the works), same for a spell checker.

timmydo:
I was able to get some very basic stuff in a couple of hours this weekend... https://github.com/lem-project/lem/commit/c7af6159c155c7eaefca82bda59bb6f37bd8b61a

![image](https://github.com/lem-project/lem/assets/1028829/545d7140-0bf6-4943-a15a-84fc68dbd931)
![image](https://github.com/lem-project/lem/assets/1028829/a725f1ea-603d-49ea-90c4-c03e675d835f)

It seems lem-button is nice but I'd like to be able to pass all the context I have to the same method call when I press Return on it in the buffer.

vindarel:
nice!

> in a couple of hours

that's the magic of Lem. I felt the codebase very approachable, discoverable, and I felt very productive when I dived into writing my first module.

> lem-button

I actually didn't know the lem/button package O_o so can't help here. However, when hearing "full context", I think "use a closure?".

timmydo:
Regarding closures, I think that would work for lem-button's click event but I'm not sure how to invoke the click event from a key press in the buffer when the key press happens over the button. If I could do that, then I wouldn't need two different code paths for invoking it. It's probably possible didn't see stumble on it when I was working on it. Will take a look next time I get a chance.

PR #1048, opened 2023-09-02
Notmuch support (work in progress)

timmydo:
Creating a draft PR for #997 

Still a ways to go. Will squash the commits later.

### DIFF
--- a/contrib/notmuch/README.md
+++ b/contrib/notmuch/README.md
@@ -1,0 +1,12 @@
+This is a module that adds some notmuch modes for reading email with notmuch.
+
+
+** Installation
+
+To use:
+`M-x site-init-add-dependency`
+then type `notmuch`
+
+this will it to your lem-site-init.asd
+
+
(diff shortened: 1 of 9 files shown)


vindarel:
Hello, what are the news of this notmuch mode? Best,

timmydo:
Thanks for following up. I can abandon this. I wasn't using Lem for a while and just tried it again last month when I had some free time during winter break. I reorganized the code I had into a single file (https://git.sr.ht/~timmydo/lem-notmuch/tree/main/item/notmuch.lisp) and was just loading it like this:

```
(load "/home/timmy/src/lem-notmuch/package.lisp")
(load "/home/timmy/src/lem-notmuch/notmuch.lisp")
```

Does lem have something like melpa? I don't think this sort of add-on should live in its source tree

vindarel:
no there is no registry for packages yet (a lot of stuff goes in^^).

PR #1055, opened 2023-09-04
Include the last newline when kill/copying linewise.

fukamachi:


extensions/vi-mode/commands.lisp
```diff
@@ -275,21 +275,21 @@
...
-            (max 0
-                 (min (1- (length (line-string (current-point)))) pos))))
+               (not ends-with-newline)
+               (not (= (position-at-point start) 1)))
+      (delete-previous-char))
     (when (eq 'vi-delete (command-name (this-command)))
+      (when (eq type :line)
+        (setf (point-charpos (current-point))
+              (max 0
+                   (min (1- (length (line-string (current-point)))) pos))))
```
cxxxr, line 292: Direct change of point-charpos is deprecated because it breaks encapsulation.
How about `(line-offset (current-point) 0 charpos)`

PR #1061, opened 2023-09-05
Fix to work Ex command for visual range.

fukamachi:


cxxxr:
It seems to be conflict.

PR #1063, opened 2023-09-05
Created simplified build scripts and instructions

blakemcbride:
Created simplified build and update scripts.  Also updated README.md to reflect these.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,8 @@
+
+**You do not want to clone this repo.  It is an expermental clone of https://lem-project.github.io/lem-page/
+Clone that project.**
+
+
 # Lem is the editor/IDE well-tuned for Common Lisp.
 
 ![](https://github.com/Shinmera/lem-icon/blob/gh-pages/icon-blue.svg)
--- a/README.md
+++ b/README.md
@@ -91,25 +91,27 @@ You can watch the screencast on Youtube.
 
 ## Installation with sbcl
 
-Please clone lem to a location where the path to asdf is accessible.
+The following assumes SBCL has been installed and nothing else.
 
 ```
-$ mkdir $HOME/common-lisp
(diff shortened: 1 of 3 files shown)


cxxxr:
Thank you for the PR.
I am not comfortable merging this change as I feel it is a bit adhok.
But I am very happy that you have tried to improve lem.

blakemcbride:
Initially, I had trouble building lem.  Now that I understand the steps, it's easy.  However, getting over that initial hurdle nearly lost me.  The script I provided automates and informs those steps.  If, you're not interested in the scripts, would you be interested in README.md updates that described (what was for me) missing information?
Thanks!

cxxxr:
Thank you for your feedback.
I am not sure it is necessary to support setup of a Common Lisp environment that is not directly related to lem.
But it might be worth mentioning in the README.

seanfarley:
This is also something that I'm currently working on (but I keep getting distracted by real life 😭) so that we can easily compile, build, and sign binary assets. One day I'll finish 😿

Sasanidas:
Thanks for the contribution and interest in Lem!

I have a few comments:
- So, it's interesting this approach, but why use these scripts instead of the Makefile?
- There is a simple chain of files in scripts mostly because of the separation of dependencies and the use of quick-patch (which solves some dependencies issues).
- I don't think the installation of quicklisp should be address by Lem, (maybe a separate script to check if it is installed, but not in the raw compilation of the executable)

So, in general, I think the roswell approach should cover some of these issues, or otherwise the Makefile should do the work.

Regards

Sasanidas:
Maybe improve the comment here (https://lem-project.github.io/lem-page/installation/getting-started/#installation-with-sbcl), changing it so quicklisp is required for lem to work. :thinking:

Issue #827, opened 2023-07-10
Add a command to toggle line-wrap


cxxxr:
### Discussed in https://github.com/lem-project/lem/discussions/826

<div type='discussions-op-text'>

<sup>Originally posted by **aykaramba** July 10, 2023</sup>
Hi.

I have been reading this page: https://github.com/lem-project/lem/releases

For [v1.9.2] it says: Rename editor variable truncate-lines to line-wrap.

When I M-x line-* I don't see line-wrap as an option to autocomplete.  Does this still exist?  Is there an alternate options that I don't understand?

Thanks.</div>

jfaz1:
@cxxxr safe to close this one

PR #1067, opened 2023-09-08
Add command to toggle line wrap in the buffer

resttime:
Addresses the issue: #827

Toggles line wrap in buffer only, unsure if should be global or not

### DIFF
--- a/src/commands/window.lisp
+++ b/src/commands/window.lisp
@@ -301,3 +301,7 @@
                      (not (eql c1 c2)))
           :while (and (character-offset p1 1)
                       (character-offset p2 1)))))
+
+(define-command toggle-line-wrap () ()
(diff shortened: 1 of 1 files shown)


cxxxr:
Thank you for the PR!
By the way, the standard command will automatically create a document when you put it in src/commands/.
https://github.com/lem-project/lem/blob/main/docs/default-keybindings.md

Would src/commands/window.lisp be better in this case?

resttime:
No problem! If I'm not mistaken that's the location where it's in right now. I put in there after seeing `line-wrap` defined above. Another possible location it could fit might be `src/commands/buffer.lisp`

cxxxr:
Ah, I mistake.
Thank you so much!

PR #1069, opened 2023-09-08
lem-sdl2: Make configurable modifier keycode by dynamic variable

masatoi:
On SDL2 / Linux, when the meta key is assigned separately from the alt key, the meta key is not recognized as meta but super.
Currently, it is hard-coded which modifier key is actually assigned to which modifier.
This PR introduces a dynamic variable `*modifier-code-table*` that allows modifier keys to be set.
By default, the behavior is the same as before.

The following can be written in the configuration file.
```lisp
#+lem-sdl2
(setf lem-sdl2/keyboard::*modifier-code-table*
      `((:shift ,sdl2-ffi:+kmod-lshift+ ,sdl2-ffi:+kmod-rshift+)
        (:ctrl ,sdl2-ffi:+kmod-lctrl+ ,sdl2-ffi:+kmod-rctrl+)
        (:meta ,sdl2-ffi:+kmod-lgui+ ,sdl2-ffi:+kmod-rgui+)
        (:super ,sdl2-ffi:+kmod-lalt+ ,sdl2-ffi:+kmod-ralt+)))
```

### DIFF
--- a/frontends/sdl2/keyboard.lisp
+++ b/frontends/sdl2/keyboard.lisp
@@ -90,18 +90,23 @@
         (modifier (get-modifier keysym)))
     (make-key-event code modifier)))
 
-(defun mod-p (mod value)
-  (= value (logand value mod)))
+(defparameter *modifier-code-table*
+  `((:shift ,sdl2-ffi:+kmod-lshift+ ,sdl2-ffi:+kmod-rshift+)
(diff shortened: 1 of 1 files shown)


rpx99:
Think ralt should not be in here, c.f. :
https://github.com/lem-project/lem/commit/e56796b3fc1d86f187639e2f3745b6aaa66f3d2f

masatoi:
Yeah, it does so by default.

cxxxr:
Thank you!

Issue #1072, opened 2023-09-08
Directory mode bug: deleting soft link is dangerous!


rsm-lisper:
In directory mode, when I delete soft link leading to directory, it does not delete soft link. Soft link stays as it was. But instead lem goes to target directory and **deletes its content**! :open_mouth:

**Before:**
![softlink0](https://github.com/lem-project/lem/assets/26094109/06d2546f-fc3e-4500-bcd8-76fd687cb5ab)

**During:**
![softlink1](https://github.com/lem-project/lem/assets/26094109/560fef93-2351-4bd6-91cc-719e6d333608)

**After:** :unamused:
![softlink2](https://github.com/lem-project/lem/assets/26094109/7c6ef9ab-3789-4f4d-b3cb-152ee877dd67)

Sasanidas:
Oh my! This is indeed a big bug, thanks for the report! :alarm_clock:

Sasanidas:
So, I added a fix that I think it does address the issue, the problem was that the command that was trigger for alll the deletions on directory-mode was ` (run-command `("rm" "-fr" ,file))`, with the r option enable, completely ignores the symlinks and kills the original host content, like you described a really undesired behavior! :face_exhaling:

PR #1074, opened 2023-09-08
Add safeguard for deleting symlinks

Sasanidas:
I think this address https://github.com/lem-project/lem/issues/1072 , the rm command was deleting the symlink directory content which is not the desire behaviour and can cause serious problems (specially when rm doesn't ask any question).


Sasanidas:
I didn't modify the windows version because I cannot test it properly, sorry.

cxxxr:
Thank you.

PR #1078, opened 2023-09-09
Ensure to reset the universal arguments even when error occurs

fukamachi:
The universal argument isn't reset when some error occurs in the command called in `universal-argument-default`.

This happens quite often with vi-mode for errors like `end-of-buffer`. For example, `10000j` ends with an `end-of-error`, and can see the remaining universal argument with `1j`.


cxxxr:
Thank you

PR #1079, opened 2023-09-10
Add `<Leader>` key to vi-mode.

fukamachi:
Vim provides a special key `<Leader>`, which is configurable by each user.

For example, an extension provides a command mapped to `<Leader> x`, then it's invoked by `\` + `x` since `Backslash` (`\`) is the default map of `<Leader>`. If the user set `g:mapleader` to `Space`, it can be invoked by `Space x`.

## Usage

```common-lisp
(define-command say-hello () () (message "Hello!"))
(define-key lem-vi-mode:*normal-keymap* "Leader x" 'say-hello)
(setf (variable-value 'lem-vi-mode:leader-key) "Space")
```

## Lem API changes

* Add `define-named-key` to define a custom named key
* Rewrite `keymap-find-keybind` as a generic function, and export it

### DIFF
--- a/extensions/vi-mode/core.lisp
+++ b/extensions/vi-mode/core.lisp
@@ -42,7 +42,9 @@
            :text-object-abort
            :text-object-abort-range
            :vi-current-window
-           :with-main-window))
+           :with-main-window
+           :vi-keymap
+           :define-keymap))
 (in-package :lem-vi-mode/core)
 
 (defvar *last-repeat-keys* '())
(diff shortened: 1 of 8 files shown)


cxxxr:
The test appears to be failing with the new changes.

fukamachi:
Sorry, that was a simple bug due to a recent renaming :P

cxxxr:
It looks very good.
Thank you.

PR #1080, opened 2023-09-10
Recognize links in buffers

cxxxr:
File paths and URLs can now be treated as links on buffers with link-mode enabled.

## Usage
On the link `M-.` or `M-x link-open`.

## Demo

https://github.com/lem-project/lem/assets/13656378/602f4e22-5471-4631-91c9-ed41cb2661bc

https://github.com/lem-project/lem/assets/13656378/6939397f-a62b-426e-a9be-fce4481b31ae


Sasanidas:
So interesting, but I do think that we may have duplicate functionality with open-at-point, I'll add some comments :+1:

Sasanidas:
Maybe the open-link functionality can be integrated into open at point, as a cond option :thinking:

Sasanidas:
The idea of thing-at-point it's to have just one command to execute a valid action base on the context and the string that's under the point. This new functionality seems the same, but with a different name. What do you think? :bulb:

cxxxr:
If so, it would be better to add a more appropriate function than the name "open-", since the action could be "press the button".

Sasanidas:
> If so, it would be better to add a more appropriate function than the name "open-", since the action could be "press the button".

That maybe true, but I'm not so great at creating names, so just adapt the GNU Emacs name

Sasanidas:
Could be something like, "interact-at-poin" or similar

Sasanidas:
I think you modified the behaviour of open-at-point

Sasanidas:
It doesn't go to the definition of the symbol anymore, maybe a bug?

Sasanidas:
https://github.com/lem-project/lem/assets/26670956/d8267da5-e30c-4840-ac6d-a690655fee16

Sasanidas:
Also, why remove the tests instead of just addapting it :thinking:

cxxxr:
> Also, why remove the tests instead of just addapting it 🤔

The reason is that the test was about path and url, which was resolved by using quri.

> It doesn't go to the definition of the symbol anymore, maybe a bug?

Hmmm, I guess I didn't take this into account enough.

Sasanidas:
I guess to mimic to functionality of thing at point it would have to:

- Go to the webpage link (when it's a webpage)
- Go to the definition of a symbol if it's a symbol (this was the last option)
- Open a file int a new buffer when it's a file at point and it exist

This was overall the functionality of thing-at-point :+1:

PR #1101, opened 2023-09-21
legit: write commit message in a buffer

vindarel:
Currently, we can only write one-line commit messages in Lem/legit :S This improves the situation, we create a special buffer to write commit messages.

I only tested this evening while working on it.

### DIFF
--- a/src/ext/legit/legit-commit.lisp
+++ b/src/ext/legit/legit-commit.lisp
@@ -1,0 +1,125 @@
+(in-package :lem/legit)
+
+#|
+Done:
+
+- "c" opens a commit message window on the right side, we can type a long message,
+- lines starting by a # are ignored.
+
+TODOs:
+
+- add C-c C-s to sign the message
+- and other shortcuts and features
+
+Future:
+
+- save previous messages, add C-p C-n commands to show them
+- find related Github issue when writing "fixes #"
+
+|#
+
+(define-major-mode legit-commit-mode lem-markdown-mode:markdown-mode
+    (:name "legit-commit-mode"
+     :syntax-table lem-markdown-mode::*markdown-syntax-table*
+     :keymap *legit-commit-mode-keymap*)
+  ;; no syntax highlihgt in fact.
+  (setf (variable-value 'enable-syntax-highlight) t))
+
+;; Validate, abort.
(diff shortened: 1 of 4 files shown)


cxxxr:
I think this is a good change.
Is it still a draft PR?

vindarel:
It is ready for those who want to keep testing lem/legit, but by looking at it I see a couple inline tests that you might not like :D (I like them… these will also be useful again shortly)

cxxxr:
Yeah, I think the inline comment are not bad.
The only reason I thought it was a draft is that the new file was not added to asd and not loaded.

vindarel:
hello, ping?

src/ext/legit/legit-commit.lisp
```diff
@@ -0,0 +1,125 @@
...
+" (clean-commit-message
+"one
+
+two
+
+# Please enter the commit message for your changes. Lines starting
+# with '#' will be ignored, and an empty message aborts the commit.
+")))
+
+(defun commit-continue ()
```
cxxxr, line 90: This is bound to C-c C-c, but it is not defined using define-command, so it cannot be called.

src/ext/legit/legit-commit.lisp
```diff
@@ -0,0 +1,125 @@
...
+         (lem-core/commands/window:previous-window)
+         ;; and refresh.
+         (legit-status))))))
+
+(define-command commit-abort () ()
+  (kill-buffer "*legit-commit*")
+  (lem-core/commands/window:previous-window))
+
+(define-command commit-abort-yes-or-no () ()
+  ;; TODO: prompt for confirmation.
```
cxxxr, line 112: Can we use prompt-for-y-or-n-p?

cxxxr (commented):
This feature is very nice!
I tried it for a while and commented on a few things that caught my attention.

src/ext/legit/legit-commit.lisp
vindarel: \+1 good catch thanks. Do we have a prompt where Y is the default choice? (so we can only press Enter, not "y")

cxxxr (approved):
LGTM

PR #1105, opened 2023-09-25
git: remove last of submodules

seanfarley:
This pull request removes the last of the git submodules and turns lem into a single repo and back to being distributed :-)

We can justify this change because lem is just a consumer of language-server-protocol and not a distributor of it.


cxxxr:
Indeed, this might be better.
What kind of operation is involved when bringing language-server-protocol up to date?

seanfarley:
> Indeed, this might be better.
> What kind of operation is involved when bringing language-server-protocol up to date?

It's pretty much how you'd expect:

```
$ git rm -r language-server-protocol
$ cp -r /path/to/language-server-protocol-version language-server-protocol
$ git add language-server-protocol
```

Then commit with a message about which version you're updating to.

cxxxr:
Hmm, still, please wait a moment.
The bundled stuff is so huge that, language-ratio-wise, this repository is no longer Common Lisp.

cxxxr:
Rather than bundling large amounts of unnecessary code, it may be better to remove it and download it only when needed.

seanfarley:
Oh yeah, that's fine too! I see you merged this already but I could try to write a makefile rule to update the directory as well.

PR #1115, opened 2023-09-28
Swift Language support

Obverser:
Created an extension for syntax highlighting and LSP support (on macOS so far) for Swift.
Adding Linux support is pretty simple, but I have to set up a Linux machine for testing.

A quick note, it seems that the LSP fights over the message box, which is a bit confusing for using autocomplete. Any suggestions here?
image

### DIFF
--- a/extensions/swift-mode/lem-swift-mode.asd
+++ b/extensions/swift-mode/lem-swift-mode.asd
@@ -1,0 +1,4 @@
+(defsystem "lem-swift-mode"
+  :depends-on ("lem" "yason")
+  :serial t
+  :components ((:file "swift-mode") (:file "lsp-config")))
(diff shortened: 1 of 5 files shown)


cxxxr:
Thank you very much for the great PR.

> A quick note, it seems that the LSP fights over the message box, which is a bit confusing for using autocomplete. Any suggestions here?

Thanks for letting me know.
I think this is a bug in lsp-mode.
It needs to be fixed in the future.

cxxxr:
Window conflicts have been fixed.

vindarel:
Nice! I created a documentation page: https://github.com/lem-project/lem-page/blob/master/content/en/modes/swift.md (with not much in it)

PR #1121, opened 2023-09-30
Allow to install Lem via Roswell.

fukamachi:
Lem cannot be installed via Roswell since it adopts Qlot to manage dependencies.

This PR allows Lem to be installed via Roswell again. It needs to run `qlot install` while running `ros install lem-project/lem`. This feature actually doesn't work in the current Roswell, so users must update it first.

### Requirement

* Roswell [v23.10.14.114](https://github.com/roswell/roswell/commit/0e7a2f1b6d9153f33edf1cc7b780825913a93618) or above.
* [Qlot](https://github.com/fukamachi/qlot)

### Installation

```
$ ros install lem-project/lem
```

No need to specify `follow-dependency=t` anymore.

### Workaround

```
$ git clone https://github.com/lem-project/lem
$ cd lem
$ qlot install
$ qlot exec ros -S . install lem
```

### DIFF
--- a/lem.asd
+++ b/lem.asd
@@ -1,13 +1,12 @@
 #+ros.installing
-(progn
-  (unless (uiop:directory-exists-p (asdf:system-relative-pathname :lem #P".qlot/"))
+(let ((project-dir (uiop:pathname-directory-pathname *load-truename*)))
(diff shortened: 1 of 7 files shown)


fukamachi:
It seems Roswell's `ros.installing` feature doesn't work expectedly. We need to wait for the next release of Roswell.

PR #1123, opened 2023-10-01
Reuse the main thread's ASDF source registry in the editor thread.

fukamachi:
This allows to load Lem extensions in .lem/init.lisp even when Lem source isn't in an ASDF load path (such as `~/common-lisp`).

### DIFF
--- a/src/lem.lisp
+++ b/src/lem.lisp
@@ -198,6 +198,11 @@ Options:
 (defun run-editor-thread (initialize args finalize)
   (bt:make-thread
    (lambda ()
+     (asdf:initialize-source-registry
+      `(:source-registry
+        :inherit-configuration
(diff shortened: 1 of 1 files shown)


src/lem.lisp
```diff
@@ -198,6 +198,11 @@ Options:
 (defun run-editor-thread (initialize args finalize)
   (bt:make-thread
    (lambda ()
+     (asdf:initialize-source-registry
+      `(:source-registry
+        :inherit-configuration
+        (:also-exclude ".qlot")
+        (:tree ,(asdf:system-source-directory :lem))))
```
fukamachi, line 205: Maybe good to add `~/.lem/extensions/`.

cxxxr: Indeed, it might be good idea.

PR #1137, opened 2023-10-16
directory-mode: add C-x C-j to open this file's directory

vindarel:
and put the cursor on the file name.

previous discussion: https://github.com/lem-project/lem/pull/868

I still find this feature useful, last time in July you seemed to find it useful too.

---

It doesn't work with an image buffer:

- buffer-filename is not set
- but if I set it (image-buffer.lisp, open-image-buffer), I have an encoding error coming from find-file-buffer.


cxxxr:
I think it is good.
Thanks again for making PR.

PR #1147, opened 2023-11-01
Add Dockerfile

Arian-D:
Hey, the [current image on dockerhub](https://hub.docker.com/r/40ants/lem) that is mentioned in the README is 4 years old. I made this little `Dockerfile` to avoid the hassle of setting up quicklisp, roswell, qlot, ncurses, etc. You can run the following to build and run it.
```sh
docker build -t lem .
docker run --rm -it -v .:/shared lem
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM docker.io/fukamachi/qlot
+
+COPY . .
+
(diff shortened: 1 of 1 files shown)


cxxxr:
Thank you!

PR #1148, opened 2023-11-02
Paredit insert vertical line

ilya-fiveisky:
To work more conveniently with `||` embraced symbols (e.g. `|aSdF|`) implemented pair insertion, deletion and wrapping similar to `()` and `""`.
With corresponding tests.


cxxxr:
Thank you!

PR #1151, opened 2023-11-06
Fix typo with-current-buffers in lem-base export

kchanqvq:
This must be a typo?


cxxxr:
Yes, it's typo.
Thank you.

PR #1153, opened 2023-11-06
Export with-current-buffer and with-buffer-point, close #1152

### LINKED ISSUE #1152: Export lem-base::*current-buffer*?
What is the current way to temporarily change `(current-buffer)` (like `with-current-buffer` in Emacs)?  Although setter to `current-buffer` is exported, implementing `with-current-buffer` using the setter is not thread-safe. The best way seems to be just bind `*current-buffer*`, however currently it is internal to `lem-base` which suggest it is only used for internal implementation.   Can we export this symbol instead?

kchanqvq:
It seems that `with-buffer-point` is also user facing (not used in lem-base itself) so I'm also exporting it.


cxxxr:
Thank you.

PR #1157, opened 2023-11-14
Add lisp-search-asdf-definition command

Sasanidas:
This commands gives autocompletion to all the asd systems on the image and then go to the .asd file that define that system.


cxxxr:
Nice, looks so useful

PR #1158, opened 2023-11-14
Add time information when the crash happened

Sasanidas:


cxxxr:
How about using log4cl if logging time?

Sasanidas:
Umm, which functionality do you have in mind with log4cl? :thinking:

cxxxr:
The log is output with time as follows
```
 <INFO> [21:16:15] lem-core lem.lisp (lem) - Starting Lem
```

There is also the problem that codes that record time are often duplicated and the format is not consistent.

Sasanidas:
I see, is there any easy way to pipe the output of log4cl to the same output as the command?

cxxxr:
I think it should be something like this.

```common-lisp
(log:config :sane :daily (merge-pathnames "logs/error.log" (lem:lem-home)))
(log:error value)
```

Sasanidas:
Oh so you mean changing the function to use log4cl entirely, I see, I'll do that then :+1:

PR #1159, opened 2023-11-15
Vi lisp word separators

ilya-fiveisky:
#### For `word`
In Lisp, characters `/`, `.`, `:`, and `-` syntactically can be part of a symbol or package name but semantically they are separators because typical symbol name is a phrase consisting of several words separated by `-` like `some-symbol-name` and package name looks like `some-package/some-subpackage`.
It would be natural to expect vi motion commands to take this into account.
#### For `WORD`
In Lisp code like `(qwer [a]sdf-uiop)` or `"qwer [a]sdf"`, where `[]` is a cursor position, nobody would expect commands like `DE` to create unbalanced parentheses or double quotes. 
### Implementation
Implemented via specializing `iskeyword` and `isseparator` options initialization on option names and syntax tables. For Lisp syntax table characters `/`, `.`, `:`, and `-` are removed from `iskeyword` option and characters `(`, `)` and `"` are added to `isseparator` option.


fukamachi:
LGTM, thanks!

Issue #1164, opened 2023-11-25
Open file in new window doesn't open desired file, instead opens directory-mode of the current file on the new window 


daboddo:
## Description
Trying to open a file in a new window (`C-x 4 f`) instead opens directory-mode on the directory the last buffer is in on the new window.

## How to replicate
Open file on a new window (`C-x 4 f`)

## Current result
A new window opens with a directory-mode buffer displaying the directory the buffer where the command was run is on.

## Expected result
A new window opens with the selected file/directory on the prompt.

PR #1165, opened 2023-11-27
Fix the pathname detection

Sasanidas:
Seems like when the file is on the ARG, not always is a PATHNAME, sometimes is a string that can be a valid absolute pathname.

Checking the absolute path seems to solve the issue.


Sasanidas:
This is related to https://github.com/lem-project/lem/issues/1164

cxxxr:
Thank you!

PR #1170, opened 2023-11-29
Lisp highlight

cxxxr:
This is a PR that adds variable highlighting to lisp mode.

lisp-highlight

This feature is experimental and can only be used in a few cases.


rpx99:
Nice!

PR #1192, opened 2023-12-17
Set font size

garlic0x1:
Add set-font-size in addition to increase and decrease functions, this is handy for me to be able to set it to 18 in my config


cxxxr:
I think it's good, thanks.

PR #1194, opened 2023-12-18
find-file unix path normalization

garlic0x1:
In find-file prompt, it is helpful to truncate the path before ~/ or // on Unix-style systems, otherwise paths can get really long.  This change modifies `lem/prompt-window::prompt-file-completion` to do this.  I do not use windows so this is behind a read-time feature condition and should only affect Linux/Mac

Thank you.


cxxxr:
Thank you!

PR #1198, opened 2023-12-21
legit: add simple Mercurial support

vindarel:
This has been laying on a branch for some weeks, I think it's better put in use.

Support is simple, and it will be hard to have great support for different VCSs, but starting support for 3 of them helped me modularize my code and spot some missing cases in Git. So it's beneficial to have it IMO.

### DIFF
--- a/extensions/legit/legit.lisp
+++ b/extensions/legit/legit.lisp
@@ -1,7 +1,8 @@
 (defpackage :lem/legit
   (:use :cl
    :lem)
-  (:export :legit-status)
+  (:export :legit-status
+           :*prompt-for-commit-abort-p*)
   (:documentation "Display version control data of the current project in an interactive two-panes window.
 
   This package in particular defines the right window of the legit interface and the user-level commands.
@@ -11,7 +12,7 @@
 (in-package :lem/legit)
 
 #|
-An interactive interface to Git.
+An interactive interface to Git, with preliminary support for other version-control systems (Fossil, Mercurial).
 
 Done:
 
--- a/extensions/legit/porcelain.lisp
+++ b/extensions/legit/porcelain.lisp
@@ -309,7 +309,7 @@ M	src/ext/porcelain.lisp
   This is used to stage hunks of files."
   (when *verbose*
     (log:info diff)
-    (with-open-file (f (merge-pathnames "lem-hunk-latest.patch" (lem-home))
+    (with-open-file (f (merge-pathnames "lem-hunk-latest.patch" (%maybe-lem-home))
                        :direction :output
                        :if-exists :supersede
                        :if-does-not-exist :create)
@@ -388,7 +388,7 @@ M	src/ext/porcelain.lisp
 (defun %maybe-lem-home ()
   "Return Lem's home directory by calling lem:lem-home only if the :lem package exists,
   otherwise return ~/.lem/.
-  We don't want a hard-coded dependency on Lem, to ease testing."
+  We don't want a hard-coded dependency on Lem in the porcelain package, to ease testing."
   (if (find-package :lem)
       (uiop:symbol-call :lem :lem-home)
       (merge-pathnames ".lem/" (user-homedir-pathname))))
--- a/extensions/legit/porcelain.lisp
+++ b/extensions/legit/porcelain.lisp
@@ -20,9 +20,31 @@
    :unstage
    :vcs-project-p
    )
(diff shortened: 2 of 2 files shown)


extensions/legit/legit.lisp
```diff
@@ -1,7 +1,8 @@
 (defpackage :lem/legit
   (:use :cl
    :lem)
-  (:export :legit-status)
+  (:export :legit-status
+           :*prompt-for-commit-abort-p*)
```
cxxxr, line 5: Is there a better way than exporting this variable and making it part of the API?
For example, how about using lem:config?
Also, I personally think that the p suffix in naming should only be used for functions or function parameters (e.g. read's recursive-p).

extensions/legit/porcelain.lisp
```diff
@@ -388,7 +621,7 @@ M	src/ext/porcelain.lisp
 (defun %maybe-lem-home ()
```
cxxxr, line 621: Is it possible that the lem package does not exist?

extensions/legit/porcelain.lisp
```diff
@@ -433,6 +666,13 @@ M	src/ext/porcelain.lisp
     (str:starts-with-p hash root)))
 
 (defun rebase-interactively (&key from)
+  (case *vcs*
+    (:git (git-rebase-interactively :from from))
+    (:hg (error "Interactive rebase not implemented for Mercurial"))
+    (:fossil (error "No interactive rebase for Fossil."))
+    (t (error "Interactive rebase not available for this VCS: ~a" *vcs*))))
```
cxxxr, line 673: Basically, ERROR is treated as an editor bug when the command loop catches it.
Wouldn't it be better to use editor-error?

extensions/legit/porcelain.lisp
```diff
@@ -20,18 +20,45 @@
...
 (declaim (type (cons) *git-base-arglist*))
 (defvar *git-base-arglist* (list "git")
   "The git program, to be appended command-line options.")
 
 (declaim (type (cons) *fossil-base-args*))
 (defvar *fossil-base-args* (list "fossil")
-  "fossil, to be appended command-line options.")
+  "The fossil program, to be appended command-line options.")
 
+(declaim (type (cons) *hg-base-arglist*))
```
cxxxr, line 57: `(proper-list string)` would be more descriptive.

cxxxr (commented):
I think your comments are very thoughtful and well-written.

I focused my comments on the API with external parties, so please check it out if you like.

extensions/legit/porcelain.lisp
vindarel: It is actually possible: I develop for Lem both in Lem and in Emacs. And in Emacs, I don't load Lem because… it is not clear to me how to do it, if I ever knew. Quickloading Lem isn't enough, one has to add its dependencies, etc.

So it's necessary for me that the porcelain package doesn't depend on Lem. And I think it's a good separation of concerns.

cxxxr: Okay, I see why.
However, the dependency on lem is still there, just hidden and not statically obvious.
Ideally, porcelain.lisp should be loosely coupled to lem, but with this, wouldn't it be better to be able to resolve symbols statically with lem:lem-home?
That should be solved with just `(ql:quickload :lem)`.

cxxxr:
I will merge this PR.

vindarel:
and I'll handle your reviews.

extensions/legit/porcelain.lisp
vindarel: I wanted to follow your suggestion and use `lem:editor-error`, but then this adds coupling, again, and I would prefer to avoid it. I found how to load Lem in Emacs, but I am hesitant. `porcelain` can be a package on its own… unless it makes no sense, since all commands that use it are defined on Lem's side. 

I'll think this through a bit more.

vindarel: OK I do it by simply using uiop:symbol-call again.

extensions/legit/legit.lisp
vindarel: I like the idea to put user-facing variables into lem:config, but this system currently has no support for documentation strings, and it has no namespace system (unlike package:\*variable\*). Is lem:config really appropriate? It stores values for stuff that we changed with a global command, such as the color theme, the font size, isearch-previous-string.

PR #1200, opened 2023-12-24
Ollama client

garlic0x1:
Hello, I have added a contrib package to talk to Ollama servers.

There is a listener-mode interface and a simple prompt command that streams into a temp buffer.
You can close the stream with C-c C-c in both interfaces.

image

I added some improvements to the listener-mode interface that weren't in the last PR.

Let me know if you have any feedback,
Thank you.


cxxxr:
Thank you!

PR #1203, opened 2023-12-28
sdl2 icons workaround

garlic0x1:
This fixes https://github.com/lem-project/lem/issues/1178

The issue exists on Arch and OpenBSD as far as I am currently aware, the freetype2 package that sdl2_ttf depends on is built without certain options on these systems that break icons in lem.

Using the icon registry, we go from having a pointer error, to just displaying an empty box icon, which is much preferred.

There are probably other icons elsewhere in the source that I haven't run into yet, but the solution is just to add them to src/base/icons.lisp

Before:
image

After:
image

### DIFF
--- a/src/base/icon.lisp
+++ b/src/base/icon.lisp
@@ -38,5 +38,4 @@
 (register-icon "folder" #x1f4c1)
 (register-icon "right-pointing-triangle" #x25B8)
 (register-icon "down-pointing-triangle" #x25BE)
-(register-icon "down-pointing-triangle" #x25BE)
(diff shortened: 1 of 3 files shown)


garlic0x1:
This change also allows users to change the icons in their config if issues arise, for example, the following will use '>' and 'v' in filer:

```lisp
(register-icon "right-pointing-triangle" #x003E)
(register-icon "down-pointing-triangle" #x0076)
(register-icon "lock" #x004C)
```
image

cxxxr:
Thank you.

PR #1204, opened 2023-12-29
Export point accessors point-line and point-linum.

mychris:
Small change which exports the two accessors `point-line` and `point-linum`.
There might be more functions which should be exported but I didn't go through the whole file.

### DIFF
--- a/src/base/package.lisp
+++ b/src/base/package.lisp
@@ -151,6 +151,8 @@
    :copy-point
    :delete-point
    :point-buffer
+   :point-line
+   :point-linum
    :point-charpos
    :point-kind
    :point=


Sasanidas:
LGTM!

cxxxr:
The line structure is an internal structure and should not be exported.
The point-linum is also an internal structure and should be exported using the function line-number-at-point instead.

cxxxr:
Nevertheless, I appreciate your attempts to improve it.

mychris:
Thanks for pointing me to the correct API. I couldn't find it, and since the API to get the point from the buffer is public, and all the other functions regarding point are public, I guessed those should be as well.

PR #1207, opened 2023-12-29
Make it possible to recenter the window to an arbitrary line.

mychris:
Adds two new keyword arguments to `window-recenter` to make it possible to scroll the window to an arbitrary line.
Emacs uses a single argument, interpreting negative numbers as `from-bottom`, but as this is not a command, I think the additional keyword argument might be better.

I did implement it in `window-recenter`. I could also introduce a new function for arbitrary scrolling, and re-implement `window-recenter` using this new function, if you think the function name is not appropriate for this feature.

### DIFF
--- a/src/window/virtual-line.lisp
+++ b/src/window/virtual-line.lisp
@@ -1,8 +1,8 @@
 (in-package :lem-core)
 
-(defun window-recenter (window &key (line nil) (from-bottom nil))
+(defun window-recenter (window &key line from-bottom)
   "Recenter WINDOW to the given LINE number.
-LINE must be omitted, NIL, or a positive number.
+LINE must be NIL or a positive number.
 If LINE is NIL, recenter to the middle of the WINDOW.
 Otherwise, recenter to the nth LINE (starting at 0), counted from the top.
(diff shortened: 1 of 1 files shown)


src/window/virtual-line.lisp
```diff
@@ -1,14 +1,23 @@
 (in-package :lem-core)
 
-(defun window-recenter (window)
-  (unless (= (window-cursor-y window)
-             (floor (window-height-without-modeline window) 2))
+(defun window-recenter (window &key (line nil) (from-bottom nil))
```
vindarel, line 3: Hi, my 2c: &key args are NIL by default already, and I don't think Lem's codebase writes nils explicitely like this. 

thanks for the feature.

mychris: Sure, I removed the default value of the keyword arguments.

cxxxr:
I think it's good.
As you say, I don't think it is appropriate to name the function that scrolls an arbitrary number of lines window-recenter.
It would be better to define another function for scrolling and have window-recenter call it, but I guess I should merge them once and leave it for future work.

PR #1208, opened 2023-12-29
Allow binding a key to a keymap.

mychris:
This allows binding a key to a keymap instead of a command.

```
(define-key *global-keymap* "C-x w" lem-some-package:*some-package-keymap*)
```

This way, a package which provides multiple commands, but not a mode with which these commands are associated, the person implementing the package can provide a keymap, with all the commands bound to an appropriate key.

Users can then simply bind the keymap to a key-combination of their choice and can start using the package right away, without the need to bind many keys to a common prefix.

### DIFF
--- a/src/keymap.lisp
+++ b/src/keymap.lisp
@@ -36,7 +36,7 @@
 
 (defun define-key (keymap keyspec command-name)
   (check-type keyspec (or symbol string))
-  (check-type command-name symbol)
+  (check-type command-name (or symbol keymap))
   (typecase keyspec
     (symbol
      (setf (gethash keyspec (keymap-function-table keymap))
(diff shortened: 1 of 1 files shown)


garlic0x1:
Works for me, seems like a good feature.

cxxxr:
Thank you for the PR!
I was wondering, is it possible to use a keymap instead of a symbol as a define-key argument?

vindarel:
This looks nice. Do you have a concrete example and code snippet, so than I can document it? (on https://github.com/lem-project/lem-project.github.io/)

mychris:
I am not sure if there is an example right now in the lem code base, since most of the keybindings are either in the global map or in a mode specific maps.

I added it, because I have an extension like bookmark.el for my personal use right now. This extension defines a keymap `bookmark-map` with all the keybindings for all the functions of the extension. Now, in my init.lisp, I can simply bind this keymap to a prefix key and use the extension right away, without the need to bind many different commands.

Maybe within core lem, one could look through some key bindings in the global map, and if for a certain feature/internal extension, there are many commands bound to the same prefix key, a new keymap could be introduced, which is then bound to a prefix.

One thing to keep in mind is, that this might make it harder for the user to bind more keys to this prefix, because those new bindings must be made in the keymap of the feature, not in the global map.

One examples from Emacs is for instance the `help-map`, which is then bound to `C-h`.

PR #1214, opened 2024-01-01
Tetris fix

garlic0x1:
In Tetris, when multiple rows need to be removed at once, calling the timer-ms accessor does not work since the function is neither exported in timer package, nor used in the Tetris source.

This fixes the game.


garlic0x1:
If there is a reason #'timer-ms wasn't exported, I can just revert the timer file and use :: scope from tetris.

cxxxr:
Indeed, thanks!

PR #1216, opened 2024-01-01
Adds the commend current-directory and the ex command pwd.

mychris:
Adds the command `current-directory` to display the directory of the buffer as a message, or insert it into the current buffer.
Adds the vi-mode ex command `:pwd`.


cxxxr:
Thank you!

PR #1220, opened 2024-01-01
Fixes rendering of long lines when line wrapping is on.

mychris:
This patch fixes rendering of long lines when **line-wrapping** and **line-numbers** are enabled.

Previously, the separation of objects was done on the whole width of the window, which includes the `left-side-width`. This makes rendering of the first line correct, but following lines of the same logical line are wrong.

Now, the separation of objects is done without the `left-side-objects` on the reduced width. The left side is rendered separately.

The `empty-left-side-objects` are needed, otherwise there might be artefacts of the message buffer on the left side, because the empty space is never rendered.

I don't think is a nice solution, but I am not so familiar with the rendering code. Feel free to suggest any changes, or drop this PR if you have a better idea. It works for me right now and might give hints to come up with a correct solution.

### DIFF
--- a/src/display/physical-line.lisp
+++ b/src/display/physical-line.lisp
@@ -256,13 +256,16 @@
                                                left-side-objects
                                                left-side-width)
   (let* ((objects-per-physical-line
-           (separate-objects-by-width
-            (append left-side-objects (create-drawing-objects logical-line))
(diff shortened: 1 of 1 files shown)


cxxxr:
Thank you for the PR.
I merged it because I think it is good.

PR #1223, opened 2024-01-02
Adds vi-mode ex command :read.

mychris:
Only a partial implementation.
Executing a command and inserting its output is not yet supported.

The range currently does not distinguish between `:0` and `:1` so it is not possible to insert at the beginning of the file.

https://vimdoc.sourceforge.net/htmldoc/insert.html#:r

### DIFF
--- a/extensions/vi-mode/ex-command.lisp
+++ b/extensions/vi-mode/ex-command.lisp
@@ -190,6 +190,26 @@
             (lem:show-message (format nil "~A: ~S" option-name (encode-value option-value))
                               :timeout 10))))))
 
+(define-ex-command "^r(?:e|ead)?$" (range filename)
+  ;; TODO: range currently does not distinguish between :0 and :1.
+  ;; This makes it impossible to insert at the beginning of the file
+  (cond
+    ((string= "" filename)
+     (lem:message "No file name"))
+    ((char= #\! (char filename 0))
+     (lem:message "Command execution not supported"))
+    (t
+     (let ((insert-point (case (length range)
+                           (0 (lem:current-point))
+                           (1 (first range))
+                           (2 (lem:point-max (first range) (second range))))))
+       (lem:move-point (lem:current-point) insert-point)
+       (lem:line-start (lem:current-point))
+       (unless (lem:line-offset (lem:current-point) 1)
+         (lem:line-end (lem:current-point))
+         (lem:newline))
+       (lem:insert-file-contents (lem:current-point) filename)))))
+
 (define-ex-command "^cd$" (range new-directory)
   (declare (ignore range))
   (let ((new-directory (change-directory (expand-filename-modifiers new-directory))))
--- a/extensions/vi-mode/ex-command.lisp
+++ b/extensions/vi-mode/ex-command.lisp
@@ -195,9 +195,9 @@
   ;; This makes it impossible to insert at the beginning of the file
   (cond
     ((string= "" filename)
-     (lem:message "No file name"))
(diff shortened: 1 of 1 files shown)


extensions/vi-mode/ex-command.lisp
```diff
@@ -190,6 +190,26 @@
             (lem:show-message (format nil "~A: ~S" option-name (encode-value option-value))
                               :timeout 10))))))
 
+(define-ex-command "^r(?:e|ead)?$" (range filename)
+  ;; TODO: range currently does not distinguish between :0 and :1.
+  ;; This makes it impossible to insert at the beginning of the file
+  (cond
+    ((string= "" filename)
+     (lem:message "No file name"))
```
cxxxr, line 198: How about using lem:editor-error?
Errors in command execution are handled in the command loop and messages are displayed when using lem:editor-error.

extensions/vi-mode/ex-command.lisp
```diff
@@ -190,6 +190,26 @@
...
+    (t
+     (let ((insert-point (case (length range)
+                           (0 (lem:current-point))
+                           (1 (first range))
+                           (2 (lem:point-max (first range) (second range))))))
+       (lem:move-point (lem:current-point) insert-point)
+       (lem:line-start (lem:current-point))
+       (unless (lem:line-offset (lem:current-point) 1)
+         (lem:line-end (lem:current-point))
+         (lem:newline))
```
cxxxr, line 210: newline is a command, and other operators (lem:move-point, lem:line-start, lem:line-end, etc...) and the layers are different and upscale.
`(lem:insert-character point #\newline)` is recommended.

cxxxr (commented):
Thanks for the PR!
I have made a few comments and would appreciate it if you could check them out.

extensions/vi-mode/ex-command.lisp
mychris: Thanks, fixed.

extensions/vi-mode/ex-command.lisp
mychris: I changed it.
Does lem has the notion of different line-endings per buffer yet? I didn't check it and expected it.

E.g. if I edit a file on Linux which has Windows line endings, I would like the editor to insert Windows line endings. and work with those properly. That's why I used NEWLINE, since I thought it might do something like this.

cxxxr: Doesn't the character `#\newline` abstract line breaks?
For example, CRLF in windows is represented by `(#\return #\linefeed)`, but LF in linux is `#\linefeed`.

However, I have not grasped whether they behave as the example you have given, so this may be a point for improvement in the future.

cxxxr:
I merged it, thanks!

extensions/vi-mode/ex-command.lisp
mychris: Well, #\newline abstracts line breaks, but for the current OS, meaning that Lem on Windows will insert Windows newlines and Lem on Linux will insert Linux newlines.

My concern was about files, which have Windows newlines, but Lem is executed on Linux and the file is worked on on Linux. Now, newly added newline characters will be those of Linux and the file has mixed newline characters.

Emacs discovers the line endings of the file when it loads the file and configures the file encoding in such a way, that newline characters stay the same. But maybe this should be done by INSERT-CHARACTER. Or some other facility which takes the file encoding into account.

But maybe I am also mistaken and all of this is already handled somewhere in the lisp abstraction. I didn't try it and didn't look into it yet.

PR #1227, opened 2024-01-03
legit/porcelain: signal a lem:editor-error

vindarel:
As per your review on https://github.com/lem-project/lem/pull/1198

> ERROR is treated as an editor bug when the command loop catches it.

### DIFF
--- a/extensions/legit/porcelain.lisp
+++ b/extensions/legit/porcelain.lisp
@@ -622,9 +622,10 @@ M	src/ext/porcelain.lisp
 
 ;; Save our script as a string at compile time.
 (defparameter *rebase-script-content*
-  (str:from-file
-   (asdf:system-relative-pathname (asdf:find-system "lem")
-                                  "scripts/dumbrebaseeditor.sh"))
+  (when (find-package :lem)
+    (str:from-file
+     (asdf:system-relative-pathname (uiop:symbol-call :asdf :find-system "lem")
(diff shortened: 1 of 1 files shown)


extensions/legit/porcelain.lisp
```diff
@@ -167,6 +167,14 @@ Mercurial:
 ;;; Getting changes.
 ;;;
 
+(defun %editor-error (message &rest args)
+  "Signal a lem:editor-error condition if the :lem package exists, otherwise signal a simple error.
+  This helps avoiding a tight dependency on Lem in the porcelain package."
+  (let ((msg (apply #'format nil message args)))
```
cxxxr, line 173: uiop:symbol-call only fools the compiler and actually makes it dependent on lem.
Besides, this kind of pattern is troublesome when refactoring and should be avoided if possible.

I have a suggestion, how about the following?

- define porcelain condition
```common-lisp
  (define-condition porcelain-error (error) ...)
```
- Call that porcelain-error on this line.
```common-lisp
(error 'porcelain-error ...)
```
- Each define-command handles porcelain-error, converts it to editor-error, and lifts it up.
```common-lisp
(define-command legit-... (...) (...)
  (with-porcelain-error () ...))
```
or
```common-lisp
(defclass legit-advice () ())

(defmethod execute (mode (command legit-advice) argument)
  (with-porcelain-error () (call-next-method)))

(define-command (legit-... :advice-classes legit-advice) () ())
  ...)
```

PR #1229, opened 2024-01-04
Adds a recent list to the frame multiplexer + small enhancements.

mychris:
Some enhancements for the frame-multiplexer.

- Adds a recent list, so one can jump between recently visited frames. Mostly used to the jump to the most recent.
- Adds names to the frames. If no name is given, the behaviour is unchanged. Names can be removed by giving a frame an empty name

New commands:
- `frame-multiplexer-switch`  jump to a frame with the given ID (prefix argument)
- `frame-multiplexer-recent` go to a frame in the recent list
- `frame-multiplexer-rename` give a name to a frame
- `frame-multiplexer-nomalize-ids` re-assign IDs in case many frames have been deleted and created, so it is easier to jump with `frame-multiplexer-switch`

Changed commands:
- `frame-multiplexer-next` Accepts a prefix argument with which the jump can get bigger
- `frame-multiplexer-prev` Same as next
- `frame-multiplexer-delete` Accepts a prefix argument with which a specific frame can be deleted

### DIFF
--- a/src/ext/frame-multiplexer.lisp
+++ b/src/ext/frame-multiplexer.lisp
@@ -1,10 +1,17 @@
 (defpackage :lem/frame-multiplexer
-  (:use :cl :lem :lem/button)
+  (:use :cl
+        :lem
+        :lem/button
+        :lem/common/ring)
   (:export :frame-multiplexer-next
            :frame-multiplexer-prev
+           :frame-multiplexer-switch
            :frame-multiplexer-create-with-new-buffer-list
            :frame-multiplexer-delete
-           :toggle-frame-multiplexer)
+           :frame-multiplexer-recent
+           :frame-mulitplexer-rename
+           :toggle-frame-multiplexer
+           :frame-multiplexer-normalize-ids)
   #+sbcl
   (:lock t))
 (in-package :lem/frame-multiplexer)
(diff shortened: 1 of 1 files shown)


mychris:
Maybe the frame-multiplexer is a good place to start with feature keymaps, as described in #1208?

I am not sure if such a change would be too disruptive for users, but I would propose to create a new exported variable `*keymap*`, which is a keymap, assigning all the commands within the frame-multiplexer to this keymap and then assigning `C-z` to the `*keymap*`.

src/ext/frame-multiplexer.lisp
```diff
@@ -283,38 +330,98 @@
...
+                                  :when (aref (virtual-frame-id/frame-table vf) (ring-ref *recent-list* ref-n))
+                                    :do (incf valid-count)
+                                  :when (= valid-count n)
+                                    :return (ring-ref *recent-list* ref-n))))
+      (if (null recent-frame-id)
+          (editor-error "No more recent frames")
+          (let ((entry (aref (virtual-frame-id/frame-table vf) recent-frame-id)))
+            (switch-current-frame vf (frame-table-entry-frame entry)))))))
+
+(define-key *global-keymap* "C-z r" 'frame-mulitplexer-rename)
```
vindarel, line 405: I think the current style is to put all the `define-key`s at the top of the file, though there is no official rule and I don't know the maintainer's opinion.

cxxxr: Indeed, I think that is better.

mychris: Fixed, moved all `define-key`s to the top of the file.

cxxxr:
I think it's very nice!
Can I merge this PR?

cxxxr:
> What do you think about my first comment, should I just create a PR so we can look at it there?

I agree that users should be able to set a prefix other than C-z.
Looking forward to the PR.

PR #1237, opened 2024-01-06
Adds a keymap for the frame-multiplexer.

mychris:
I am not sure about the name of the `*keymap*` variable.

Currently, there is a lot of repetition in `init.el`. If the variable is named `*keymap*`, the user writes this into it's init.el to remap it:
```
(define-key *global-keymap* "C-v" lem/frame-multiplexer:*keymap*)
```

But now, the name of the keymap does not match the of the variable. But naming the variable `*frame-multiplexer-keymap*` introduces a lot of repetition in `init.el`

```
(define-key *global-keymap* "C-v" lem/frame-multiplexer:*frame-multiplexer-keymap*)
```

What do you think? What is the name of the keymap used for, is it just for user information?

### DIFF
--- a/src/ext/frame-multiplexer.lisp
+++ b/src/ext/frame-multiplexer.lisp
@@ -3,7 +3,8 @@
         :lem
         :lem/button
         :lem/common/ring)
-  (:export :frame-multiplexer-next
+  (:export :*keymap*
+           :frame-multiplexer-next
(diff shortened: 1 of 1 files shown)


cxxxr:
I think `*keymap*` is a better reference here.
https://google.github.io/styleguide/lispguide.xml#Packages

Issue #1243, opened 2024-01-08
Issue with line wrapping with SDL2 frontend when line numbers is disabled


HappyEnt:
Hello,

its me again. Unfortunately #1241 did not fix the problem completely with the SDL2 line wrapping implementation,
as it now only works when line numbers are also enabled. When they are disabled the following error is thrown upon encountering
a too long line:

```
 SDL Error (#<SDL-SURFACE {#X00000000}>): Text has zero width
Backtrace for: #<SB-THREAD:THREAD "main thread" RUNNING {7006240513}>

...

21: ((LAMBDA (LEM-CORE::E) :IN LEM-CORE::CALL-WITH-DISPLAY-ERROR) #<SDL2-TTF::SDL-TTF-ERROR {700D848853}>)
22: (SB-KERNEL::%SIGNAL #<SDL2-TTF::SDL-TTF-ERROR {700D848853}>)
23: (ERROR SDL2-TTF::SDL-TTF-ERROR :RC #<SDL2-FFI:SDL-SURFACE {#X00000000}> :STRING "Text has zero width")
24: (SDL2-TTF:RENDER-UTF8-BLENDED #<SDL2-FFI:TTF-FONT {#X1480E2A00}> #.(SB-SYS:INT-SAP #X600000CE49F0) 208 208 208 0)
25: (LEM-SDL2/DRAWING::MAKE-TEXT-SURFACE #<LEM-SDL2/DISPLAY:DISPLAY {7003972AB3}> "" NIL :LATIN)
26: (LEM-SDL2/DRAWING::MAKE-TEXT-SURFACE-WITH-CACHE #<LEM-SDL2/DISPLAY:DISPLAY {7003972AB3}> "" NIL :LATIN)
27: ((:METHOD LEM-SDL2/DRAWING::GET-SURFACE :AROUND (T T)) #<LEM-CORE/DISPLAY:TEXT-OBJECT {700D82B703}> #<LEM-SDL2/DISPLAY:DISPLAY {7003972AB3}>) [fast-method]
28: ((:METHOD LEM-SDL2/DRAWING::OBJECT-WIDTH (LEM-CORE/DISPLAY:TEXT-OBJECT T)) #<LEM-CORE/DISPLAY:TEXT-OBJECT {700D82B703}> #<LEM-SDL2/DISPLAY:DISPLAY {7003972AB3}>) [fast-method]
[...]

PR #1244, opened 2024-01-08
Fixes line-rendering for wrapping and scrolling.

mychris:
The next attempt to fix line rendering for #1243 

The sdl2 front-end requires the whole line to be rendered. Otherwise, the caching of lines won't work. It also doesn't like empty strings.

The `left-side-width` depends on the front-end. For ncurses, it is the number of characters, for sdl2 the number of pixels.

I also tried to fix the scrolling version with line-numbers. Because the line-numbers are appended before the calculations are done, those will be scrolled out as well.

Hopefully this works now.

### DIFF
--- a/src/display/physical-line.lisp
+++ b/src/display/physical-line.lisp
@@ -255,18 +255,24 @@
                                                logical-line
                                                left-side-objects
                                                left-side-width)
-  (let* ((objects-per-physical-line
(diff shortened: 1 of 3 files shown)


mychris:
I added another change which makes use of the `window-left-width` in the calculations of virtual lines. This makes the navigation of `next-line` and other functions better, when line-wrapping and line-numbers are enabled.
It is not perfect yet, but seems better than before.

PR #1247, opened 2024-01-10
Actually enable/disable the auto-save mode.

mychris:
The auto-save-mode was never actually enabled/disabled. Since `toggle-auto-save` does not call the mode, the user never sees the mode in the modeline.

Also: adding and removing the hook should be independent from wether the timer is set or not. Otherwise, disabling the mode will not remove the hook if the timer is not set, because the set `auto-save-checkpoint-frequency` to `nil`.

### DIFF
--- a/src/ext/auto-save.lisp
+++ b/src/ext/auto-save.lisp
@@ -1,7 +1,8 @@
 (defpackage :lem/auto-save
   (:use :cl :lem)
   (:export :*make-backup-files*
-           :toggle-auto-save)
+           :toggle-auto-save
+           :auto-save-mode)
   #+sbcl
   (:lock t))
 (in-package :lem/auto-save)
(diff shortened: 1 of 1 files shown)


cxxxr (commented):
I think it is good.
By the way, if you register a function as a function object, it will refer to the old function even if we redefine the function.
So in this case I prefer to do it via symbols, what do you think?

mychris:
I didn't think about it that much. But I reverted that change, its better to stick with one approach.

I already asked myself how to mimic emacs defadvice feature, because it's very useful, especially in the `vi-mode`. There are many functions which don't work in normal mode, because the cursor can never be at the "end-of-line" position, but is always on the last character. This means that function assuming that the thing they want to operate on is before the cursor don't work. `lisp-macroexpand` is one example.

cxxxr:
@mychris 
Did you know that the execute method is called when the command is executed and can be customized by defining its defmethod?

https://github.com/lem-project/lem/blob/bdd653570dd22c4219d307ebd6ba6d2a758b04c9/contrib/selection-mode/selection-mode.lisp#L92
https://github.com/lem-project/lem/blob/bdd653570dd22c4219d307ebd6ba6d2a758b04c9/src/commands/edit.lisp#L103

PR #1251, opened 2024-01-12
Fixes describe-bindings for keys that are bound to keymaps.

mychris:
With the latest change that the frame-multiplexer prefix key `C-z` is now a `keymap` (#1237), `describe-bindings` is broken, because it can't handle this case.

The implementation is now recursive and traverses each keymap which is bound to a key. In the overview, those "sub-keymaps" are shown under the section of the actual keymap and the prefix is added to each command. For `C-z` it looks like this:

```
Global (*global-keymap*)
key             binding
---             -------
C-n             next-line
...
C-z c           frame-multiplexer-create-with-new-buffer-list
C-z d           frame-multiplexer-delete
C-z p           frame-multiplexer-prev
C-z n           frame-multiplexer-next
C-z r           frame-mulitplexer-rename
```


src/commands/help.lisp
```diff
@@ -24,19 +24,26 @@
...
-                               (lambda (kseq command)
-                                 (unless (equal "UNDEFINED-KEY" (symbol-name command))
-                                   (format s "~va~(~a~)~%"
-                                           column-width
-                                           (keyseq-to-string kseq)
-                                           (symbol-name command)))))
-              (setf keymap (keymap-parent keymap))
-              (terpri s))))
+    (labels ((create-describe-func (key-prefix)
+               #'(lambda (kseq command)
```
vindarel, line 28: style: this `#'` before a lambda is from old times and not necessary.

mychris: Thanks, fixed.

mychris:
Maybe it's also worth thinking about `traverse-keymap`. Otherwise every single user of it needs to handle the keymap case. The prefix command case is handled there.
So maybe this PR should be dropped and traverse-keymap is the right place to fix?

mychris:
Yes, the fix for `traverse-keymap` is much simpler and I think the right thing to do. I pushed another commit.

cxxxr:
@mychris Can I merge this PR?

mychris:
Yes, I think this fix is good now.

PR #1252, opened 2024-01-12
Adds a signal handler to catch SIGINT and abort.

mychris:
Sometimes, the sdl2 frontend just hangs if one interrupts the process with SIGINT. I guess this is because the native sdl2 might install a signal handler for it, but there are also signal handlers from the lisp runtime, so if this is the case, something might go wrong. That's why I added the `:no-signal-handlers` hint.
https://wiki.libsdl.org/SDL2/SDL_HINT_NO_SIGNAL_HANDLERS

I didn't see any more hangs, but since it wasn't a deterministic behaviour, I don't know if its fixed or not.

In addition, I added a handler case for SIGINT to abort, and not launch the lisp debugger. I don't know if this is desired or not. I wrapped it into a read time conditional for unix, because I have no idea what happens on other systems. Might be good if someone outside linux could test?


cxxxr:
Thank you.
I don't know if it will work, but I'll merge it once.

PR #1256, opened 2024-01-15
Add command to go to quickdocs using the reference at point

Sasanidas:
A simple command to navigate to https://quickdocs.org from lisp mode.


extensions/lisp-mode/misc-commands.lisp
```diff
@@ -66,3 +66,24 @@
                              (current-buffer)))
          (package (first (lem/detective:references-packages buffer-references))))
     (%send-test-suite package)))
+
+
+(defvar *quickdocs-url* "https://quickdocs.org/")
+
+(defvar *quickdocs-check-url* nil)
```
vindarel, line 73: So is this a user-level variable? It needs a lil' docstring IMO.

PR #1258, opened 2024-01-17
Associated python shebang with python-mode.

mychris:
Associate python shebang with python mode.
Would maybe be good if it would be possible to define a regex for that?


vindarel:
The regexp can come later^^

PR #1269, opened 2024-01-21
welcome page

garlic0x1:
Add an extension package to print a nice welcome screen on startup.

ncurses screenshot:
image

sdl2 screenshot:
Screenshot from 2024-01-21 11-06-51

adding color should be possible but I am not sure how to use ANSI escape codes in a buffer.

### DIFF
--- a/extensions/welcome/lem-welcome.asd
+++ b/extensions/welcome/lem-welcome.asd
@@ -1,0 +1,4 @@
+(defsystem "lem-welcome"
+  :depends-on (:lem :lem-vi-mode)
+  :serial t 
+  :components ((:file "welcome")))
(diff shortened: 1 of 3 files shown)


garlic0x1:
This feature can be disabled by adding `(setf lem-welcome:disable-welcome t)` to your config, and the message data is exported so it can be customized as well.  If opt-in is preferred over opt-out that can be done, some people might prefer it.

I also aligned the "Welcome to Lem!" line a bit better in this last commit.

garlic0x1:
Changed `disable-welcome` to `enable-welcome` so it's off by default, just to be more conservative.

cxxxr:
that's so cool :+1:

vindarel:
@garlic0x1 Nice. Can you document it in https://github.com/lem-project/lem-project.github.io/ ?

PR #1270, opened 2024-01-22
lem-format

garlic0x1:
# Formatting for Lem
This package exposes the `format-buffer` command and supports auto-formatting on save.

# Register formatters
Formatting for JavaScript, JSON, C, Go, and Common Lisp are included as defaults, but you can register other formatters with `register-formatter` and `register-formatters`

```lisp
;; register multiple formatters at once:
(register-formatters
  ('lem-go-mode:go-mode #'gofmt)
  ('lem-c-mode:c-mode   #'clang-format))
    
;; register a single formatter
(register-formatter 'lem-c-mode #'clang-format)
```

Handlers take a buffer to format as an argument.

# Usage
You can either enable auto-formatting with `(setf lem-format:auto-format? t)` or manually use `M-x format-buffer`.  If no formatter is registered for the major mode, an error message is displayed.

# Default formatters
Current formatters and their dependencies:
- JavaScript: prettier
- JSON:       prettier
- C:          clang-format
- Go:         gofmt
- Lisp:       none

The C formatter looks for the nearest `.clang-format` spec file walking up the file tree.  I keep Torvalds' spec file at `~/.clang-format`, but a project-local one would override that.


extensions/format/README.md
```diff
@@ -0,0 +1,30 @@
+# Formatting for Lem
+This package exposes the `format-buffer` command and supports auto-formatting on save.
+
+# Register formatters
+Formatting for JavaScript, JSON, C, Go, and Common Lisp are included as defaults, but you can register other formatters with `register-formatter` and `register-formatters`
+
+```lisp
+;; register multiple formatters at once:
+(register-formatters
+  ('lem-go-mode:go-mode #'gofmt)
```
vindarel, line 10: Is quoting `'lem-go-mode:go-mode` really needed inside the macro?

extensions/format/defaults.lisp
```diff
@@ -0,0 +1,53 @@
+(defpackage #:lem-format/defaults
+  (:use :cl :alexandria-2 :lem :lem-format))
```
vindarel, line 2: Generally it is seen as bad practive to "use" a big library like alexandria. You can create a package-local-nickname or cherry-pick imports.

extensions/format/format.lisp
```diff
@@ -0,0 +1,41 @@
+(defpackage #:lem-format
+  (:use :cl :lem)
+  (:export #:auto-format? 
+           #:register-formatter 
+           #:register-formatters 
+           #:format-buffer))
+(in-package :lem-format)
+
+;; Set this to true to format on save.
+(defvar auto-format? nil)
```
vindarel, line 10: style: `*auto-format*` ? Lem doesn't have a style guide yet, but generally we use `*earmuffs*`. It's the first time I see "?" in the code base too, also lately cxxxr told me he preferred to _not_ use a suffix (I was using "-p") for a user-level parameter.

vindarel:
Nice!! I added style comments, do what you want with them.

extensions/format/README.md
vindarel: proposal/idea: in the macro, instead of `,(first binding) ` have `',(first binding) ` (quote added) so than we can write 

```lisp
(register-formatters
  (lem-go-mode:go-mode     #'gofmt)
  …)
```

which looks better to me.

I tested lightly (it compiles).

Sasanidas:
This is a big change for the PR, but maybe you can consider this approach instead:

Adding the option to add the formatter on the `define-major-mode` definition (here for example https://github.com/lem-project/lem/blob/main/extensions/c-mode/c-mode.lisp#L25)

So each formatter it's already separated in each `major-mode` file. What do you think?

vindarel:
I tried: if I have unsaved code in the buffer, calling M-x format-buffer erases it :S

garlic0x1:
@vindarel 
Fixed those style mistakes, thanks for pointing them out, also now when `M-x format-buffer` is called directly it first runs `#'save-current-buffer`,  good catch, I was mostly testing the auto-formatting where it was already hooked after save so I didn't notice it.

@Sasanidas 
Not sure, I will have to look into how `define-major-mode` is implemented to see if I can add it without too much trouble.

extensions/format/README.md
garlic0x1: Fixed this in `register-formatter` instead of the plural form so now quotes are not required for either.

extensions/format/format.lisp
garlic0x1: Not sure if I like removing the suffix, but I will do it to keep in line with cxxxr's style.

vindarel:
@garlic0x1 cool stuff. Will you add some doc on lem-project's website?

PR #1278, opened 2024-01-25
rustfmt

garlic0x1:
Integrate rustfmt into the new formatting system.  Removed unused save hooks.


vindarel:
LGTM, but we need to talk about backward-compatible changes:

    (defvar *rust-format-on-save* t)

this line is deleted. Idea:

    (defvar *rust-format-on-save* (error "This variable is deprecated, please use … … …"))

cxxxr:
Compatibility considerations are great, but I don't think anyone uses rust-mode enough to consider that much now, so I'm merging it.

PR #1286, opened 2024-02-02
remove old define-command

garlic0x1:
This should fix Ejdă's error, but I can't recreate it, so please let him test it first.


garlic0x1:
It works for him now.

arthur-dog:
Yep, works for me! (I'm Ada btw :D)

PR #1295, opened 2024-02-11
Remove hard dependency on quicklisp

atgreen:
This patch removes a hard dependency on quicklisp, and allows for the use of ocicl instead.  There should be no change for quicklisp users, but users with the `ocicl-runtime` package loaded will pull dependencies from an ocicl OCI registry.

### DIFF
--- a/src/system-utils.lisp
+++ b/src/system-utils.lisp
@@ -6,7 +6,9 @@
  systems. Loader behavior is modified by VERBOSE and SILENT."
   (unless (listp systems)
     (setf systems (list systems)))
-  (handler-case
+  (handler-bind ((error (lambda (e)
+                          (unless error-on-failure-p
+                            (return-from maybe-load-systems nil)))))
       (flet ((try-load-system (system)
                (or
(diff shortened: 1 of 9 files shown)


src/system-utils.lisp
```diff
@@ -0,0 +1,27 @@
+(in-package :lem-core)
+
+(defun maybe-load-systems (systems &key (silent t) verbose error-on-failure-p)
+  "Load system SYSTEMS, potentially downloading them from an external
+ repository.  SYSTEMS may be a single system or a list of
+ systems. Loader behavior is modified by VERBOSE and SILENT."
+  (unless (listp systems)
+    (setf systems (list systems)))
+  (handler-case
```
cxxxr, line 9: I think that when you use handler-case, the stack is rolled back and you cannot see the error content of the contents of the load-system.

How about using hanlder-bind?
```common-lisp
(handler-bind ((error (lambda (e)
                        (unless error-on-failure-p
                          (return-from maybe-load-systems nil)))))
  ...)
```

PR #1300, opened 2024-02-16
export accessors from the point class

garlic0x1:
All the other accessors from the `point` class are exported, and I would like to export these as well.


cxxxr:
Please look at this. https://github.com/lem-project/lem/pull/1204

PR #1301, opened 2024-02-17
Enabled focusing leftside window with mouse in ncurses frontend

asutoshpalai:
With the ncurses front-end, I could not switch the focus to the left side window (Filer window) by clicking on it. The mouse support was enabled by loading the `mouse-sgr1006` contrib library.

This PR fixes it by considering all the windows when calculating the window to which it should shift focus to.


cxxxr:
Thank you.

PR #1303, opened 2024-02-17
Feature/erlang mode

ioolkos:
First version of an Erlang mode. Improvements are certainly possible and welcome.
Currently configured for the Whatsapp ELP Erlang language server, but the Erlang LS could certainly be used as well.

### DIFF
--- a/extensions/erlang-mode/README.md
+++ b/extensions/erlang-mode/README.md
@@ -1,0 +1,22 @@
+# Erlang mode for Lem
+
+## Features:
+
+### Syntax Highlighting
+
+### Shell (Run Erlang)
+
+`M-X run-erlang`
+
+### Language Server
+
+Erlang mode will connect to your local ELP [Erlang Language Platform](https://github.com/WhatsApp/erlang-language-platform) installation.
+Configure the ELP binary's location by setting *lsp-elang-elp-server-path*:
+
+```
+(defvar *lsp-erlang-elp-server-path*
+  (uiop:native-namestring "/usr/local/bin/elp")
+   "Adapt to your system's ELP path.")
+```
+
+Currently, Erlang mode does not add any LSP features on its own. The integration is experimental with the purpose of figuring out how to use the existing Lem/LSP features and how to add Erlang specific features.
--- a/extensions/erlang-mode/erlang-mode.lisp
+++ b/extensions/erlang-mode/erlang-mode.lisp
@@ -9,9 +9,6 @@
 
 (in-package :lem-erlang-mode)
 
-(defvar *current-theme* "emacs-light")
-
-
 ; Defvars are taken from Erlang OTP's "erlang.el"
 ; to follow official naming.
 ; (https://github.com/erlang/otp/tree/master/lib/tools/emacs)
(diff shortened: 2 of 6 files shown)


extensions/erlang-mode/erlang-mode.lisp
```diff
@@ -0,0 +1,469 @@
...
+  (:import-from #:ppcre
+                #:parse-string)
+  (:export :*erlang-mode-hook*
+           :erlang-mode
+           :*erlang-syntax-table*
+           :*erlang-mode-keymap*))
+
+(in-package :lem-erlang-mode)
+
+(defvar *current-theme* "emacs-light")
```
vindarel, line 12: Is that necessary? Seems a dev artifact.

extensions/erlang-mode/erlang-mode.lisp
```diff
@@ -0,0 +1,469 @@
...
+      "timestamp"
+      "universaltime"
+      "universaltime_to_localtime"
+      "universaltime_to_posixtime"
+      "unique_integer"
+      "yield"))
+    
+(defun tokens (boundary strings)
+  (let ((alternation
+          `(:alternation ,@(sort (copy-list strings) #'> :key #'length))))
```
vindarel, line 361: Would `alexandria:flatten` do the job instead of `,@`? (didn't test this case)

extensions/erlang-mode/lsp-config.lisp
```diff
@@ -0,0 +1,25 @@
+(defpackage :lem-erlang-mode/lsp-config
+  (:use :cl :lem-lsp-mode :lem-lsp-base/type))
+
+(in-package :lem-erlang-mode/lsp-config)
+
+(defvar *lsp-erlang-elp-server-path*
+  (uiop:native-namestring "~/Applications/elp")) ; adapt to your own elp path
```
vindarel, line 7: you could place the comment as the defvar documentation.

(Is it really the default location?? Maybe add a runtime check that it exists?)

vindarel:
Great! 

I left comments (I'm a helping hand).

Could you add a paragraph on the features this brings? Here, or on the package's `:documentation`, or in a new README… so that we can easily add doc to Lem's website.

I saw:

- syntax highlighting
- run Erlang
- send and eval region (async)
- LSP support (details?)
- REPL?

re style: I would define lists / property lists with the default `(list :key val)` instead of `'(:key ,val)` (and ,@).

ioolkos:
@vindarel Thank you for your review & suggestions! I'll try to adapt & extend the PR, based on them.

extensions/erlang-mode/README.md
```diff
@@ -0,0 +1,22 @@
...
+### Syntax Highlighting
+
+### Shell (Run Erlang)
+
+`M-X run-erlang`
+
+### Language Server
+
+Erlang mode will connect to your local ELP [Erlang Language Platform](https://github.com/WhatsApp/erlang-language-platform) installation.
+Configure the ELP binary's location by setting *lsp-elang-elp-server-path*:
```
vindarel, line 14: watch out, "*" needs to be escaped in code: \`\*foo\*\`

vindarel:
LGTM

cool, thanks. If the PR is ready you can add cxxxr as reviewer, I'll let him merge.

cxxxr:
I think it's good and I merged.
Thank you for the PR!

PR #1304, opened 2024-02-17
Interactive markdown

garlic0x1:
This adds interactive features to markdown code blocks, like how src blocks in org-mode work.

The API consists of 3 commands: `markdown-eval-block`, `markdown-eval-block-and-insert`, and `markdown-kill-block-result`, as well as one macro, `register-block-evaluator` for users to add support for other languages (lisp and bash work by default).

`markdown-eval-block` evaluates the source block your cursor is in, and outputs the result to a popup buffer (just like `lisp-macroexpand`)

`markdown-eval-block-and-insert` evaluates the block and inserts it's result below the block.  If there was already a result there, it first deletes the old one.

`markdown-kill-block-result` deletes the result block below the code block your cursor is in, this is run automatically before `markdown-eval-block-and-insert`.

If you try to evaluate a block and there is no registered evaluator, a message is displayed saying so.  If you use any of these commands outside a code block, nothing happens.  If you use these commands while not in markdown-mode, a warning is displayed.

Evaluation commands work asynchronously and you can do whatever you want while it is working.

Here is a demo:
[Screencast from 2024-02-17 14-15-46.webm](https://github.com/lem-project/lem/assets/43155857/63621806-b594-42b6-842e-788ca5e36203)

### DIFF
--- a/extensions/markdown-mode/interactive.lisp
+++ b/extensions/markdown-mode/interactive.lisp
@@ -133,6 +133,8 @@
   "Register evaluator for Lisp blocks."
   (lem-lisp-mode:check-connection)
   (lem-lisp-mode:lisp-eval-async
-   `(handler-case (eval (read-from-string (format nil "(progn ~a)" ,string)))
-      (error (c) (format nil "Error: ~a" c)))
+   (read-from-string (format nil "(progn ~a)" string))
    callback))
+
+;; `(handler-case (eval (read-from-string (format nil "(progn ~a)" ,string)))
+;;    (error (c) (format nil "Error: ~a" c)))
--- a/extensions/markdown-mode/interactive.lisp
+++ b/extensions/markdown-mode/interactive.lisp
@@ -45,27 +45,29 @@
       (let ((end (current-point)))
         (values lang (points-to-string start end))))))
 
-(defun delete-old-result ()
+(define-command kill-block-eval-result () ()
   "Searches for a result block below the current code block, and kills it."
-  (with-constant-position
-    (search-forward-regexp (current-point) "```")
-    (line-offset (current-point) 2)
-    (when (equal "result" (block-fence-lang (line-string (current-point))))
-      (loop :while (not (equal "```" (line-string (current-point))))
-            :do (kill-whole-line)
-            :do (line-offset (current-point) 1))
-      (kill-whole-line)
-      (kill-whole-line))))
+  (when (with-constant-position (block-at-point (current-point)))
+    (with-constant-position
+      (search-forward-regexp (current-point) "```")
+      (line-offset (current-point) 2)
+      (when (equal "result" (block-fence-lang (line-string (current-point))))
+        (loop :while (not (equal "```" (line-string (current-point))))
+              :do (kill-whole-line)
+              :do (line-offset (current-point) 1))
+        (kill-whole-line)
+        (kill-whole-line)))))
 
 (defun pop-up-eval-result (result)
   "Display results of evaluation."
   (pop-up-buffer "*literate*" (format nil "~a" result)))
 
 (defun insert-eval-result (result)
-  (delete-old-result)
-  (search-forward-regexp (current-point) "```")
-  (insert-string (current-point)
-                 (format nil "~%~%```result~%~a~%```" result))
+  (with-constant-position
+    (kill-block-eval-result)
+    (search-forward-regexp (current-point) "```")
+    (insert-string (current-point)
+                   (format nil "~%~%```result~%~a~%```" result)))
   (redraw-display))
 
 (defun eval-block-internal (handler)
--- a/extensions/markdown-mode/interactive.lisp
+++ b/extensions/markdown-mode/interactive.lisp
@@ -12,12 +12,11 @@
          (lambda (,string ,callback)
            ,@body)))
 
-(defmacro with-constant-position (&body body)
-  "This allows you to move around the buffer point without worry."
-  `(let* ((buf (current-buffer))
-          (p (copy-point (buffer-point buf))))
-     (prog1 ,@body
-       (move-point (buffer-point buf) p))))
+(defmacro with-constant-position ((point) &body body)
+  "This allows you to move around the point without worry."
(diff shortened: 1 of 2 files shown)


extensions/markdown-mode/interactive.lisp
```diff
@@ -0,0 +1,124 @@
...
+
+(defmacro register-block-evaluator (language (string callback) &body body)
+  "Convenience macro to register block evaluators, wraps setf."
+  `(setf (gethash ,language *block-evaluators*)
+         (lambda (,string ,callback)
+           ,@body)))
+
+(defmacro with-constant-position ((point) &body body)
+  "This allows you to move around the point without worry."
+  `(let ((tmp (copy-point ,point)))
```
vindarel, line 18: Did you know of `with-point(s)`?

extensions/markdown-mode/interactive.lisp
```diff
@@ -0,0 +1,124 @@
...
+        (funcall evaluator block (curry handler point))
+        (message "No evaluator registered for ~a." lang)))))
+
+(define-command eval-block () ()
+  "Evaluate current markdown code block and display results in popup."
+  (when-markdown-mode
+    (eval-block-internal (copy-point (current-point)) #'pop-up-eval-result)))
+
+(define-command eval-block-and-insert () ()
+  "Evaluate current markdown code block and display results in popup."
```
vindarel, line 95: copy-paste typo: "in popup" again

extensions/markdown-mode/interactive.lisp
```diff
@@ -0,0 +1,124 @@
...
+  (lem-lisp-mode:lisp-eval-async
+   `(eval (read-from-string ,(format nil "(progn ~a)" string)))
+   callback))
+
+;;
+;; Keybindings
+;;
+
+(define-keys lem-markdown-mode::*markdown-mode-keymap*
+  ("C-c C-e" 'eval-block)
```
vindarel, line 122: so far I have seen keybindings listed at the top of the files.

extensions/markdown-mode/lem-markdown-mode.asd
```diff
@@ -1,4 +1,5 @@
-(defsystem "lem-markdown-mode"
+(asdf:defsystem "lem-markdown-mode"
```
vindarel, line 1: I'm afraid you will be tempted to add this "asdf:" to every .asd files. I learned (the hard way) that it may not be necessary (and is discouraged by the ASDF doc). We should load .asd files with `asdf:load-asd` instead of `load`, in that case load-asd knows the defsystem. In Emacs and Lem, C-c C-k in the .asd uses asdf:load-asd too. And yeah, load-asd wants a full path and doesn't work with just (asdf:load-asd "myfile.asd") but "path/to/myfile.asd", annoying…

I feel you and I'm also very tempted, my 2c.

garlic0x1: Okay, I think Lem might have a bug here then.

Screenshot from 2024-02-18 06-25-39

extensions/markdown-mode/interactive.lisp
garlic0x1: changed all "popup" to "pop-up"

extensions/markdown-mode/interactive.lisp
garlic0x1: Doesn't seem to work the same, I will play around with it for a bit though and see if I can get it to work.

I tried to replace 
```lisp
(with-constant-position (point)
  ...)
```

with 
```lisp
(with-point ((point point))
  ...)
```

extensions/markdown-mode/interactive.lisp
```diff
@@ -0,0 +1,120 @@
...
+  (search-backward-regexp point "```")
+  (when-let ((lang (block-fence-lang (str:trim (line-string point)))))
+    (search-forward point (format nil "~%"))
+    (let ((start (copy-point point)))
+      (search-forward-regexp point "```")
+      (search-backward point (format nil "~%"))
+      (let ((end point))
+        (values lang (points-to-string start end))))))
+
+(define-command kill-block-eval-result (&optional (point (current-point))) ()
```
cxxxr, line 60: The command name defined in define-command has only a global namespace, so it is better to have a prefix.

extensions/markdown-mode/interactive.lisp
cxxxr: If you don't call delete-point for copy-point, the memory leaks.

garlic0x1:
What is your guys' opinions on adding a feature like this that allows custom post-processing:

````
```http
GET http://localhost/json
```json
````

The idea is that this could be added on to output formatted JSON, this could be used in any situation where the `~a`
format is not desired.  Examples include formatted JSON and HTML as well as using `~s` or custom functions handling database tables.

Downside is that it doesn't look pretty in a markdown renderer, although for default behavior it would change nothing.

What are your thoughts?

vindarel:
hello, where are we on this PR?

garlic0x1:
> hello, where are we on this PR? 

I don't use Lem anymore, if anyone wants to take this over feel free.  iirc it had a minor bug to fix before it was ready to merge.

Waelwindows:
> What is your guys' opinions on adding a feature like this that allows custom post-processing:
> 
> ```
> ```http
> GET http://localhost/json
> ```json
> ```
> 
> The idea is that this could be added on to output formatted JSON, this could be used in any situation where the `~a` format is not desired. Examples include formatted JSON and HTML as well as using `~s` or custom functions handling database tables.
> 
> Downside is that it doesn't look pretty in a markdown renderer, although for default behavior it would change nothing.
> 
> What are your thoughts?
> 
> Edit: prototype here https://github.com/garlic0x1/lem/tree/block-metadata

I think it's better to add them after the language line, that's what i've seen other markdown implementations do. See [this CommonMark example](https://spec.commonmark.org/0.31.2/#example-143)

vindarel:
We should keep and merge this PR.

cxxxr:
I will take over.

cxxxr:
Thank you for good feature! @garlic0x1

PR #1305, opened 2024-02-18
fix usage of erase-buffer in tetris

resttime:
erase-buffer requires a buffer in the argument
https://github.com/lem-project/lem/blob/1c365fe3d64b543d4a161cf52d9bec2748d6321f/src/buffer/internal/basic.lisp#L253

Due to timers running in threads a `defvar` is used to store the tetris buffer


cxxxr:
Thank you.

Issue #1271, opened 2024-01-22
Docker container issues


blakemcbride:
I have two problems with using a docker container as follows:

Problem n°1

```
$ docker build -t lem .
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

```

Problem n°2

When I try to save a file to /shared I get permission denied.  I tried starting it up in /tmp and got the same error.  (/tmp is 777).

blakemcbride:
The second problem is fixed by running:  `docker run --rm -it --privileged -v .:/shared lem`

elken:
Seems neither are related to Lem, but:

1. `docker buildx build -t lem .` (assuming you're on docker 19.03 or higher, otherwise follow the link for install instructions)

2.  As you said, it's a permissions issue

I can try to spend some time on the Docker setup & integrating ghcr soon :smile:

PR #1318, opened 2024-02-28
Container improvements

elken:
Nothing earth-shattering here, changes documented in the commits but I'll summarize here as well as provide the necessary post-merge setup

## Clean up image

The qlot image that this is based on installs qlot to the root user only, which breaks when trying to run the container as a user other than root. The reason for alpine is it produces a smaller image (200MB smaller) and otherwise runs just fine. Happy to revert this commit if needed, the main reason for the PR is the second point

## Add support for GHCR

Since it's free, might as well take advantage of hosting an image on github so that users don't have to build their own and deal with the issues associated (like #1271). You can see the result of the build [on my fork](https://github.com/elken/lem/pkgs/container/lem) and the only bit of post-merge setup is go to [here](https://github.com/lem-project/lem/settings/actions) and change the below setting to reflect the image
image

Also not sure exactly what logic should determine what triggers a new image deployment, currently it'll run regardless of the CI building/failing but I can have it linked to that if needed; though it seems to be [failing](https://github.com/lem-project/lem/actions/workflows/test.yml)...

And to show that it works
image

If preferred, I can also look into pushing to docker hub instead; but I don't think there's much point outside of saving having to type `ghcr.io/`.

Thanks all!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,14 @@
-FROM docker.io/fukamachi/qlot
+FROM alpine:latest
+
+WORKDIR /app
 
 COPY . .
 
-RUN apt-get update && apt-get install gcc libncurses-dev -y
+RUN apk add --no-cache curl bash build-base ncurses-dev sbcl git
 
-RUN qlot install
+RUN curl -L https://qlot.tech/installer | bash
 
-RUN qlot exec sbcl --noinform --load scripts/build-ncurses.lisp
+RUN qlot install && \
+    qlot exec sbcl --noinform --load scripts/build-ncurses.lisp
 
 ENTRYPOINT qlot exec sbcl --noinform --eval "(ql:quickload :lem-ncurses)" --eval "(lem:lem)" --quit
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,11 +1,14 @@
...
 
-RUN apt-get update && apt-get install gcc libncurses-dev -y
+RUN apk add --no-cache curl bash build-base ncurses-dev sbcl git
 
-RUN qlot install
+RUN curl -L https://qlot.tech/installer | bash
 
-RUN qlot exec sbcl --noinform --load scripts/build-ncurses.lisp
+RUN qlot install && \
+    qlot exec sbcl --noinform --load scripts/build-ncurses.lisp
```
cxxxr, line 12: sbcl installed from apt is out of date, so we need to use fukamachi/qlot or fukamachi/sbcl

cxxxr: also, thank you very much for the excellent PR.

elken: Alpine, it only looks out by 1 minor version https://pkgs.alpinelinux.org/package/edge/community/x86_64/sbcl :)

cxxxr: I didn't know that.
That's good then.

PR #1320, opened 2024-03-01
translate docstrings: buffer, basic

vindarel:
I kept a couple japanese sentences when I wasn't totally sure of the translation. Also it could help fellow japanese contributors. I am sorry for english predominance[0].

[0]: I'd prefer french like in the 18th century :p


cxxxr:
Thanks

PR #1322, opened 2024-03-01
Include missed env values

elken:
Due to a rebase, these were dropped for some reason...

Follow-up to #1318


elken:
https://github.com/lem-project/lem/pkgs/container/lem there we go :partying_face:

PR #1323, opened 2024-03-01
Added scrolling of unfocused window with mouse in ncurses frontend

asutoshpalai:
This feature allows us to scroll other windows under the mouse while the focus is on one window.

I use this feature heavily while referencing another code in a split window. This also matches the behaviour of Emacs and Vim (at least on macOS).

This is gated under the parameter `lem-mouse-sgr1006:*scroll-unfocused-window*` with the default value as `nil`. I am open to setting this as default behaviour, but I was not sure other users would prefer this.


cxxxr:
Thank you.
it's probably ok, I'll merge it.

Issue #1327, opened 2024-03-03
Building Lem with both SDL2 and ncurses support simultaneously


frejanordsiek:
At the present time, the build files either build Lem with SDL2 support (`make sdl2`) or ncurses support (`make ncurses`), with no option for building Lem with both.

I tried making my own `scripts/build-sdl2-ncurses.lisp` as

```common-lisp
(ql:quickload :lem-sdl2)
(ql:quickload :lem-ncurses)

(lem:init-at-build-time)

(sb-ext:save-lisp-and-die "lem"
                          :toplevel #'lem:main
                          :executable t)
```

and added a new make rule to the `Makefile`

```makefile
sdl2-ncurses:
	qlot install
	$(LISP) --noinform --no-sysinit --no-userinit --load .qlot/setup.lisp --load scripts/build-sdl2-ncurses.lisp
```

Doing `make sdl2-ncurses` builds, but when I run it I only get the terminal interface.

Is there some undocumented command line switch to control which interface to use?

If not, I suggest adding a command line switch and some sensible default, just like GNU Emacs' `-nw` argument and defaulting to GUI (could of course make different choices than they did).

frejanordsiek:
> What about `make all`? The target name is free.

It is free, but isn't that generally meant for making making more than one target file rather than a file with all features.

> I am not aware of a switch (but I'm rusted on the install process), and indeed as a user I'd prefer something shorter than `-i/--interface sdl2/ncurses`.

Could do it like Emacs and have it be `-nw` (short for `--no-window-system`) to force use of `ncurses` when it is built with both. The main reason I went with this scheme was that it could work the same for any future interfaces that could be added or removed.

frejanordsiek:
Merge Request #1329 has resolved this.

PR #1329, opened 2024-03-05
Add partial combined SDL2 and ncurses build

frejanordsiek:
This adds a partial combined SDL2 and ncurses build to address #1327 . The changes add a `-i,--interface INTERFACE` option to specify the desired interface: `sdl2` or `ncurses` (case-insensitive). If whatever is given is not available, the logic goes down the existing chain of finding a default implementation.

But, there are a few issues with this merge request, which are:

1. It doesn't auto-detect whether lem is run from a tty or not and (default on a tty should probably be ncurses, but elsewhere it might make sense for it to be SDL2). This is part of why this is a "partial" solution.
2. It is a bit of a hack. Again, "partial" solution
3. The choice of command line argument might not be the best
4. My Common Lisp is really rusty (haven't used it in about 6-7 years) and the indentation and choice of functions might not be the best.


src/lem.lisp
```diff
@@ -122,6 +124,12 @@ Options:
                                      (error "Please, specify a filename to log to."))
                                    (setf (command-line-arguments-log-filename parsed-args)
                                          filename)))
+                                ((member arg '("-i" "--interface") :test #'equal)
+                                 (let ((interface (pop args)))
+                                   (unless (and interface (/=  0 (array-total-size interface)))
```
vindarel, line 129: Are you looking for `plusp` and `length` for strings?

src/lem.lisp
```diff
@@ -122,6 +124,12 @@ Options:
...
+                                     (error "Please, specify an interface to use."))
```
vindarel, line 130: I don't think erroring here is cool. I'd rather have a default.

Also an error throws a backtrace right? A simple message to the user might have been nicer.

src/lem.lisp
```diff
@@ -122,6 +124,12 @@ Options:
...
+                                   (setf (command-line-arguments-interface parsed-args)
+                                         (intern (string-upcase interface) "KEYWORD"))))
```
vindarel, line 132: Also `alexandria:make-keyword` (with the string upcase)

> Interns the string designated by NAME in the KEYWORD package.

but if this works it works.

src/lem.lisp
frejanordsiek: Makes sense. I've changed it to printing a string to stderr in a new commit.

src/lem.lisp
frejanordsiek: I switched it to `alexandria:make-keyword` in a commit I just pushed to make it more clear what is happening to someone skimming the code fast.

frejanordsiek:
Bringing the discussion of the command line argument in #1327 over here.

Right now, the `-i INTERFACE` or `--interface INTERFACE` argument scheme is a bit unwieldy. The one advantage it has is that it can handle any number of interfaces, where as switches can only handle 2 very easily

Emacs uses `-nw` (short for `--no-window-system`) as its one switch.

Of course, it would be nice to make it autodetect if it is on a tty and default to using ncurses then and default to SDL2 if it isn't, but I am not sure how to do that. ncurses doesn't seem to have any function to detect this. Nor does SDL2. My best guess is to try `(interactive-stream-p *standard-input*)`.

frejanordsiek:
OK, I had a "duh" moment and figured out why my attempts at using `(interactive-stream-p *standard-input*)` were failing so badly. I just made a new commit that defaults the interface to ncurses if stdin is interactive (probably a terminal) and sdl2 if it isn't (probably launched from a Desktop app launcher or file browser).

frejanordsiek:
A potentially simpler scheme for the command line arguments could be

* `-nw` / `--no-window-system` from Emacs for any terminal interface (`ncurses` is the only one right now)
* `-gw` / `--graphical-window` for any graphical interface (`sdl2` is the only one right now)
* `-i INTERFACE` / `--interface INTERFACE` for explicitly picking an interface by name (might be important if there is every more than one terminal and/or graphics interface).

How does this sound?

vindarel:
mmh I find "-nw" not so explicit. What about

- `--nogui` defaults to ncurses
- `--gui` -> SDL2
- and -i

but wait, simply:

- `--ncurses`
- `--sdl` or `--sdl2`
- -i

frejanordsiek:
> mmh I find "-nw" not so explicit. What about
> 
>     * `--nogui` defaults to ncurses
> 
>     * `--gui` -> SDL2
> 
>     * and -i
> 
> 
> but wait, simply:
> 
>     * `--ncurses`
> 
>     * `--sdl` or `--sdl2`
> 
>     * -i

Why not both?

* `--ncurses` for using ncurses
* `--sdl` and `--sdl2` for using sdl2
* `--nogui` for using whatever terminal interface is available (in case anything beyond ncurses is added)
* `--gui` for using whatever GUI interface is available (in case anything beyond sdl2 is added)
* `-i` / `--interface` for specifying interface by name with a single option.

frejanordsiek:
I added them.

cxxxr:
Thank you for the great PR.
I think it is generally good, but I would like to avoid unnecessarily adding more command line options.
I think "-i" "--interface" is sufficient, do we need to add other options?

PR #1335, opened 2024-03-12
Wrap SDL2 call with trivial-main-thread to run in main thread

resttime:
On Mac OSx the GUI needs to run in the main thread. If running from another thread such as a REPL, the SDL2 frontend will fail so it's necessary to ensure execution is running in the actual main thread.


cxxxr:
Thank you

PR #1347, opened 2024-04-02
Fixed cursor movement by mouse when line numbers are enabled

asutoshpalai:
Without this fix, the cursor position was be off by the width of the line number view to the right.


cxxxr:
Thank you!

PR #1349, opened 2024-04-13
Handle ANSI escape codes.

Homo-carbonis:
I've extended the regex used to parse graphics control codes in the repl and shell to catch and discard other ANSI escape codes as well.


cxxxr:
Thank you!

PR #1360, opened 2024-05-02
prompt-after-hook

garlic0x1:
This will let people hook prompt completion when a prompt window is activated like this:

```lisp
(add-hook *prompt-after-activate-hook*
          (lambda ()
            (call-command 'lem/prompt-window::prompt-completion nil)))

(add-hook *prompt-deactivate-hook*
          (lambda ()
            (lem/completion-mode:completion-end)))
```

This will show completion items without needing to press `<Tab>`

### DIFF
--- a/src/ext/popup-window.lisp
+++ b/src/ext/popup-window.lisp
@@ -20,6 +20,7 @@
   ((offset-x :initarg :offset-x :accessor gravity-offset-x :initform 0)
    (offset-y :initarg :offset-y :accessor gravity-offset-y :initform 0)))
 (defclass gravity-center (gravity) ())
+(defclass gravity-top-display (gravity) ())
 (defclass gravity-top (gravity) ())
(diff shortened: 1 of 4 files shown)


garlic0x1:
Added another display style for prompts that are anchored at the top of display and fill width

image

Configured with these lines in config:

```lisp
(setf lem-core::*default-prompt-gravity* :top-display)
(setf lem/prompt-window::*fill-width* t)
```

garlic0x1:
Added gravities for bottom prompt, it looks like this:

image

Configuration shown:

```lisp
(setf lem-core::*default-prompt-gravity* :bottom-display)
(setf lem/prompt-window::*prompt-completion-window-gravity* :horizontally-above-window)
(setf lem/prompt-window::*fill-width* t)
```

cxxxr:
I think it's good!
Thank you so much!

PR #1363, opened 2024-05-12
Translated most (all?) doc-strings.

BierLiebHaber:
These are mostly machine translated, but I found them to still be helpful.\
If anyone has some better translation I'd be happy to include it.\
Especially the `editor-variable` doc-string is a little unclear to me, how do 'local variables' relate to 'global values'?

### DIFF
--- a/src/buffer/internal/buffer-insert.lisp
+++ b/src/buffer/internal/buffer-insert.lisp
@@ -1,10 +1,10 @@
 (in-package :lem/buffer/internal)
 
 (defvar *inhibit-read-only* nil
-  "Tなら`buffer`のread-onlyを無効にします。")
+  "If T, disables read-only for `buffer`.")
 
 (defvar *inhibit-modification-hooks* nil
-  "Tなら`before-change-functions`と`after-change-functions`が実行されません。")
+  "If T, prevents `before-change-functions` and `after-change-functions` from being called.")
 
 (define-editor-variable before-change-functions '())
 (define-editor-variable after-change-functions '())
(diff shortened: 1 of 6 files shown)


cxxxr:
Thank you for the PR.

I think this is a difference in expression between Japanese and English, or my Japanese ability.
Could you understand the following explanation?

local: has a closed scope within a buffer, variables in one buffer are not shared with variables in other buffers.
global: has a global scope like defvar.

BierLiebHaber:
Ok, then editor-variables can be both local and global? If so: the global is the default for the local, correct?

How does the changed-value-hook play into it? It it only called when the global is changed or whenever the value is changed?

cxxxr:
change-value-hook is called only when the global variable is changed.
Currently, it is not used in any other scope.

BierLiebHaber:
I changed the doc-string of the `editor-variable` class to be clearer.

One Implementation question:
The `kind` parameter passed to `variable-value` is the scope of the variable correct?
Why is the default value for `kind` different between `variable-value` and `setf variable-value` (it's `:default` for the getter and `:global` for setf)? Especially since there doesn't appear to be a method speciallizing on the `:default` kind.

cxxxr:
https://github.com/lem-project/lem/blob/e1aa6be815e1d8f9fd94ee05c91218dd351d0e8d/src/common/var.lisp#L61
https://github.com/lem-project/lem/blob/e1aa6be815e1d8f9fd94ee05c91218dd351d0e8d/src/common/var.lisp#L76

cxxxr:
Also, Thank you for the changes.

BierLiebHaber:
Ah sorry I meant the `variable-value-aux` and `setf variable-value-aux` methods.
here:
https://github.com/lem-project/lem/blob/e1aa6be815e1d8f9fd94ee05c91218dd351d0e8d/src/common/var.lisp#L57
https://github.com/lem-project/lem/blob/e1aa6be815e1d8f9fd94ee05c91218dd351d0e8d/src/common/var.lisp#L69
They are both specialized on `:global` but the `[setf] variable-value` functions pass `:default`

Also would it be ok if I rename `kind` to `scope`?

cxxxr:
Hmm, I don't know, but there doesn't seem to be any particular reason.


> Also would it be ok if I rename kind to scope?

Ok, I'd prefer it that way too.

BierLiebHaber:
Ah sorry i figured it out (a simple `M-.` answered my question 🤦‍♂️).

You can merge this, unless you have any additions/changes.

cxxxr:
It's very good!
Thank you very much.

PR #1366, opened 2024-05-16
Make lisp-compile-and-load-file ignore shebangs (`#!`)

BierLiebHaber:
A small tweak to `lisp-compile-and-load-file` so it checks for and ignores a shebang at the start of the file.
Added so CIEL and other scripts can be run/debugged easier from lem.
If it detects `#!` at the start of the buffer it calls `lisp-compile-region` on the region of the buffer without the first line.

Is there is a better way to check for `#!` at the start of the file?


cxxxr:
It's good.
Thank you!

vindarel:
Very nice, thanks! 

https://github.com/ciel-lang/CIEL/discussions/48

Issue #1221, opened 2024-01-02
vi-mode: paste in visual mode copies the selected region


anotherlusitano:
In normal mode, I can paste the yanked text normally, but when I go to visual mode and paste the text with `p`, it yanks the text instead of pasting it.

rdeckertula:
Looking at this, inside of `vi-paste-after` it seems like we call `vi-delete` on the text in the visual selection region, which pushes it onto the killring/replaces the clipboard buffer. Then when we call `paste-yank`, which calls `(lem:yank)` without any arguments. That pulls the deleted text from the clipboard.

I'm not sure if there are best practices established around when to use vi registers, the killring, or the clipboard -- so with all of this in mind, I can think of a couple of ways of addressing this bug.

1. We just call `(copy-to-clipboard-with-killring string)` immediately after `vi-delete` in `vi-paste-after`, like this:
```diff
diff --git a/extensions/vi-mode/commands.lisp b/extensions/vi-mode/commands.lisp
index 62ff3c88..02181899 100644
--- a/extensions/vi-mode/commands.lisp
+++ b/extensions/vi-mode/commands.lisp
@@ -508,6 +510,7 @@
          (multiple-value-bind (beg end type)
              (visual-region)
            (vi-delete beg end type))
+         (copy-to-clipboard-with-killring string)
          (when (and (not visual-line)
                     (eq type :line))
            (insert-character (current-point) #\Newline))))
...
```
2. We disable the clipboard before calling `vi-delete` inside of `vi-paste-after`, and rotate the killring after.
```diff
diff --git a/extensions/vi-mode/commands.lisp b/extensions/vi-mode/commands.lisp
index 62ff3c88..1dcf8d66 100644
--- a/extensions/vi-mode/commands.lisp
+++ b/extensions/vi-mode/commands.lisp
@@ -505,9 +506,11 @@
     (cond
       ((visual-p)
        (let ((visual-line (visual-line-p)))
-         (multiple-value-bind (beg end type)
-             (visual-region)
-           (vi-delete beg end type))
+         (lem-core::with-enable-clipboard nil
+           (multiple-value-bind (beg end type)
+               (visual-region)
+             (vi-delete beg end type))
+           (rotate-killring (current-killring)))
          (when (and (not visual-line)
                     (eq type :line))
            (insert-character (current-point) #\Newline))))
...
```
[...]

PR #1367, opened 2024-05-20
disable setting clipboard when pasting in visual mode

matsuyani:
This is intended to fix [issue 1221](https://github.com/lem-project/lem/issues/1221)

This prevents the clipboard from being set inside of the call to `delete-region`, and it rotates the killring back to the original selection after the call to `delete-region`.


cxxxr:
Thank you!

PR #1368, opened 2024-05-20
Add ignore-all-space option in legit

cxxxr:


extensions/legit/legit.lisp
```diff
@@ -45,6 +46,8 @@ Ongoing:
 (defvar *legit-verbose* nil
   "If non nil, print some logs on standard output (terminal) and create the hunk patch file on disk at (lem home)/lem-hunk-latest.patch.")
 
+(defvar *ignore-all-space* t)
+
```
vindarel, line 50: I would love a docstring ;)

The name is nearly perfectly self-explanatory, but maybe not for everybody. Also, it will be helpful for the future auto-generated user documentation.

Proposal:

> "If non t, show all spaces in a diff. Spaces are ignored by default.
>
> Currently Git-only. Concretely, this calls Git with the -w option."

Thanks for the ping, I appreciate to follow the changes in Legit.

cxxxr: Thank you for the your review.
I added that docstring.

PR #1369, opened 2024-05-21
Use theme colors for rainbow parens

jfaz1:
As it stands, `paren-coloring` colors are hardcoded, which can look jarring when using a theme. This PR makes it so that it just uses  `8`-`0d` of the current base16 [palette](https://github.com/chriskempson/base16/blob/main/styling.md). I tried it on a few themes and it seems to be working fine.

image

PS: I'm new to CL, so I might have done something dumb. Feel free to make whatever changes you see fit.


cxxxr:
Thank you!

PR #1370, opened 2024-05-22
Add file drag and drop support for SDL2

jfaz1:
Add an event handler for SDL's drop event so dragging files onto Lem opens it up.


vindarel:
Very nice! It just works.

PR #1371, opened 2024-05-22
Dynamic vertically-adjacent gravity

BierLiebHaber:
Documentation popups going over the edge of the screen have been driving me mad recently.
This adds a new gravity `vertically-adjacent-window-dynamic`, that ensures the popup is visible, and sets it as the default gravity for documentation popups.
It also makes the gravity type for documentation popups configurable with `*documentation-popup-gravity*`

I thought about replacing `vertically-adjacent-window` but figure some people prefer the old way.

### DIFF
--- a/extensions/lisp-mode/completion.lisp
+++ b/extensions/lisp-mode/completion.lisp
@@ -4,14 +4,16 @@
            :eval-completions
            :make-completion-items
            :symbol-completion
-           :region-completion))
+           :region-completion
+           :*documentation-popup-gravity*))
 (in-package :lem-lisp-mode/completion)
 
-(defvar *documentation-popup-gravity* :vertically-adjacent-window-dynamic)
-(setf (documentation '*documentation-popup-gravity* 'variable)
-      (format nil "Set the gravity (position) of the documentation popup.~%Possibilities are:~%~s"
-              (handler-case (lem/popup-window::ensure-gravity nil)
-                (type-error (arg) (cdr (type-error-expected-type arg))))))
+(defvar *documentation-popup-gravity* :vertically-adjacent-window-dynamic
+  "The `gravity` (anchor position) used for documentation popup windows.
+To prevent the window from going off screen the default choice tries to place
+the window to the left of the completion window, unless the space is insuficient.
+Common coices are :VERTICALLY-ADJACENT-WINDOW, :VERTICALLY-ADJACENT-WINDOW-DYNAMIC,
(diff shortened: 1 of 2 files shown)


extensions/lisp-mode/completion.lisp
```diff
@@ -7,6 +7,8 @@
            :region-completion))
 (in-package :lem-lisp-mode/completion)
 
+(defvar *documentation-popup-gravity* :vertically-adjacent-window-dynamic)
```
vindarel, line 10: Even if the setting might not be user facing, a docstring would help documentors and users: what does this solve and what are the possible values? 🙏

BierLiebHaber: I would actually like this setting to be user facing, so you are right some documentation is necessary.
Is there a better way to expose user facing variables?
Does lem have something like settings page that I missed?

BierLiebHaber: would you be fine with this as documentation?
```lisp
(setf (documentation '*documentation-popup-gravity* 'variable)
      (format nil "Set the gravity of the documentation popup. Possibilities are:~%~s"
              (handler-case (lem/popup-window::ensure-gravity nil)
                (condition (arg) (cdr (type-error-expected-type arg))))))
```
I don't really want people to have to change the documentation just because they added a new gravity, since it's not really related to the gravities themselves.

vindarel: oh, this, really? cxxxr will judge^^

> set the gravity

I didn't really follow what this does, so I'm a good guinea ping: what does the gravity solve for the documentation popup?

"Set the gravity of the documentation popup. By default, we ensure the popup is displayed *like this* to prevent *this*."

It might enough to give a couple default possibilities and refer to a function to know more.

Future doc generators might want an existing docstring in the `defvar`. It really depends.

---

You can pick cxxxr as reviewer when you're ready.

cxxxr: For now, it is better to prepare docstrings as static strings.

BierLiebHaber: I understand. I was mostly intrigued by the possibility of programmatic doc-strings, it does seem like an extremely powerful feature to autogenerate part of the documentation.

I hope the new doc-string is more helpful. There should probably be some documentation for grativies in general. But I'm not sure how it can be presented well to the user, since gravities are created from keyword symbols. One could do something along the lines of: `(setf (documentation :my-cool-gravity 'variable) "the best gravity")` for all the gravity-keywords, lem would then display the doc-string during completion.

Slightly off topic: Are there any plans for a better config-system? Figuring out what can be set in `.lem/config.lisp` is somewhat annoying.

cxxxr:
I think it's very good!

PR #1372, opened 2024-05-23
Bugfix out-of-range substitution of vi visual mode

gos-k:
Bugfix substitution next line of visual line selection in vil mode.


cxxxr:
Thanks!

PR #1375, opened 2024-05-24
Make define-command accept more verbose/understandable argument-descriptors

BierLiebHaber:
A comment on discord mentioned that strings as argument-descriptors are sub optimal, to which I agree.
This merge-request overhauls the `define-command` macro so that it accepts both old-style string descriptors, as well as list argument-descriptors.

The following descriptors were added:
`:universal`, `:universal-1`, `:universal-nil`, `:string`, `:number`, ~`:num`,~ `:buffer`, `:other-buffer`, `:file`, `:new-file` and `:region`
with `:universal-1` ~and `:num`~ being aliases of `:universal` ~and `:number`~ respectively.

I also added some documentation to the existing `:splice` descriptor and exposed the extremely helpful comment in the command-function as a doc-string.

### DIFF
--- a/src/defcommand.lisp
+++ b/src/defcommand.lisp
@@ -40,7 +40,7 @@ Descriptors (old char in parenthesis):
                            ((:universal :universal-1) `(list (or ,universal-argument 1)))
                            (:universal-nil `(list ,universal-argument))
                            (:string `(list (prompt-for-string ,(second arg-descriptor))))
-                           ((:num :number)
+                           (:number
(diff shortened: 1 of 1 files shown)


cxxxr:
I think it's very nice.

By the way, I would like to avoid preparing num, what do you think?
I think we'll want str, too, and there's no end to it.

BierLiebHaber:
I agree. If people want to type less they can still use the string-descriptors.

Issue #1382, opened 2024-05-26
Empty window on both sdl2 and ncurses


BierLiebHaber:
I get an empty (black) window on both ncurses and sdl2 after https://github.com/lem-project/lem/commit/acb910afc36b8d97c93c2e39aa41253921669ad3

BierLiebHaber:
The reason is https://github.com/lem-project/lem/blob/26583dad15d11efc56857283717a102e2296790d/src/defcommand.lisp#L25
I don't think we should `break` here. I think a `warn` is fine but breaking all custom commands seems extreme 😅

cxxxr:
Yes, I have changed it.

PR #1383, opened 2024-05-26
Replaced break on deprecation with more helpful warnings

BierLiebHaber:
This fixes https://github.com/lem-project/lem/issues/1382
Instead of `break`ing  on a deprecated string descriptor it prints a helpful warning to the `*message*` buffer.


BierLiebHaber:
Nice you already fixed it much nicer :laughing:

PR #1385, opened 2024-05-28
Fallback to asdf:load-system to create a micros server

garlic0x1:
the function `send-swank-create-server` assumes the user has quicklisp installed and fails if they don't.

I have replaced
 ```lisp
  "(ql:quickload :micros)"
```
with
```lisp
   "(handler-case (eval (read-from-string \"(ql:quickload :micros)\"))
      (error (c) (asdf:load-system :micros)))"
```

so that it falls back to `asdf:load-system` if quicklisp is not available.

Catching a reader error like this is a bit hacky, maybe we should just always use `asdf:load-system`? I'm not sure, but this seems to work.

Also, auto-formatting.


cxxxr:
Thank you!

PR #1387, opened 2024-05-28
Add :bn and :bp vi commands

trevorcode:
There are some vim users that expect to be able to cycle through their next buffer and previous buffers by using the commands:

`:bn` and `:bp`

This just simply adds those commands to the `ex-command.lisp`


cxxxr:
Thank you

Issue #1402, opened 2024-06-11
vi-mode: commands with % movement select the wrong text


davidsun0:
For the text `123 (abc) 456`, performing `y%` or `d%` yanks / deletes the wrong text. In this example, it selects `123 (` and `123 (abc)` when the cursor is on the opening and closing parens, respectively. The expected behavior is to yank / delete `(abc)` in both cases.

In larger buffers, the beginning of the incorrect selection isn't necessarily the beginning of the buffer. It sometimes starts hundreds of characters before the opening paren.

davidsun0:
This bug was introduced with #1389, but the root cause is in how vi-mode executes commands.

Movements with no numeric argument have an implicit value of one. So `w` is the same as `1w`. But interprets `y%` (yank to the matching paren) as `y1%` (yank to the line at 1% of the buffer).

The implicit numeric argument of 1 is assigned in `call-motion-command`.

https://github.com/lem-project/lem/blob/fe84475d493ba72a45a54674071c3a2afda5d581/extensions/vi-mode/commands/utils.lisp#L115-L125

I'm not sure I fully understand the function of `:universal-nil`, maybe this is a bug in how vi-mode handles it?

PR #1405, opened 2024-06-13
Fix commands with % movement

davidsun0:
- Fix issue where commands like y% were interpreted as y1% (#1402)
- Add regression tests


cxxxr:
Thank you.
By the way, the vi-mode test is failing, but is this related to your changes?

davidsun0:
The failing test is for vi macros, but the test doesn't use %, so I don't think it's caused by my changes. I'll try debugging it and see what I find.

cxxxr:
I'm sorry to have merged without waiting for CI, but it seems that I've started to fail with this PR.
https://github.com/lem-project/lem/pull/1395

davidsun0:
@cxxxr Thanks for waiting. I fixed my mistake and improved the `dd` test case.

cxxxr:
Thank you!

PR #1413, opened 2024-06-24
Add lem-extesion-manager

Sasanidas:
So, with this PR I want to add https://github.com/lem-project/lem-extension-manager for the main Lem repository.

This would be a draft until I add the commands https://github.com/lem-project/lem-extension-manager/blob/main/main.lisp#L225 that are not used on the extension-manager to the main Lem codebase and I would also like to add some documentation.


Sasanidas:
I have to comment the library, but I think all the basic code is in place, and the package system can start working.

Some improvements would be the ability to update git packages and quicklisp ones (both are quite easy to add). 

A lock file may also be nice, tho the idea is to keep the package manager light and easy to use

Sasanidas:
Doc written https://github.com/lem-project/lem-extension-manager

cxxxr:
Sorry, I've been meaning to look at this sometime, but it's been on the back burner for a long time.

cxxxr:
The word “package” seems to be the unit of extension, but it overlaps with the term “package” in Common Lisp, so I found it difficult to understand.
Wouldn't extension be better?

cxxxr:
The README says to do it like this, but I'm worried that the information will be duplicated with asd.

```
                                       :source (:type :git
                                                :url "https://github.com/smithzvk/Versioned-Objects.git"
                                                :branch "advance-versioning")
                                       :dependencies (("ip-management"
                                                       :source
                                                       (:type :git
                                                        :url "https://github.com/Sasanidas/ip-management.git"))
                                                      ("1am" :source (:type :quicklisp))))
```

cxxxr:
What is `simple-package`? The reason it has `simple` in the name is probably just to avoid a conflict with `cl:package`. Wouldn't it be better to just call it `extension`?

Sasanidas:
> The README says to do it like this, but I'm worried that the information will be duplicated with asd.
> 
> ```
>                                        :source (:type :git
>                                                 :url "https://github.com/smithzvk/Versioned-Objects.git"
>                                                 :branch "advance-versioning")
>                                        :dependencies (("ip-management"
>                                                        :source
>                                                        (:type :git
>                                                         :url "https://github.com/Sasanidas/ip-management.git"))
>                                                       ("1am" :source (:type :quicklisp))))
> ```

So, the idea is to let the user have the ability to quickly load packages from the configuration, quicklisp is just one way of getting the packages. I think ASDF would be able without too much of a hassle to figure out the correct package, and in the case of a conflict (for example, lem uses a x version of 1am, and u need y version for z package), this is the user duty to be able to get out of this "dependency hell".

Still, this is intended for small packages that the user can import that should be  compatible with Lem, or at least not have a conflict with it. Given the profile of the average Lem user, I don't think this would be a problem.

Sasanidas:
I think I adress some issues so, ping again @cxxxr

cxxxr:
sorry, I didn't notice this notification and found out via Discord.
Let's merge it for now.

PR #1414, opened 2024-06-29
add def* to grammar

garlic0x1:
This will make macros like `defmemo`, `defroute`, `defcommand`, etc. be highlighted the same as `define-*` macros.


cxxxr:
Thank you.

PR #1415, opened 2024-06-30
Directory mode jump

jsparkes:
This parallels the dired-jump command found in emacs dired-x.  Using C-x C-j in a file opens the directory that contains the file in directory-mode.   If used in a directory-mode buffer, the it moves up to the parent directory, like ^ in directory-mode.

I have this command hard-wired into my brain, and miss it in lem.

If this is not acceptable, I have an alternate change that exports directory-mode and directory-mode-up-directory so that I can implement it in my init file.


vindarel:
Hi, there is already a command on C-x C-j that mimics Dired (open the directory of the current file, place the cursor on the file name).  

>  Using C-x C-j in a file opens the directory that contains the file in directory-mode

so it seems this feature is already baked in, but better, because you only call `(find-file (buffer-directory)` when `find-file-directory` does a bit more (move the cursor, more checks). You don't re-use the existing function.

>  If used in a directory-mode buffer, the it moves up to the parent directory, like ^ in directory-mode.

looks like this is the only addition you want.

So, I am personally not happy with your PR because I use C-x C-j and you rebind it to another command, this introduces a breaking change.

jsparkes:
I've made the change to update find-file-directory only.   The directory up command should also select and recenter, but I couldn't make it work easily.

jsparkes:
It took an hour, but I now have the directory up command selecting the previous directory name similar to how it was done for find-file-directory.  I hope the code isn't too ugly.

vindarel:
nice, this LGTM thanks.

(maybe the concatenate and split-sequence could be terser by using `str:concat` and `str:split`, a minor point)

You can add cxxxr as a reviewer.

cxxxr:
I think it's good.
Thank you.

PR #1425, opened 2024-07-11
Add `make install` option for Linux systems

jfaz1:
This PR adds a simple `install` option to the Makefile which:

- Builds `sdl2-ncurses` lem
- Installs the binary in /usr/bin/
- Installs the icon and XDG desktop file for system discovery

The desktop file launches with SDL2 and has MIME types defined so it's easy to set file associations in GNOME/KDE/etc. Should work on most distros, I tested on Fedora and Ubuntu.


cxxxr:
It's very good.
If I had one comment, Don't you think it would be better to put it in /usr/local rather than /usr?

PR #1426, opened 2024-07-11
Add legit-log function (depends on cl-git)

anlsh:
Add a command, `legit-log`, which displays history starting from the current `HEAD`.

I definitely could have implemented this agnostic of `cl-git` but... I didn't. The reasons for this are

1. git is what I'm interested in: I don't really care about `fossil` and `hg`

2, I've had a lot of trouble with `magit` slowness on large repos (linux, specifically). Personally I'm very interested in keeping `legit` fast for large repos, which I think will naturally involve using `libgit2`, doing things lazily, etc. Tbh I'm hoping to establish precent and ~~push~~ help move `legit` along in that direction

Anyways so that's my motivation, and should explain some of the ~shortcuts~ choices in the chain

----------------

Also, refactor some code along the way to remove the `*collector*` global variable.

### DIFF
--- a/extensions/lsp-mode/lsp-mode.lisp
+++ b/extensions/lsp-mode/lsp-mode.lisp
@@ -1772,7 +1772,8 @@
 
 ;;;
 (define-command lsp-restart-server () ()
-  (dispose-workspace (buffer-workspace (current-buffer)))
+  (when-let (workspace (buffer-workspace (current-buffer) nil))
+    (dispose-workspace workspace))
   ;; TODO:
   ;; 現在のバッファを開き直すだけでは不十分
   ;; buffer-listを全て見る必要がある


anlsh:
Screenshot of what `legit-log` looks like

Screenshot from 2024-07-10 19-41-36

extensions/legit/legit.lisp
```diff
@@ -538,11 +545,74 @@ Currently Git-only. Concretely, this calls Git with the -w option.")
                            ;; Save the hash on this line for later use.
                            (when hash
                              (put-text-property start point :commit-hash hash)))))
-              (lem/peek-legit:collector-insert "<none>")))
+              (lem/peek-legit:collector-insert collector "<none>")))
 
         (add-hook (variable-value 'after-change-functions :buffer (lem/peek-legit:collector-buffer collector))
                   'change-grep-buffer)))))
 
+(defun get-clgit-walker (path)
```
vindarel, line 553: you would put that on porcelain (on another independent package at least).

extensions/legit/legit.lisp
```diff
@@ -538,11 +545,74 @@ Currently Git-only. Concretely, this calls Git with the -w option.")
...
+    cl-git:repository-head
+    cl-git:target
+    (lambda (x) 
+      (cl-git:revision-walk x :ordering :topological))))
+
+(define-command legit-log () ()
+  "Show log buffer."
+  (with-current-project ()
+    (lem/peek-legit:with-collecting-sources (collector :read-only nil)
+      (if (not (lem/porcelain:git-project-p))
```
vindarel, line 566: so until now we did that check and dispatch on porcelain.lisp. See `porcelain-error`.

extensions/legit/legit.lisp
```diff
@@ -538,11 +545,74 @@ Currently Git-only. Concretely, this calls Git with the -w option.")
...
+            ;; Commits.
+            (let* ((entry-group-size 20) ;; TODO(Magic number) Add parameter to control number?
+                   (shorthash-size 12)   ;; TODO(Magic number) Add parameter to control hash length?
+                   (walker (get-clgit-walker (buffer-directory)))
+                   (buffer (lem/peek-legit:collector-buffer collector)))
+              (labels ((add-log-entries () 
+                         (loop for i below entry-group-size
+                               for commit = (cl-git:next-revision walker)
+                               while (not (null commit))
+                               for shortlog = (first (split-sequence:split-sequence #\Newline (cl-git:message commit)))
```
vindarel, line 586: please move the cl-git logic either to porcelain or to another package. The legit command should be VCS-agnostic and the more about displaying as possible.

you also have `str:split` or `str:words`, choose your weapon :p

vindarel:
This PR is in conflict with https://github.com/lem-project/lem/pull/1423 right?

lem.asd
```diff
@@ -257,7 +257,8 @@
 
 (defsystem "lem/legit"
   :serial t
-  :depends-on ("lem" "lem-patch-mode" "lem-yaml-mode" "lem-markdown-mode")
+  :depends-on ("lem" "lem-patch-mode" "lem-yaml-mode" "lem-markdown-mode"
+               "cl-git" "split-sequence" "arrow-macros")
```
vindarel, line 261: Now legit depends on `libgit2-dev` on Debian, and on other libs for other systems. We need at least a check and a user-friendly message if it doesn't exist.

Now lem depends on libgit2, we'll need to enhance the installation scripts, the Docker image, the qlfile…

Shall the use of cl-git be optional, activated only if libgit2 is present?

extensions/legit/peek-legit.lisp
```diff
@@ -68,8 +68,6 @@ Notes:
 (define-attribute highlight
   (t :background :base0D))
 
-(defvar *collector*)
```
vindarel, line 71: I have mixed feelings because now we need the `collector` variable in front of many function calls :/ Is there an obvious advantage?

IMO a global variable is OK in CL.

vindarel:
Inviting @cxxxr @Sasanidas and @mmontone to discuss the implications of relying on the libgit2 foreign library.

@jfaz1 Did you try https://github.com/fiddlerwoaroof/cl-git ? (pure CL)

Also, as said in a comment, what if this code was optional, activated only in the presence of libgit2 ?

lem.asd
vindarel: At the very least you need to add a note in the README ;)

    apt install libgit2-dev

cxxxr:
Looking at the cl-git repository, it seems to be a public archive.
https://github.com/russell/cl-git
> This repository has been archived by the owner on Jan 22, 2023. It is now read-only.

I have concerns about relying on libraries that are clearly not maintained, and about the increasing number of new shared object dependencies.

Anyway, I think this legit-log is wonderful.
I wish it could be integrated with legit-status.

vindarel:
> Looking at the cl-git repository, it seems to be a public archive.

no, don't worry it's on https://sr.ht/~rsl/cl-git/ (and I talked with the developer about sphinx-cl-domain, he's still active)

vindarel:
@anlsh FYI it's very likely we merge the other PR before this one so you'll have a little rebasing task. You will also have the opportunity to adapt with cl-git the commits pagination we are adding to it.

vindarel:
hey @anlsh what do you think about introducing legit functions relying on libgit2, but only in the presence of the library? 

(it would be great to know how much of a speed up it gives too)

anlsh:
Thanks for the review @vindarel. I think adding a `libgit2`-based implementation when the library is present is a good idea, but I think `legit` could use some refactoring along the way. Just put up.

I'm (slowly) working on that now, and just put up the first PR (#1451)

PR #1430, opened 2024-07-12
Add `legit-jump-to-hunk`

### LINKED ISSUE #1424: Legit - open currently selected hunk in buffer
Currently if you hit enter on a file in legit's status, it'll open the file in a buffer. We should add the same behavior if you hit enter on a hunk.

I'll probably do this one

jfaz1:
Closes #1424 

Currently if you hit enter on a file in legit's status, it'll open the file in a buffer. This adds the same behavior to their respective hunks, making it easy to go to the exact line that's changed. Also defines a variable for the amount of diff context so users' git/mercurial configurations don't get in the way.

Usage:
- Write changes to a file
- Open legit status
- Select the file you want to view and press tab
- Select the hunk and press enter


vindarel:
@jfaz1 merge conflicts…

jfaz1:
@vindarel should be good now

PR #1434, opened 2024-07-13
Change compute-buffer-size to return multiple values

anlsh:
I don't plan to start submitting a million of these, but I was exploring this code and figured I'd make the change


cxxxr:
I think this is a matter of taste, but I don't like it.
Wouldn't it be more correct to define a structure?

anlsh:
Ah, you're right- I forgot that multiple values get downgraded to single ones in places where they're not explicitly expected. It's been a while :sweat_smile: 

You're right that a struct is probably the best solution but... this isn't worth our time. Closing

PR #1436, opened 2024-07-13
[Workaround] Fix broken length call

anlsh:
Broken with 42f8ad3b458a39e5875faa260d9d644538b85ccf @vindarel


vindarel:
oh my, thanks for the quick PR that I didn't see. I fixed it upstream.

PR #1439, opened 2024-07-14
Add better untracked file support to legit status

jfaz1:
This adds some minor QoL changes to untracked files in legit status:

- Show top-level list of files/directories inside of an untracked directory
- Fix an error when trying to display untracked file contents for binaries

### DIFF
--- a/extensions/legit/legit.lisp
+++ b/extensions/legit/legit.lisp
@@ -185,15 +185,31 @@ Currently Git-only. Concretely, this calls Git with the -w option.")
         ((eq type :deleted)
          (show-diff (format nil "File ~A has been deleted." file)))
         ((and (not cached) (not type))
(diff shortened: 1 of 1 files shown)


vindarel:
Nice, LGTM and it works fine.

Did you rebase on master? Looks like this branch is a bit behind. Also something is strange on my side, could you rebase? 

I am experiencing failures with "(ql:quickload "lem/legit")" on lem-init.lisp, even if it loads properly :thinking:

vindarel:
testing:

- press Enter on a subdirectory of the untracked sectiond and I get an error: "the value of PATHNAME is NIL…"

jfaz1:
@vindarel Went ahead and rebased, let me know if that fixes the issue for you?

Whoops I was fixing the directory issue, I'll push up another PR when I sort that one out

PR #1442, opened 2024-07-15
Use `call-with-current-project` when opening unstaged files/directories in legit status

jfaz1:
Opening files/directories in legit status can sometimes yield an error if you don't have a buffer in the current project open. This changes it so we run `call-with-current-project` before so we have the right context.

### DIFF
--- a/extensions/legit/peek-legit.lisp
+++ b/extensions/legit/peek-legit.lisp
@@ -349,17 +349,14 @@ Notes:
 
 
 (define-command peek-legit-select () ()
-  (alexandria:when-let ((file (get-matched-file)))
+  (alexandria:when-let ((path (get-matched-file)))
     (quit)
-    (alexandria:if-let
-        ((buffer (or (and (uiop:file-exists-p file)
-                          (find-file-buffer file))
-                     (find-file-buffer
-                      (merge-pathnames
-                       (lem-core/commands/project:find-root (buffer-filename))
-                       file)))))
-      (switch-to-buffer buffer)
-      (editor-error "File ~a doesn't exist." file))))
+    (uiop:symbol-call :lem/legit :call-with-current-project
+                      (lambda () (let ((full-path (merge-pathnames path (uiop:getcwd))))
+                                   (if (or (uiop:file-exists-p full-path)
+                                           (uiop:directory-exists-p full-path))
+                                       (find-file (namestring full-path))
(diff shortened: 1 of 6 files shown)


extensions/legit/peek-legit.lisp
```diff
@@ -349,17 +349,14 @@ Notes:
...
-    (alexandria:if-let
-        ((buffer (or (and (uiop:file-exists-p file)
-                          (find-file-buffer file))
-                     (find-file-buffer
-                      (merge-pathnames
-                       (lem-core/commands/project:find-root (buffer-filename))
-                       file)))))
-      (switch-to-buffer buffer)
-      (editor-error "File ~a doesn't exist." file))))
+    (uiop:symbol-call :lem/legit :call-with-current-project
```
vindarel, line 354: Can you not use `with-current-project` inside this command?

jfaz1: `symbol-call` doesn't work with macros, and `peek-legit` is defined before `legit` so I can't import it directly. Is there another way?

vindarel: so, mmh, we can move the macro definition in peek-legit or in a common file. I'm annoyed to see call-with-current-project being exported and this lambda.

you ok to give it a try?

jfaz1: Sure, I just used `symbol-call` cause I saw it being used elsewhere in `peek-legit`, I'll try to find somewhere to move it

vindarel: let's say: if this is non trivial, if doing this pulls too many strings, we can do it later. If you find a quick solution, great.

jfaz1: Works fine in `porcelain`, also kinda makes sense for it to be there. Do you prefer if I replace the calls with `lem/porcelain:with-current-project` or should I just use `import-from`?

cxxxr: The use of symbol-call seems to me to be a sign of poor design.
It is also vulnerable to refactoring, which increases the possibility that changes on the lem side will break the legit.

jfaz1: @cxxxr ye we're moving away from it, don't worry

vindarel: great, thank you!

PR #1443, opened 2024-07-16
add STYLEGUIDE

vindarel:
A styleguide can be beneficial to… guide new contributors.

There aren't a lot of "don't"s or "must"s.

I gathered remarks that were applied on some pull requests, and you might recognize one or two preferences of mine.

This first attempt is here for debate.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -16,6 +16,33 @@ For instance, we won't use `alexandria-2:line-up-first`.
 
 Try to not use `alexandria:curry` and prefer higher-order functions.
 
+## Dynamic bindings, functional style
+
+Avoid dynamic symbol calls (`uiop:symbol-call`) but rething your architecture instead.
+
+Use `defvar` and `defparameter` for user-facing variables, but avoid
+using them as global variables that store state and that are used from
+functions to functions. Have a more functional style, give explicit
+arguments to functions.
+
+Example:
+
+```lisp
+;; avoid this
+(defvar *var* 1)
+(defun foo ()
+   *var*)
+(let ((*var* 2))
+    (foo))
+```
+
+instead, have `foo` take one argument.
+
+
+## Don't ignore compiler warnings
+
+Please take attention to compiler warnings.
+
 
 ## Deprecation warnings
 
@@ -45,13 +72,17 @@ or at least, open an issue about it so we don't forget to do it, thank you.
 
 ## Errors in Lem
 
-Don't signal an `error` but an `editor-error`. This one is displayed nicely to the user.
+`error` is for internal errors, and `editor-error` is displayed nicely to the user.
 
 ## File layout
 
 Variables and parameters (`defvar`, `defparameter`) should be grouped
 and appear near the top of the file, before conditions, classes and
 functions.
+
+### Major and minor modes keybindings layout
+
+Lem modes should define all their keybindings at the top of the file.
 
 ## Git and pull requests
 
@@ -60,6 +91,7 @@ Please rebase and squash small commits together (you can do this with lem/legit 
 When your changes are about a mode or a feature, we like the commit message to say it upfront, for example:
 
     legit: add k to discard changes of unstaged files
+
 
 ## Loop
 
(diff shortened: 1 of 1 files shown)


cxxxr:
I guess this is a letter for my usual messy changes.

STYLEGUIDE.md
```diff
@@ -0,0 +1,83 @@
...
+
+Please rebase and squash small commits together (you can do this with lem/legit ! ;) ).
+
+When your changes are about a mode or a feature, we like the commit message to say it upfront, for example:
+
+    legit: add k to discard changes of unstaged files
+
+## Loop
+
+Loop keywords are written as keywords, with the `:`:
```
cxxxr, line 61: :+1:

STYLEGUIDE.md
```diff
@@ -0,0 +1,83 @@
...
+
+```lisp
+(loop :for key :in … )
+```
+
+## Macros, backquote, comma
+
+Don't write long macros, use the "call-with-" pattern.
+
+Don't define lists with backquote and comma, use the `list` constructor.
```
cxxxr, line 71: :+1:

cxxxr: it would be even better if there were some actual examples.

STYLEGUIDE.md
```diff
@@ -0,0 +1,83 @@
...
+Please also contribute to Lem's website if you add or change a feature.
+
+https://github.com/lem-project/lem-project.github.io/
+
+or at least, open an issue about it so we don't forget to do it, thank you.
+
+
+## Errors in Lem
+
+Don't signal an `error` but an `editor-error`. This one is displayed nicely to the user.
```
cxxxr, line 48: This depends on the situation.
In some cases, it is better to use error than editor-error in places where an internal error should be used.

STYLEGUIDE.md
```diff
@@ -0,0 +1,83 @@
...
+Lem depends on `alexandria`, so you can use `if-let`, `when-let` and similar functions.
+
+You can `:import-from` these symbols in your package definition, use the `alexandria:` package prefix, and define a package-local nickname.
+
+Try to not use new utility functions that you don't see in the codebase yet.
+
+For instance, we won't use `alexandria-2:line-up-first`.
+
+Try to not use `alexandria:curry` and prefer higher-order functions.
+
```
cxxxr, line 18: How about also specifying things like avoiding dynamic writing (e.g. uiop:symbol-call) and not ignoring compiler warnings?

STYLEGUIDE.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Errors in Lem
+
+Don't signal an `error` but an `editor-error`. This one is displayed nicely to the user.
+
+
+## Git and pull requests
+
+Please rebase and squash small commits together (you can do this with lem/legit ! ;) ).
```
cxxxr, line 53: What granularity are you assuming?
For example, is extracting functions, changing names, and changing to a call-with pattern too small to be included in a single commit?

anlsh:
Can we add a section on dynamic binding?

What I'm really asking: can we ban dynamic binding unless it's really necessary?

`defvar/defparameter` make a lot of sense for, say, defining variables which are to be customized by the user. But actually using dynamic extent to substitute for passing function parameters, a la 

```
(defvar *var* 1)
(defun foo ()
   *var*)
(let ((*var* 2))
    (foo))
```

is, IMO, a trap. Sure, you save a function parameter: on the other hand, you torpedo your ability to run code in separate threads.

Obviously, it's not like avoiding dynamic binding will magically make all of our code thread safe, and Lem probably doesn't really use concurrency at the moment. But I think we should plan for the future and learn from Emacs's experience in this regard

cxxxr:
I too think you are right to avoid unnecessary dynamic bindings as a good code writing practice.

However, dynamic bindings are thread-safe.
https://www.sbcl.org/manual/#Special-Variables

anlsh:
Oh wow, I had no idea that bindings were thread-localm. Thanks for clearing that up

cxxxr:
@vindarel  Is it okay to merge this?

vindarel:
Hi yes it is IMO, I edited it after your remarks.

PR #1444, opened 2024-07-16
legit: enable by default. No more "lem/legit" system.

vindarel:
Previously, we loaded legit with (ql:quickload "lem/legit") but we had to change the system name.

This is a small breaking change for everyone who has `(ql:quickload "lem/legit")` in their init file.


extensions/legit/porcelain.lisp
```diff
@@ -764,7 +764,7 @@ M	src/ext/porcelain.lisp
 (defparameter *rebase-script-content*
   #+(or lem-ncurses lem-sdl2)
   (str:from-file
-   (asdf:system-relative-pathname (asdf:find-system "lem")
+   (asdf:system-relative-pathname (asdf:find-system "lem/legit")
```
cxxxr, line 767: Is it not lem-legit?

vindarel: It is! thanks.

(rebasing was working though O_o )

PR #1445, opened 2024-07-17
Vindarel/legit common less packages better separation

vindarel:
`with-porcelain-error` and `with-current-project` belong to lem/legit to avoid tight coupling between the porcelain package and Lem, which is very important to me for both ease of development and better engineering (porcelain being independent could be used elsewhere, who knows). Actually the previous commit re-introduced the tight coupling, so here we are back to normal.

We add them to a file common to peek-legit and legit.

We also delete the peek-legit package to simplify everything, as peek-legit and legit were very tightly coupled anyways. This simplifies calling symbols from one package or another and it removes the need of dealing with so many exports.

### DIFF
--- a/extensions/legit/legit-commit.lisp
+++ b/extensions/legit/legit-commit.lisp
@@ -92,7 +92,7 @@ two
       ((str:blankp cleaned-message)
        (message "No commit message, do nothing."))
       (t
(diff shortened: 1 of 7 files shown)


cxxxr (commented):
I think it's good.

cxxxr (approved):

PR #1446, opened 2024-07-18
Asynchronously run external commands with find-file

superduperchudmeister:
First off, thank you creating Lem. I earnestly believe it could be the future of Emacs like text editors. I use it everyday.

The goal of this commit is to run external programs asynchronously by simply switching from uiop:run-program to uiop:launch-program. This in consequence changes the return value to a uiop:process-info object.

As far as I know, the find-file command is the only function reliant on the uiop:run-program return value. To remedy this, I added bufferp check in it.


cxxxr:
Thank you, I think it's good.

PR #1452, opened 2024-07-22
Add hard drop to tetris

johnlorentzson:
Come on, it's the 21st century. Every tetris clone needs a hard drop button.


cxxxr:
Very nice

PR #1454, opened 2024-07-23
Refactor buffer structure

cxxxr:


anlsh (commented):
What do you think of pulling in `defclass-std` or something similar (I think `defclass-star` exists too) to cut down on this cruft? Any reason not to? (Does go-to-definition get worse maybe?)

cxxxr:
It is futile to worry about extra dependence when there is little benefit.

PR #1455, opened 2024-07-24
Add SLIME's C-c ~ command (sync package) to lisp-mode

johnlorentzson:
In Emacs, pressing C-c ~ in a Lisp buffer with SLIME will cause the REPL's package to be switched to the buffer's active package. This change implements that functionality.

Note that it doesn't have the functionality of switching Lisp's current directory. This is partially because I found that behaviour janky in Lem, and partially because I don't like it doing that in Emacs either.

### DIFF
--- a/extensions/lisp-mode/lisp-mode.lisp
+++ b/extensions/lisp-mode/lisp-mode.lisp
@@ -473,14 +473,19 @@
 (define-command lisp-listen-in-current-package () ()
   (check-connection)
   (alexandria:when-let ((repl-buffer (repl-buffer))
-                        (package (buffer-package (current-buffer))))
+                        (package (buffer-package (current-buffer)))
+                        (original-buffer (current-buffer)))
     (save-excursion
-      (setf (current-buffer) repl-buffer)
-      (destructuring-bind (name prompt-string)
-          (lisp-eval `(micros:set-package ,package))
-        (new-package name prompt-string)))
-    (start-lisp-repl)
-    (buffer-end (buffer-point repl-buffer))))
+      (cond ((find-package package)
+             (setf (current-buffer) repl-buffer)
+             (destructuring-bind (name prompt-string)
+                 (lisp-eval `(micros:set-package ,package))
+               (new-package name prompt-string))
+             (start-lisp-repl)
+             (buffer-end (buffer-point repl-buffer)))
+            (t
+             (message "Package ~A not found" package)
+             (setf (current-buffer) original-buffer))))))
 
 (define-command lisp-current-directory () ()
   (message "Current directory: ~a"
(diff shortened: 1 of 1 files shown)


cxxxr:
Doesn't `M-x lisp-listen-in-current-package` behave the same way?

johnlorentzson:
I can't believe I missed that one, kind of embarrassing. They behave slightly differently though, `lisp-listen-in-current-package` switches the focus to the REPL whereas `lisp-sync-package` keeps the active buffer in focus (and has a few technical flaws I'm only now noticing looking at the other one).

I'm honestly not sure which behaviour is more useful. Out of habit I would lean towards SLIME's behaviour, but when just thinking about it, I realize that switching the REPL's package only makes sense if the user is about to use the REPL, so Lem's behaviour could make more sense.

Either way I think the general idea deserves a default keybind (and matching SLIME's seems appropriate).

johnlorentzson:
Two new commits, they map the keybind to `lisp-listen-in-current-package` and change the way it responds to missing packages.

Normally `lisp-listen-in-current-package` would simply signal an error if the package is missing, but on the machines I've tested on that leads to Lem freezing (and even if it didn't, it would be a little annoying to get a whole error for a quick key press at the wrong time), so I gave it a message box instead. It also doesn't enter a REPL when the package is missing, so one doesn't need to deal with even more distractions after just forgetting to load their code first.

extensions/lisp-mode/lisp-mode.lisp
```diff
@@ -472,14 +473,19 @@
...
+                        (package (buffer-package (current-buffer)))
+                        (original-buffer (current-buffer)))
     (save-excursion
-      (setf (current-buffer) repl-buffer)
-      (destructuring-bind (name prompt-string)
-          (lisp-eval `(micros:set-package ,package))
-        (new-package name prompt-string)))
-    (start-lisp-repl)
-    (buffer-end (buffer-point repl-buffer))))
+      (cond ((find-package package)
```
cxxxr, line 479: I think it is necessary to check whether the package exists on the server side.
This would search for the package within the client (within the runtime of lem), so it would only behave correctly when lem and the server are in the same runtime.

cxxxr:
I knew I would have to fix the behavior when there is no package at some point.
Also, the addition of key bindings seems convenient.

extensions/lisp-mode/lisp-mode.lisp
cxxxr: something like this

```
(lisp-eval `(not (null (cl:find-package ,package))))
```

johnlorentzson: New commit made to address this. I keep forgetting there's a server side, my bad.

johnlorentzson:
(I didn't see your comment suggesting `(not (null ...))` before making the change, hence why I bothered to explain the reasoning in the commit message.)

cxxxr:
Thank you

PR #1456, opened 2024-07-25
Use theme colors for multiplexer bar

jfaz1:
Currently the multiplexer bar has hardcoded colors. This PR:

- Creates a `*theme-load-hook*` 
- Uses hook to set multiplexer bar colors to match theme
- Fixes `frame-mulitplexer-rename` typo

Before (rose-pine-moon):
image

After (rose-pine-moon):
image

After (grayscale):
image


cxxxr:
Great! Thank you!

PR #1459, opened 2024-07-27
Make 'terminal' command launch shell in buffer's current directory

anlsh:
I'm not totally sure if the `-c "cd directory"; $SHELL` approach is really the *right* thing to do, but it at least works on both bash and fish.

I haven't regenerated any .so files: I figure @cxxxr might want to do that themselves. Although we really shouldn't be keeping binary blobs in the repo in the first place.


anlsh:
@cxxxr Did you regenerate the `.so` files btw?

cxxxr:
Yes

PR #1460, opened 2024-07-27
Add Makefile hierarchy for extensions

anlsh:
This is half-baked, admittedly. extensions/Makefile should be looping over its subdirectories somehow to avoid the big hardcoded string, and "make ext" only works on linux.

Also, the hardcoded paths in the Makefile are not great, but I'd like to change this before merge. Ideally instead of having a lib/linux and lib/macosx both with shared objects (and in-tree), we'd just have a lib/ (or build/, which I think is a bit better) which "make ext" can just fill with the right .so

We'd have to update the github releases to distribute (and generate) independent osx and linux packages, but that seems like a fair price for eliminating the blobs from the tree

Also: I'm not sure where the relevant docs (so that I can specify that that libvterm 3.3+ is required) are: seems the sources for https://lem-project.github.io/installation/ncurses/linux/ aren't actually in this repo?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,53 +35,14 @@ update:
 	git pull
 	qlot install
 
+ext:
+# linux is hardcoded because I tried copy-pasting other detection
+# stuff from StackOverflow and somehow it didn't work.
+	LEM_OS=linux make -C extensions/
+
+clean:
+	make -C extensions/ clean
+
 lint:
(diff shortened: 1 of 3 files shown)


cxxxr:
The document repository is here
https://github.com/lem-project/lem-project.github.io

by the way, I don't think there has been agreement on not including shared objects in the repository.

PR #1462, opened 2024-07-29
Fix malformed eval background color

daedsidog:
Fixes `compute-evaluated-background-color` producing a malformed color string. 

Also makes the highlight color darken instead of lighten when evaluating on light light themes.

It appears that the `compute-evaluated-background-color` function worked only through luck because it seems okay on most themes. On my theme, however, which had a pitch-black background color, the evaluation highlight looked really bad:

image

This PR corrects the issue, but it may be necessary to change the values used in the function in the future because now it probably won't look as it looked before on some themes.


cxxxr:
Thank you

PR #1463, opened 2024-07-29
Add recent file history

jfaz1:
I'm working on a dashboard similar to [emacs-dashboard](https://github.com/emacs-dashboard/emacs-dashboard) and I needed some form of recent file history. This PR:

- Adds an optional `limit` to `history`
- Adds a `find-file-history` command that displays the last 10 accessed files
- Binds it to `C-x C-h`


src/common/history.lisp
```diff
@@ -31,15 +32,16 @@
   (and (not (equal input last-input))
        (not (equal input ""))))
 
-(defun make-history (&key pathname)
+(defun make-history (&key pathname (limit nil))
```
vindarel, line 35: style: `(limit nil)` is redundant (it's nil by default) and AFAIK not the style followed so far (eg "pathname" is nil).

src/commands/file.lisp
```diff
@@ -391,3 +394,38 @@ With prefix argument INSERT, insert the directory of the active buffer at point.
...
   (format-buffer))
+
+(defvar *files-history*)
+
+(defun file-history ()
+  "Return or create the files' history struct.
+  The history file is saved on (lem-home)/history/files"
+  (unless (boundp '*files-history*)
+    (let* ((pathname (merge-pathnames "history/files" (lem-home)))
+           (history (lem/common/history:make-history :pathname pathname :limit 10)))
```
vindarel, line 405: The hardcoded 10 could be (should ?) a parameter!

src/common/history.lisp
jfaz1: Hahah good catch, left that by accident

vindarel (approved):

vindarel:
nice addition : )

maybe a history of 10 by default is small? But I didn't try, usage will tell.

jfaz1:
> maybe a history of 10 by default is small? But I didn't try, usage will tell.
Maybe. I started at 20 and tried some values, at the end I think it should probably just fit without scrolling by default, I don't think most using the feature are gonna try finding a file they accessed >20 opens ago probably.

cxxxr:
It's great!

cxxxr:
I think it would be good to integrate this with find-file (like readline's isearch or modern editors).

daedsidog:
I was really missing this from Emacs. Thanks.

PR #1465, opened 2024-07-31
Bind M-m to new move-back-to-indentation command.

jsparkes:
A command binding similar to emacs that I miss in lem.


cxxxr:
Thank you for the PR.
By the way, I deleted this command in the past because it can be replaced with move-to-beginning-of-line.
I'd like to keep M-m for setting a better command.

PR #1469, opened 2024-08-02
Add document attributes

### LINKED ISSUE #1468: Standard document/non-programming attributes
I'll do this one if you guys think it's a good idea.

We should consider coming up with some standard attributes for document/non-programming modes, for both consistency and to make life easier for extension writers and theme creators.

If you look at `markdown-mode`, several programming attributes are used so that it works with themes out of the box. For example, `syntax-constant-attribute` is used for titles. 

I propose lem include a set of generic attributes specifically for non-programming or presentation formats (e.g. markdown, org-mode, legit commit messages, etc.) that cover most bases. For example, something like 6 header/title attributes, a hyperlink attribute, bold/italic attributes, etc.

This way, we can make it easier for themes to have customized styling for text that doesn't bleed into programming modes, and extension writers have defaults they can rely on to work with most themes. 

@cxxxr @vindarel any thoughts?

jfaz1:
Closes #1468

This PR adds document attributes we can use with markup/document views (e.g. org mode, markdown, legit, etc.) and sets some working colors for the default lem theme. I'll push up a separate PR moving `markdown-mode` over to these alongside changes to the default theme ones if they don't work too well.

### DIFF
--- a/src/attribute.lisp
+++ b/src/attribute.lisp
@@ -233,7 +233,72 @@
 (define-attribute syntax-builtin-attribute
   (t :foreground "#FF87FF"))
 
-
+(define-attribute document-header1-attribute
+  (:light :foreground "#000000" :bold t)
+  (:dark :foreground "#FFFFFF" :bold t))
+
+(define-attribute document-header2-attribute
+  (:light :foreground "#005A9C" :bold t)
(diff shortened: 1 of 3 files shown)


cxxxr:
Thank you.

PR #1471, opened 2024-08-05
Completion ranking

jfaz1:
This PR adds basic ranking capabilities to `completion`. Currently only active in a couple of places, mainly `C-x p f` / `prompt-for-files-recursively`.

I wasn't quite satisfied with how project file completion worked, since the fuzzy finding pretty much only looked at matching characters, meaning you'd end up having to type most of the name a lot of the time (and sometimes the full name wasn't enough):

image

This adds optional ranking rules, enabled for `string-completions`:

image

For files the ranking rules go: 

exact match > prefix match (file) > substring match (file) > prefix match (dir/path) > substring match (dir/path) > fuzzy match

This sorts the filtered test results once after calculating the rank so it might introduce a bit of overhead. That being said I haven't felt any in difference in the projects I've tried (e.g. Linux kernel). If you guys feel any let me know, I tried a few different ways of doing this and this one ended up being the most straightforward.

### DIFF
--- a/src/completion.lisp
+++ b/src/completion.lisp
@@ -28,6 +28,8 @@
               :while pos))))
 
 (defun completion (name elements &key (test #'search) separator key rank)
+  "Perform completion on ELEMENTS matching NAME. Returns matching elements, 
+   optionally sorted by RANK function."
   (labels ((apply-key (elt) (if key (funcall key elt) elt))
            (test-with-separator (elt)
              (let* ((elt (apply-key elt))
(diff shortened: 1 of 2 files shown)


src/completion.lisp
```diff
@@ -39,10 +41,47 @@
...
+      ; Substring match in file name
+      ((search name file-name) 2)
+      ; Prefix match in full path
+      ((str:starts-with-p name elt) 3)
+      ; Substring match in full path
+      ((search name elt) 4)
+      ; Fuzzy match, rank by length
+      (t (length elt)))))
+
+(defun completion-strings (str strings &key key files)
```
cxxxr, line 80: I was a little confused because the argument files looks like it passes a list of files.
For future extensibility, could we use a different interface instead of boolean?
For example, for files, it could be completion-files instead of completion-strings.

cxxxr:
Thank you, this is a feature I wanted.

src/completion.lisp
jfaz1: Sure, I wanted to do that at first but there was another `completion-file` down there and I thought it'd be confusing. You think if I move it to a new `completion-files` function that's fine?

cxxxr: I see, it might be confusing to have two `completion-files` and `completion-file`.
I thought it would be good to have a `rank` parameter for completion-strings.
This could be refactored in the future.

PR #1479, opened 2024-08-07
More extensive Markdown syntax parsing

jfaz1:
This PR uses the new `document` attributes for `markdown-mode`, as well as adds more extensive syntax parsing. Adds distinction between header levels, bold/italic/underline/link, task list item, tables, etc. 

Before:

image

After:

image


cxxxr:
Thank you!

PR #1484, opened 2024-08-12
Add project grep

jfaz1:
Adds `project-grep` command and binds it to `C-x p g`. Feels nice not having the grep directory show up every time (also having a keybind).


cxxxr:
Thank you.
Looks good.

PR #1487, opened 2024-08-13
Add `move-to-top` to history, use with file history

jfaz1:
Adds `move-to-top` to `add-history`, which moves the entry to the top of the stack if it already exists. Then uses it with `file-history` to re-order the most recent entry if it's already in the history.

### DIFF
--- a/src/common/history.lisp
+++ b/src/common/history.lisp
@@ -57,30 +57,45 @@
     (aref (history-data history)
           (1- (length (history-data history))))))
 
-(defun add-history (history input &key (allow-duplicates t) (test #'equal))
-  "Add this input to the history.
+(defun add-history (history input &key (allow-duplicates t) (test #'equal) move-to-top)
+  "Add this INPUT to the HISTORY.
 
-  Don't add the same input as the previous one.
-  If allow-duplicates is non t, don't add duplicates at all.
-  If limit is set, overwrite oldest entry when reached."
-  (when (or allow-duplicates
-            (not (find input (history-data history) :test test)))
-    (when (require-additions-to-history-p input (last-history history))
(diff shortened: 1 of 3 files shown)


src/common/history.lisp
```diff
@@ -57,30 +57,42 @@
...
+                                (position input (history-data history) :test test))))
+    (cond
+      ((and existing-position move-to-top)
+       (let ((item (aref (history-data history) existing-position)))
+         ;; Remove the item from its current position
+         (replace (history-data history)
+                  (history-data history)
+                  :start1 existing-position
+                  :start2 (1+ existing-position))
+         (decf (fill-pointer (history-data history)))
```
jfaz1, line 77: @vindarel Is decrementing the fill-pointer fine here or should I do something like:
```
(setf (history-data history)
      (delete item (history-data history) :test test :count 1))
```

src/common/history.lisp
```diff
@@ -57,30 +57,42 @@
...
-                    :start1 0 
-                    :start2 1
-                    :end2 limit)
-           (setf (aref (history-data history) (1- limit)) input))
-          (t
-           ;; Add new element normally
-           (vector-push-extend input (history-data history)))))
-      (setf (history-index history)
-            (length (history-data history)))))
+(defun add-history (history input &key (allow-duplicates t) (test #'equal) move-to-top)
```
cxxxr, line 60: https://github.com/lem-project/lem/blob/main/tests/common/history.lisp
If possible, I would be happy if a test were added here.

jfaz1: Let me know if that works :smile:

cxxxr:
I think it's good.
Thank you.

PR #1489, opened 2024-08-14
Add variables for git command and args.

jsparkes:
I have added *grep-command* and *grep-args* for customizing the grep command.  This is useful for windows, which doesn't have grep by default.  I exported the variables, not sure if that's desired.

An alternative would be to rename *last-query* to *grep-command* and put the entire command in there.  Let me know if you would prefer that.

I have added the following to my init.lisp:
#+win32
(setq *grep-command* "c:/cygwin64/bin/grep.exe")

### DIFF
--- a/src/ext/grep.lisp
+++ b/src/ext/grep.lisp
@@ -1,7 +1,10 @@
 (defpackage :lem/grep
   (:use :cl
         :lem)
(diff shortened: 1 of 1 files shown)


cxxxr:
Thank you.

PR #1491, opened 2024-08-17
Move back to indentation, bound to M-m

jsparkes:
Another emacs command I miss in lem.


cxxxr:
https://github.com/lem-project/lem/pull/1465#issuecomment-2259947869

PR #1492, opened 2024-08-18
Re-enable super key support in SDL2

theangelperalta:
The super key seems to be working properly on my M1 MacBook Pro 16'  running on macOS 14.5 (23F79). Not sure if the reason behind remove/disabling it is still valid: 292d2dad5489d255749b5e6d3b9b4e3e62ef5d55


cxxxr:
Thank you.
Let's merge it and see how it goes.

PR #1494, opened 2024-08-21
Add `frame-multiplexer-create-with-previous-buffer`

jfaz1:
Adds `frame-multiplexer-create-with-previous-buffer`, bound to `C-z C-C`, that creates a multiplexer frame and sets it to the previous buffer. Useful when you want to open a frame in the same dir. 

Maybe can be replaced with a `frame-multiplexer-duplicate` that copies over the entire frame at some point?


cxxxr:
Thank you

PR #1495, opened 2024-08-24
Add `lem-dashboard`

jfaz1:
This PR adds a customizable dashboard to lem, inspired by how distributions like doom/spacemacs/lazyvim ship out of the box. Besides a more friendly greeting than a `*tmp*` buffer, it ships with useful shortcuts like starting a lisp scratch or seeing recent files/projects at a glance.

image

The dashboard is simply a list of `dashboard-item`s, of which a few are included by default, like a splash/footer message, 
commands/urls, working directory, and recent projects/files.

When making a `dashboard-item`, the only method you need to implement is `draw-dashboard-item`, which is called whenever the dashboard is (re)drawn. 

Let's say you want to add a `dashboard-item` that prints a random fact/quote from an http endpoint, you can make it like this (multithreading is overkill here, just for fun):

```lisp
(defclass dashboard-random-fact (dashboard-item)
  ((fact :initform nil :accessor fact)
   (fetching :initform nil :accessor fetching))
  (:documentation "Creates a dashboard item that displays a random fact"))

(defmethod draw-dashboard-item ((item dashboard-random-fact) point)
  ;; Display initial or current fact
  (insert-string point
                 (create-centered-string (or (fact item) "Loading fact..."))
                 :attribute (item-attribute item))
  (unless (or (fact item) (fetching item))
    (setf (fetching item) t)
    ;; Start a background thread to fetch the fact
    (bt2:make-thread
     (lambda ()
       (handler-case
           (let* ((response (dexador:get "https://uselessfacts.jsph.pl/api/v2/facts/random"))
                  (json-data (json:decode-json-from-string response))
                  (fact-text (cdr (assoc :text json-data))))
             (send-event (lambda () 
                           (setf (fact item) fact-text)
                           (redraw-dashboard))))
         (error (e)
           (send-event (lambda () (message (format nil "Error fetching fact: ~A" e))))))
       :name "random-fact-fetcher"))))
```

To make a custom dashboard layout, all you have to do is call `set-dashboard` and pass a list of items:

```lisp
(in-package :lem-dashboard)

(set-dashboard (list (make-instance 'dashboard-splash
                                    :item-attribute 'document-metadata-attribute
                                    :splash-texts '("Welcome!" "Second splash message!")
                                    :top-margin 4
                                    :bottom-margin 2)
                     (make-instance 'dashboard-working-dir)
                     (make-instance 'dashboard-recent-files 
                                    :file-count 5
                                    :bottom-margin 1)
                     (make-instance 'dashboard-random-fact
                                    :bottom-margin 2
                                    :item-attribute 'document-header3-attribute)
                     (make-instance 'dashboard-command
                                    :display-text " New Lisp Scratch Buffer (l)"
                                    :command 'lem-lisp-mode/internal:lisp-scratch 
                                    :item-attribute 'document-header2-attribute
                                    :bottom-margin 2)))
```

image

Users might want to keep most of the default dashboard's functionality, but customize it a bit (e.g. change the splash, number of projects, etc.). There's a `set-default-dashboard` available to make it easy so they don't have to re-do the whole layout:

```lisp
(lem-dashboard:set-default-dashboard :project-count 10 :file-count 8 :hide-links t :splash '("hello" "hello2"))
```

Notes:

- Set `*dashboard-enable*` to nil to disable
- Only vertical stacking is currently supported. If you want to stack multiple elements horizontally, you'll have to make a dashboard item that does so. At some point I'd like to revisit this, but the way of centering elements here is just adding space padding. I think maybe lem itself should have line attributes, like left-adjust/right-adjust/center. Then we also wouldn't need to redraw on resize.
- I threw in some cheesy footer messages, feel free to add/remove whatever you want from there :laughing: 
- @vindarel let me know if you want me to do the docs for this one


extensions/lem-dashboard/lem-dashboard.lisp
```diff
@@ -0,0 +1,130 @@
...
+(defmethod draw-dashboard-item :around ((item dashboard-item) point)
+  "Inserts a :dashboard-item text property in between the starting and ending POINT, useful for tracking."
+  (let ((start (copy-point point :temporary)))
+    (call-next-method)
+    (let ((end (copy-point point :temporary)))
+      (put-text-property start end :dashboard-item item)
+      (delete-point start)
+      (delete-point end))))
+
+(defvar *dashboard-layout* nil
```
vindarel, line 71: variables and parameters should be together at the top of the file, before functions.

(pretty sure it's a general convention and the one we follow here)

extensions/lem-dashboard/dashboard-items.lisp
```diff
@@ -0,0 +1,147 @@
...
+                            :attribute (item-attribute item)))
+
+;; Recent projects
+(defclass dashboard-recent-projects (dashboard-item)
+  ((project-count :initarg :project-count :accessor project-count :initform *dashboard-project-count*))
+  (:documentation "Displays a list of recent projects, limited to the last PROJECT-COUNT.")
+  (:default-initargs
+   :item-attribute 'document-text-attribute
+   :action (lambda ()
+             (let ((project (string-trim '(#\Space #\Tab) (line-string (current-point)))))
```
vindarel, line 92: see also str:trim if you want to trim more whitespaces: 

```lisp
(defvar *whitespaces* (list #\Backspace #\Tab #\Linefeed #\Newline #\Vt #\Page
                            #\Return #\Space #\Rubout
                            #+sbcl #\Next-Line #-sbcl (code-char 133)
                            #+lispworks #\no-break-space #-lispworks #\No-break_space)
  "On some implementations, linefeed and newline represent the same character (code).")
```

extensions/lem-dashboard/dashboard-items.lisp
```diff
@@ -0,0 +1,147 @@
...
+         (title-line (create-centered-string title))
+         (recent-files (reverse (lem/common/history:history-data-list (lem-core/commands/file:file-history)))))
+    (insert-string point title-line :attribute 'document-header1-attribute)
+    (insert-character point #\Newline)
+    (insert-character point #\Newline)
+    (let* ((longest-file (reduce #'(lambda (a b) (if (> (length a) (length b)) a b)) recent-files))
+           (max-length (length longest-file))
+           (left-padding (floor (- (window-width (current-window)) max-length) 2)))
+      (loop for file in (subseq recent-files 0 (min (file-count item) (length recent-files)))
+            do (insert-string point (format nil "~v@{~A~:*~}" left-padding " "))
```
vindarel, line 146: mmh does `(str:fit left-padding " ")` do the job? Or `make-string`?

vindarel:
Wow, it's super cool :rocket:  

LGTM (only a minor request with the defvar)

vindarel:
user feedback:

- watch out for the logo vertical size. On my screen, not a small one, I can see the logo, the cwd, "recent projects", then I have to scroll to see the rest. Not very ergonomic.
- have the cursor on a meaningful by default?
- have the ascii logo for ncurses, but the beautiful image in the GUI?
- more shortcuts like M-n and M-p to move between items?

last but not least:

- it may be a big win to display a short help, like C-x C-f to open files, M-x (alt-x) help to read some help or M-x documentation-describe-bindings.
  - the *link* to the online manual is great but not self-contained.

vindarel:
> @vindarel let me know if you want me to do the docs for this one

accepted, thanks. I can see a short description with screenshot in the usage page, then link to a more detailed page of its own.

Also, what about a `dashboard-help`? (where we see the lack of more self-documentation features)

cxxxr:
That's great!

It just seems a little buggy.
For example, I get the following error at startup

```
invalid number of arguments: 0
Backtrace for: #<SB-THREAD:THREAD tid=3075 "editor" RUNNING {7005B23643}>
0: ((LAMBDA (LEM-DASHBOARD::A LEM-DASHBOARD::B) :IN LEM-DASHBOARD::DRAW-DASHBOARD-ITEM)) [external]
1: (REDUCE #<FUNCTION (LAMBDA (LEM-DASHBOARD::A LEM-DASHBOARD::B) :IN LEM-DASHBOARD::DRAW-DASHBOARD-ITEM) {700CF7260B}> NIL)
2: ((:METHOD LEM-DASHBOARD::DRAW-DASHBOARD-ITEM (LEM-DASHBOARD::DASHBOARD-RECENT-PROJECTS T)) #<LEM-DASHBOARD::DASHBOARD-RECENT-PROJECTS {700CFD9353}> #<LEM-CORE:CURSOR (30, 0) "" {700CC28693}>) [fast-method]
3: ((SB-PCL::EMF LEM-DASHBOARD::DRAW-DASHBOARD-ITEM) #<unused argument> #<unused argument> #<LEM-DASHBOARD::DASHBOARD-RECENT-PROJECTS {700CFD9353}> #<LEM-CORE:CURSOR (30, 0) "" {700CC28693}>)
4: ((:METHOD LEM-DASHBOARD::DRAW-DASHBOARD-ITEM :AROUND (LEM-DASHBOARD::DASHBOARD-ITEM T)) #<LEM-DASHBOARD::DASHBOARD-RECENT-PROJECTS {700CFD9353}> #<LEM-CORE:CURSOR (30, 0) "" {700CC28693}>) [fast-method]
5: (LEM-DASHBOARD::REDRAW-DASHBOARD)
6: (LEM-DASHBOARD:OPEN-DASHBOARD)
7: (SB-INT:SIMPLE-EVAL-IN-LEXENV (FUNCALL #<FUNCTION LEM-DASHBOARD:OPEN-DASHBOARD>) #<NULL-LEXENV>)
8: (EVAL (FUNCALL #<FUNCTION LEM-DASHBOARD:OPEN-DASHBOARD>))
9: (LEM-CORE::APPLY-ARGS #S(LEM-CORE::COMMAND-LINE-ARGUMENTS :ARGS (#1="~/.lem/log" (FUNCALL #<FUNCTION LEM-DASHBOARD:OPEN-DASHBOARD>)) :DEBUG NIL :LOG-FILENAME #1# :NO-INIT-FILE NIL :INTERFACE NIL))
10: (LEM-CORE::TOPLEVEL-COMMAND-LOOP #<FUNCTION (LAMBDA NIL :IN LEM-CORE::RUN-EDITOR-THREAD) {700B0E58FB}>)
11: ((LAMBDA NIL :IN LEM-CORE::RUN-EDITOR-THREAD))
12: ((LAMBDA NIL :IN LEM-CORE::RUN-EDITOR-THREAD))
13: ((FLET BORDEAUX-THREADS-2::RUN-FUNCTION :IN BORDEAUX-THREADS-2::ESTABLISH-DYNAMIC-ENV))
14: ((LABELS BORDEAUX-THREADS-2::%ESTABLISH-DYNAMIC-ENV-WRAPPER :IN BORDEAUX-THREADS-2::ESTABLISH-DYNAMIC-ENV))
15: ((FLET SB-UNIX::BODY :IN SB-THREAD::RUN))
16: ((FLET "WITHOUT-INTERRUPTS-BODY-" :IN SB-THREAD::RUN))
17: ((FLET SB-UNIX::BODY :IN SB-THREAD::RUN))
18: ((FLET "WITHOUT-INTERRUPTS-BODY-" :IN SB-THREAD::RUN))
19: (SB-THREAD::RUN)
```

cxxxr:
Can we use buttons for dashboard-item?
https://github.com/lem-project/lem/blob/main/src/ext/button.lisp

This allows for common mouse clicks, etc.

extensions/lem-dashboard/lem-dashboard.lisp
```diff
@@ -0,0 +1,128 @@
...
+  "Redraws the dashboard, clearing and redrawing all content while attempting to maintain point position."
+  (let* ((buffer (create-dashboard-buffer))
+         (old-line (line-number-at-point (buffer-point buffer)))
+         (old-column (point-column (buffer-point buffer))))
+    (setf (buffer-read-only-p buffer) nil)
+    (erase-buffer buffer)
+    (let ((point (buffer-point buffer)))
+      (dolist (item *dashboard-layout*)
+        (draw-dashboard-item item point)))
+    (setf (buffer-read-only-p buffer) t)
```
cxxxr, line 95: could you use `with-buffer-read-only`

extensions/lem-dashboard/lem-dashboard.lisp
```diff
@@ -0,0 +1,128 @@
...
+    (change-buffer-mode buffer 'dashboard-mode)
+    (move-to-line (buffer-point buffer) old-line)
+    (move-to-column (buffer-point buffer) old-column)))
```
cxxxr, line 98: Was there a problem with save-excursion?

extensions/lem-dashboard/dashboard-items.lisp
```diff
@@ -0,0 +1,147 @@
...
+  ((file-count :initarg :file-count :accessor file-count :initform *dashboard-file-count*))
+  (:documentation "Displays a list of recent files, limited to the last FILE-COUNT.")
+  (:default-initargs
+   :item-attribute 'document-text-attribute
+   :action (lambda ()
+             (let ((file (str:trim (line-string (current-point)))))
+               (when file
+                 (find-file file))))))
+
+(define-command move-to-recent-files () ()
```
cxxxr, line 128: Since define-command has a global namespace, it would be better to have a dedicated prefix.

extensions/lem-dashboard/lem-dashboard.lisp
```diff
@@ -0,0 +1,128 @@
...
+      (funcall (action item)))))
+
+(defmethod draw-dashboard-item :around ((item dashboard-item) point)
+  "Inserts a :dashboard-item text property in between the starting and ending POINT, useful for tracking."
+  (let ((start (copy-point point :temporary)))
+    (call-next-method)
+    (let ((end (copy-point point :temporary)))
+      (put-text-property start end :dashboard-item item)
+      (delete-point start)
+      (delete-point end))))
```
cxxxr, line 72: Can we use buttons for dashboard-item?
https://github.com/lem-project/lem/blob/main/src/ext/button.lisp

This allows for common mouse clicks, etc.

cxxxr:
Your screenshot seems to be different from mine.
Do you have a special setting?
スクリーンショット 2024-08-25 16 35 15

extensions/lem-dashboard/dashboard-items.lisp
```diff
@@ -0,0 +1,147 @@
...
+  (:documentation "Creates link/button with DISPLAY-TEXT that opens URL externally.")
+  (:default-initargs
+   :item-attribute 'document-text-attribute))
+
+(defmethod initialize-instance :after ((item dashboard-url) &key)
+  (with-slots (keybind-command action) item
+    (setf action (lambda () (funcall keybind-command)))))
+
+(defmethod draw-dashboard-item ((item dashboard-url) point)
+  (lem/button:insert-button point
```
cxxxr, line 34: It might be better to use local-nicknames

extensions/lem-dashboard/dashboard-items.lisp
```diff
@@ -0,0 +1,147 @@
...
+  (let ((point (buffer-point (current-buffer))))
+    (buffer-start point)
+    (search-forward-regexp point "Recent Files")
+    (line-offset point 2)
+    (move-to-beginning-of-line)))
+
+(defmethod draw-dashboard-item ((item dashboard-recent-files) point)
+  (let* ((title (format nil "~A Recent Files (f)" (icon-string "file-text")))
+         (title-line (create-centered-string title))
+         (recent-files (reverse (lem/common/history:history-data-list (lem-core/commands/file:file-history)))))
```
cxxxr, line 138: It might be better to use local-nicknames

cxxxr (commented):
I commented on a few things that bothered me.
Nevertheless, I think this feature and implementation is excellent.
Thank you.

jfaz1:
> Your screenshot seems to be different from mine. Do you have a special setting? スクリーンショット 2024-08-25 16 35 15

Hmm strange it should be the same (except for the smaller ascii art)... I'll look into it now. That's the default dashboard?

> For example, I get the following error at startup,,

Bah my bad, I lost logic I had there in a commit I reverted. Thanks for pointing that out.

extensions/lem-dashboard/lem-dashboard.lisp
jfaz1: It didn't work for me... I might be using it wrong. Here's the redraw function, works fine (keeps cursor position when resizing window)

```lisp
(defun redraw-dashboard ()
  (let* ((buffer (create-dashboard-buffer))
         (old-line (line-number-at-point (buffer-point buffer)))
         (old-column (point-column (buffer-point buffer))))
    (with-buffer-read-only buffer nil
      (erase-buffer buffer)
      (let ((point (buffer-point buffer)))
        (dolist (item *dashboard-layout*)
          (draw-dashboard-item item point)))
      (change-buffer-mode buffer 'dashboard-mode)
      (move-to-line (buffer-point buffer) old-line)
      (move-to-column (buffer-point buffer) old-column))))
```

Here's after changing it:

```lisp
(defun redraw-dashboard ()
  (let* ((buffer (create-dashboard-buffer)))
    (save-excursion
      (with-buffer-read-only buffer nil
        (erase-buffer buffer)
        (let ((point (buffer-point buffer)))
          (dolist (item *dashboard-layout*)
            (draw-dashboard-item item point)))
        (change-buffer-mode buffer 'dashboard-mode)))))
```

After making that change, if I resize the window it'll reset it to the top. I tried using it in a few other spots and couldn't get it to work.

extensions/lem-dashboard/lem-dashboard.lisp
jfaz1: I was gonna do that but here are some issues with making the base `dashboard-item` class a button:

- This bit sets the entire `dashboard-item` region to a button. That's fine for some items, like commands/urls (which are already buttons), but not for projects/files, since it would make the entire region a single button. We probably want each individual project/file to be its own button instead of one big one.
- Some items might not want to be buttons (e.g. splash/footer)

cxxxr: OK, I got it.

cxxxr (approved):

extensions/lem-dashboard/default-dashboard.lisp
```diff
@@ -0,0 +1,93 @@
...
+                                         :display-text " Getting Started (s)"
+                                         :url "https://lem-project.github.io/usage/usage/"
+                                         :item-attribute 'document-header3-attribute)
+                          (make-instance 'dashboard-url
+                                         :display-text " GitHub (g)"
+                                         :url "https://github.com/lem-project/lem"
+                                         :item-attribute 'document-header3-attribute
+                                         :bottom-margin 2))))
+      (define-key *dashboard-mode-keymap* "s" 'open-lem-docs)
+      (define-key *dashboard-mode-keymap* "g" 'open-lem-github))
```
cxxxr, line 79: I just noticed this code, is there any reason why it is not defined at the top level?

jfaz1: The way I had bindings previously worked a bit differently. The idea was to not apply the keybind until it's used in case the user has a dashboard that doesn't contain an item like the docs/github link. That being said, after changing the way the default dashboard is created, it doesn't work since it gets called at compile time, so further calls to `set-default-dashboard` won't remove the stale binds. 

I went ahead and moved it to the top-level, users can just replace the stale bind in their config if they want to bind something else there.

jfaz1:
@cxxxr I see what vindarel was talking about with the default font making the ascii art look too big. I can try making it smaller.

cxxxr:
I think it's good

vindarel:
User feedback:

- (minor) ncurses version, I see "NIL recent projects (r)": "NIL" instead of an icon or a character
- bug: on any line of the recent projects, pressing Enter makes me open… files from the cwd, not the project my cursor is on.

jfaz1:
> * (minor) ncurses version, I see "NIL recent projects (r)": "NIL" instead of an icon or a character

Hmm I don't see that, I'm guessing I just happen to have the font it needs installed? Is there a way to detect if the font is available and then just not show it if it's not? I guess worst-case scenario I can omit icons if on ncurses.

> * bug: on any line of the recent projects, pressing Enter makes me open… files from the cwd, not the project my cursor is on.

Whoops, that's my bad. I tested in same folder so I missed that :laughing: The `arg` in `project-find-file` is unused so I had to switch first. Fixed now, thanks.

vindarel:
alright the PR looks great to me and cxxxr approved the changes. @jfaz1 you're done, I merge?

cxxxr:
Looks good, so I'll merge it.

PR #1502, opened 2024-08-29
Use buttons for theme list

jfaz1:
Makes themes clickable in the theme list. Also moves the theme list to its own file and adds a `ui` module.


cxxxr:
Thank you.

PR #1511, opened 2024-09-02
Add `require-final-newline`

jfaz1:
POSIX standard is that text files end in a newline. Currently, lem doesn't have a way of managing this behavior besides relying on a formatter and enabling `auto-format`. 

I think this behavior shouldn't be handled by the formatter given it's the standard for text files in general. There are a lot of ways to handle this but I came upon this one while playing around and thought I should push it up as a start...

Basically, I added a `require-final-newline` slot to `major-mode`, enabled by default. This way modes themselves dictate whether a file should end in a newline, instead of being a global toggle like in some other editors. I also disabled it for fundamental mode as a precaution in case you edit a binary file for whatever reason and save it.

Let me know what you guys think or if there are any (saveable) modes where this behavior should be disabled by default. Or maybe there's a better way of doing this.

### DIFF
--- a/src/fundamental-mode.lisp
+++ b/src/fundamental-mode.lisp
@@ -1,7 +1,8 @@
 (in-package :lem-core)
 
 (define-major-mode lem/buffer/fundamental-mode:fundamental-mode nil
-    (:name "Fundamental"))
+    (:name "Fundamental"
+     :require-final-newline nil))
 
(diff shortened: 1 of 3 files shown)


jfaz1:
Oh I just noticed you can pass in `scope` to `variable-value`. I think using buffer local variables is neater; closing this.

PR #1516, opened 2024-09-05
lsp-mode: update client-capabilities.json for code actions

alanz:
Rust analyzer will not return any code actions unless `codeActionLiteralSupport` is enabled.

This adds it to the default client capabilities, based on what is configured in emacs lsp-mode, see

https://github.com/emacs-lsp/lsp-mode/blob/c36b95be6625dac5a37d3874a1a738e0c84ac39f/lsp-mode.el#L3731-L3738

With this change, they do show up provided you run `M-x lsp-code-action` when in a valid location.


cxxxr:
Thank you

PR #1517, opened 2024-09-06
WIP: more multiple-cursor commands

Book-reader:
this pr adds more commands to `multiple-cursors.lisp` that emulates the `mark-next/previous-like-this` functionality from [multiple-cursors.el](https://github.com/magnars/multiple-cursors.el) as well as adding a `add-cursors-to-previous-line` command to match the `add-cursors-to-next-line` command.

[example.webm](https://github.com/user-attachments/assets/c76a1ce7-e6ba-474b-bd54-7c618d2e30d7)

things I have not done yet:
 - [x] fix `mark-previous-like-this` not working at all
 - [ ] fix `mark-previous-like-this` skipping every second match when cursor is to the left of the mark (buffer-point < buffer-mark)
 - [x] fix `mark-next-like-this` always placing fake cursor on the right side of the mark, even when the real cursor is on the left side
   - [x] find a way to start of the mark as the first point marked instead of the top-left always being the start of the mark.
 - [X] allow multiple cursors to be on the same line (now cannot be on the same char)
 - [ ] add tests
 - [ ] add default keyboard shortcuts
 - [ ] add description to `mark-previous-like-this` and `mark-next-like-this`
 - [X] make `mark-next/previous-like-this` run `add-cursors-to-next/previous-line` when nothing is marked (to match `multiple-cursors.el`)
 - [X] allow multiple fake cursors to be on the same line, but not on the same point
 - [X] fix `mark-next-like-this` not working with multiple lines
   - [X] initial support
   - [X] fix only deleting the last line marked by the fake cursor (probably something to do with not setting point-line)
   - [X] fix not working when mark does not start at the beginning of the line
 - [x] fix multiple cursors being in the same space until something is typed, duplicating the letter typed

issues with multiple cursors out of scope of this PR:
 - once you mark something with a fake cursor it has a phantom mark that stays until the fake cursor is removed or until something else is marked.
 - undoing an edit done by a fake cursor causes the real cursor to jump to that point.
 - in `multiple-cursors.el`, the fake cursors stay when using `keyboard-escape-quit` and are only cleared with `keyboard-quit`. this allows calling `keyboard-escape-quit` to unmark the current buffer without removing all the fake cursors. in lem there is only `keyboard-quit` which both unmarks the buffer and removes all the cursors.

I am not very good with lisp so please tell me if there are any problems with my code.
I have moved `multiple-cursors.lisp` to `extensions/` because `add-cursors-to-next-like-this` uses isearch and I could not figure out how to import the needed functions and commands when it was in lem-core.


Book-reader:
I'm a little confused about how to check if there is a mark in the current buffer, `check-marked` uses `(unless (buffer-mark (current-buffer))` but `(buffer-mark (current-buffer)` only returns `nil`  when a buffer has never been marked, and after marking a buffer once will always be non-nil even when the buffer is no longer marked.

[phantom-mark.webm](https://github.com/user-attachments/assets/ffc12e65-de98-4ee1-81c3-d473071c9377)

Book-reader:
> I'm a little confused about how to check if there is a mark in the current buffer, `check-marked` uses `(unless (buffer-mark (current-buffer))` but `(buffer-mark (current-buffer)` only returns `nil` when a buffer has never been marked, and after marking a buffer once will always be non-nil even when the buffer is no longer marked.
> phantom-mark.webm

ok, `(buffer-mark-p (current-buffer))` works

PR #1525, opened 2024-09-10
Readded keymap export

tetrazolium:


tetrazolium:
I'm having some trouble with the tests. However, I'm getting the same corruption and memory fault issues even when I run the tests on a fresh clone of the current repository.

Am I doing something wrong? I've never really used CI before. Also I'm not sure how such a simple change could fail.

cxxxr:
The reason CI is down is me.

@vindarel What do you think of this PR?

PR #1538, opened 2024-09-22
Legit closify remove global vcs

vindarel:
Legit's code base is now a little OO and the VCS logic is split into their own files.

The goal that is now achieved without bugs is that we can have more than 1 interactive rebase in process, without re-inventing an OO with too many global variables and manual dispatch.

I tested the best I can, specialy the interactive rebase, for git and other VCSs (which must show a message, not fail).

replaces https://github.com/lem-project/lem/pull/1451


vindarel:
the failed CI is

```

;; testing 'lem-tests/language-server/tests'
initialize
CORRUPTION WARNING in SBCL pid 4134 tid 4187:
Memory fault
```

PR #1550, opened 2024-09-30
Fix lisp completion current argument highlighting

jfaz1:
Fix highlighting the current argument:

image

`(change-buffer-mode buffer 'lisp-mode)` was overriding the attribute, so I set it back to fundamental mode before we apply the highlight.


cxxxr:
Thank you!

PR #1551, opened 2024-10-01
Minor grep improvements

jfaz1:
Adds a couple minor grep improvements:

- `n`/`p` for navigation
- `a` takes you to the start of the content in the current line


cxxxr:
There is a function that allows you to edit the source when editing the Grep buffer, but doesn't that conflict with it?

jfaz1:
@cxxxr it's only enabled when on the filename:
```lisp
                      (insert-string point file :attribute 'lem/peek-source:filename-attribute
                                     :mode 'peek-grep-mode
                                     :read-only t)
```

cxxxr:
Ah, I see.
that's very nice.

PR #1555, opened 2024-10-02
List listener history filtering by prefix

naryl:
Use like this (in `init.lisp`):

```lisp
(define-key lem/listener-mode:*listener-mode-keymap* "M-p" 'lem/listener-mode:listener-previous-startswith-input)
(define-key lem/listener-mode:*listener-mode-keymap* "M-n" 'lem/listener-mode:listener-next-startswith-input)
```

This is the behaviour I currently have in emacs. No idea where it came from but I like it so I implemented it for Lem too.

### DIFF
--- a/src/ext/listener-mode.lisp
+++ b/src/ext/listener-mode.lisp
@@ -21,6 +21,8 @@
    :listener-return
    :listener-previous-input
    :listener-next-input
+   :listener-previous-startswith-input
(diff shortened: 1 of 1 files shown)


cxxxr:
This is very convenient, thank you.
Incidentally, we can use M-r to search your history.

rpx99:
I think that's a general problem. Buried knowledge

Issue #1556, opened 2024-10-02
`define-motion` commands seem to always get 1 as default `n`


naryl:
Trying to implement proper vi-like C-u/C-d scrolling, I'm getting this bug with the following code:
```lisp
(define-motion vi-scroll-up (&optional (n nil)) (:universal)
    (:default-n-arg nil)
  ;; Somehow n is still 1 by default...
  (when (eql n 1)
    (setf n (floor (/ (window-height (current-window)) 2))))
  (previous-line n)
  (scroll-up n))
```

naryl:
Tried setting both `&optional` and `:default-n-arg` to 0 or 2. `n` is still 1 when called without the argument.

naryl:
`arg-descriptor` should be `:universal-nil` instead of `:universal`.

PR #1559, opened 2024-10-02
Vi fixes

naryl:
Fixing some discrepancies with vim as I stumble upon them.

Fixed:
* `C-u`/`C-d` should scroll half a page by default and preserve cursor location relative to window.
* `%` when not on a paren should find the following paren on the same line and then do its jump.

Found, not fixed:
* ` is not implemented.
* `()` and `{}` motions are not implemented.


naryl:
`C-u`/`C-d` depends on this: https://github.com/lem-project/lem/issues/1556

naryl:
`{}` are implemented. Work with paredit-mode like this:
```lisp
(define-key lem-paredit-mode:*paredit-mode-keymap* "{" nil)
(define-key lem-paredit-mode:*paredit-mode-keymap* "}" nil)
(define-key lem-vi-mode:*insert-keymap* "{" 'lem-paredit-mode:paredit-insert-brace)
(define-key lem-vi-mode:*insert-keymap* "}" 'lem-paredit-mode:paredit-close-brace)
```

I feel like there should be a better way to make vi-mode and paredit-mode be nice to each other.

extensions/vi-mode/commands.lisp
```diff
@@ -689,14 +705,23 @@ Move the cursor to the first non-blank character of the line."
 (defun vi-forward-matching-paren (window point &optional (offset -1))
   (declare (ignore window))
   (with-point ((point point))
+    (loop :until (or (syntax-open-paren-char-p (character-at point))
+                     (syntax-closed-paren-char-p (character-at point))
+                     (= (point-charpos point) (length (line-string point))))
+          :do (incf (point-charpos point)))
```
cxxxr, line 711: Oh, I realized after merging that changing point-charpos breaks encapsulation.
I think it would be better to use line-end-p and character-offset instead.

naryl: Ok, I'll do it within 24h. Thanks!

naryl: Would it still a problem with temporary point? Like `(with-point ((point (copy-point point))) ...)`

cxxxr: yes, but copy-point is not necessary because it is used automatically with-point.
`(with-point ((point point)) ...)`

naryl: `(apropos "line-end-p")` doesn't find anything. There's a struct `line-end-item-p` but I don't think it's what I need here.

naryl: Oh, you mean `(line-end p)`. :)

cxxxr: Sorry, it was end-line-p.

PR #1562, opened 2024-10-05
Make use of the paren-finding logic only on %

naryl:
Followup to https://github.com/lem-project/lem/pull/1559

1. `point` encapsulation.
2. It changed `forward-matching-paren` and `backward-matching-paren` behaviour too. Restored it.


cxxxr:
Thank you

PR #1563, opened 2024-10-06
Make slurp and close-paren work as Emacs's paredit

naryl:
* Slurp should remove all whitespace around origin point if there were no non-whitespace characters in the list before slurping.
* close-paren should remove all whitespace characters before closed paren and always skip *exactly one* paren.
* paredit-splice-forward was unbound.

That's what Emacs's paredit does but it may not always be best. I'll leave this PR for discussion for a few days.


mmontone:
> That's what Emacs's paredit does but it may not always be best.

When I'm in doubt I make things configurable.

naryl:
> > That's what Emacs's paredit does but it may not always be best.
> 
> When I'm in doubt I make things configurable.

Do you think it's enough to make a single setting to "make paredit agressively remove extra whitespace"?

mmontone:
Yes IMO

naryl:
What's the recommended way? `config` and `setf config`? `defvar`?

mmontone:
define-attribute I believe
https://github.com/lem-project/lem/blob/80ab76c2946569a8a009fa0c0dcd29d29d9190a4/src/attribute.lisp#L151

naryl:
Looks like attributes are for styling text.

mmontone:
Yes. I don't know then. I guess it may be defvar. I don't know if other way
should be used.

naryl:
I checked some example configs and none use the `config` facility. So I guess it's plain old `defvar`.

mmontone:
What is the config facility? Where is it implemented? I'm thinking of
trying an implementation myself...

mmontone:
Ah. I see the `config` facility now. I don't know if you should use it or not, but it persists the config changes, and that could be useful. But it is not clear to me, perhaps a defvar that can be tuned in init file by the user.

naryl:
@cxxxr Github shows that you added config.lisp about a year ago. What was its intended use? I don't see it used in either vindarel's or sasanidas's `init.lisp` (linked at the bottom here: https://github.com/lem-project/lem/wiki/Configuring-Lem)

mmontone:
You can see where it is used if you eval command `find-references` over `lem-core::config`. But I don't know if you should use it in this case.

cxxxr:
@naryl It is used to persist variables.
For example, the isearch string, access token, etc.

cxxxr:
@naryl 
> EDIT: Or more specifically, should I use it in this case?

In this case, I think a special variable would be fine.

naryl:
@cxxxr Yep, that's what I did. https://github.com/lem-project/lem/pull/1563/files#diff-364eb1deeea9a0c8bb551bb59f2e0ca1c03190b6c297f7114a2a3f5acb29b80bR40

extensions/paredit-mode/paredit-mode.lisp
```diff
@@ -309,16 +312,19 @@ link : http://www.daregada.sakuraweb.com/paredit_tutorial_ja.html
...
+           (forward-char))
           ((ignore-errors (or (scan-lists p 1 1)) t)
            (with-point ((new-p p))
              (character-offset new-p -1)
              (move-point (current-point) new-p)
              (with-point ((p new-p))
                (skip-whitespace-backward p)
-               (delete-between-points p new-p))))
+               (delete-between-points p new-p)
+               (forward-char))))
```
cxxxr, line 327: In this case, it is better to use character-offset than forward-char.

```lisp
(character-offset p 1)
```

naryl: It was already used above in the same function. I guess I'll change both to `character-offset`.

cxxxr:
Thank you for the nice PR.

PR #1568, opened 2024-10-09
Customize

mmontone:
I'm submitting this not for merge, but for a review, and to showcase its current implementation.
To load, no need to recompile Lem, just load the `src/customize.lisp` file from a Lem editor, and then evaluate `M-x customize`.
Any feedback welcomed!
Thanks.

imagen

### DIFF
--- a/src/customize.lisp
+++ b/src/customize.lisp
@@ -1,6 +1,6 @@
 ;; https://www.gnu.org/software/emacs/manual/html_node/elisp/Customization.html
 
-(defpackage :lem/settings
+(defpackage :lem/customize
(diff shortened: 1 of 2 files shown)


cianoc:
This is cool, and it would definitely help with onboarding new users. Have you made any more progress on it?

mmontone:
> This is cool, and it would definitely help with onboarding new users. Have you made any more progress on it?

I haven't looked at this since I created it in 2024, but I could have a look and see where it stands.
You could try it yourself, also, in theory. Try loading customize.lisp file.

PR #1570, opened 2024-10-17
Basic support for ruby language

sl33pii:
This commit has really basic support for the ruby language, this includes:
- Syntax highlighting
- indentation

I reused a lot of the logic implemented in elixir-mode, but i changed the language *keywords* and the way the indentation is calculated, which could be better, but it seems to work alright from what i've tested.

I'm still working in adding LSP support for the language, but if this is ok, i guess that could be merged in another branch, since i don't know how long it might take me or if anyone else wants to work on it.

I appreciate any feedback/improvement btw.


cxxxr:
I think it's very good.
Thank you!

vindarel:
and a basic page: https://lem-project.github.io/modes/ruby/

PR #1578, opened 2024-10-25
Enable fractional scaling on all platforms

wakira:
HiDPI support for MacOS was enabled via #799. 
This PR just removes the `#+darwin` restriction.

It seems that the original concern (https://github.com/lem-project/lem/pull/799#issuecomment-1614346368) is no longer a problem.

Tested on Linux KDE Plasma 6.2.2
- [X] no change when running on the default X backend
- [X] no more blurry text when running on wayland backend with SDL_VIDEODRIVER=wayland


cxxxr:
Thank you very much!

PR #1579, opened 2024-10-30
Add Nix flake

KitRedgrave:
This PR adds a Nix flake that can build lem-ncurses and lem-sdl2 on both Linux and macOS.

This is mostly code from https://github.com/dariof4/lem-flake, with some polishing and updating.

Currently all of the dependencies are manually tracked, which is a bit of a maintenance chore. Maybe there's some way to automate this bit with ql2nix or something.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,6 @@
+(import (let lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+in fetchTarball {
+  url =
+    lock.nodes.flake-compat.locked.url or "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+  sha256 = lock.nodes.flake-compat.locked.narHash;
+}) { src = ./.; }).defaultNix
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,75 @@
+{
+  "nodes": {
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1696426674,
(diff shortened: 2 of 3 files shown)


KitRedgrave:
Some things are probably still broken because Quicklisp isn't loaded... there's at least one patch missing, and I know that currently `M-x load-library` fails because of the dependency.

cxxxr:
I've installed nixos, and I'll merge it in once because it's very helpful to have it.

KitRedgrave:
> Thank you for the wonderful PR. I recently installed nix and am reading tutorials, so I am very happy about this PR.
> 
> I also have few questions.
> 
> Could you please tell me how to use this? I built it with `nix --extra-experimental-features 'nix-command flakes' build .#lem-ncurses` and confirmed that I could start it with result/bin/lem Is this method correct? Or is there another way? If there is a PR description or README for this, that would be great.
> 
> Can you maintain this nix continuously? I don't know much about nix yet, so I can't fix this. As you say, it would be preferable if it could be automatically generated using information such as qlfile.lock, using ql2nix, etc.
> 
> I'm glad to see that lem is becoming more popular among nix users. I hope that it will be possible to install lem from the official nix site.

This comes a little late since you've already merged this, but indeed that works for building it. You can also build this with the older `nix-build` command. To do so for x86_64-linux lem_sdl2, use `nix-build -A packages.x86_64-linux.lem-sdl2

vindarel:
I everyone, where can you document it?

wiki, website, README?

https://github.com/lem-project/lem/wiki

https://lem-project.github.io/installation/ncurses/linux/

thanks

PR #1593, opened 2024-11-10
Nix: use qlot bundle instead of manual tracking

KitRedgrave:
This PR replaces the manual dependency pinning with a build process that uses `qlot bundle` to handle downloading dependencies based on the `qllock` file.


cxxxr:
Thank you

PR #1601, opened 2024-11-12
Update isearch.lisp

SequentialDesign:
fix `isearch-next-highlight` and `isearch-prev-highlight` so that numeric prefixes work correctly.  

in the original code, Shift+F3 skips a highlight, a prefix argument of 0 (zero) goes to the previous highlight with `prev-highlight` and `next-highlight`, and prefix arguments don't work properly when the point is going backwards because they skip a highlight


cxxxr:
Thanks

PR #1613, opened 2024-11-19
vi-mode: add ignorecase option

cxxxr:


sakurawald:
I think the :ignorecase option for `regex scanner` is good.
The default `isearch` function in `vim editor` is also based on `regex`.

Issue #1617, opened 2024-11-20
improve: should not disable the kwin compositor by the default sdl2 hints.


sakurawald:
Related issue: https://github.com/lem-project/lem/issues/1608

The related articles:
1. https://kde-bugs-dist.kde.narkive.com/ti0RQpQv/kwin-bug-371480-new-sdl2-apps-causing-desktop-effects-to-disappear-task-manager-function-weirdly
2. https://discourse.libsdl.org/t/why-is-compositing-disabled-on-normal-x11-windows/37712/3

PR #1618, opened 2024-11-20
fix: sdl2 should not disable the kwin compositor, which will cause the window effects invalid, including the window opacity.

sakurawald:
To fix issue: https://github.com/lem-project/lem/issues/1617


cxxxr:
Thanks

PR #1622, opened 2024-11-20
vindarel/legit stash: push, pop, list, show, drop

vindarel:


vindarel:
ready to merge,

beware that I changed the message timeout from 1s to 2s.

Issue #1583, opened 2024-11-03
feature: new command `maximize-frame`


sakurawald:
Add new commands like `maximize-frame` to maximize the editor automatically at startup.
Similar: https://github.com/lem-project/lem/blob/main/src/commands/frame.lisp

sakurawald:
To start up lem editor and maximize it, append this to `init.lisp`:
```
(lem-sdl2/frame:maximize-frame)
```

PR #1624, opened 2024-11-20
feature: add `maximize-frame` and `minimize-frame` command for lem sdl2 version.

sakurawald:
The issue: https://github.com/lem-project/lem/issues/1583

### DIFF
--- a/frontends/sdl2/frame.lisp
+++ b/frontends/sdl2/frame.lisp
@@ -1,0 +1,14 @@
+(defpackage :lem-sdl2/frame
+  (:use :cl :lem)
+  (:export :maximize-frame
+           :minimize-frame))
+(in-package :lem-sdl2/frame)
+
+(define-command maximize-frame () ()
+  "Maximize the frame."
+  (sdl2:maximize-window (lem-sdl2/display::display-window (lem-sdl2/display:current-display))))
+
+(define-command minimize-frame () ()
+  "Minimize the frame."
+  (sdl2:minimize-window (lem-sdl2/display::display-window (lem-sdl2/display:current-display))))
+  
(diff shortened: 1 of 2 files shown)


sakurawald:
To start up lem editor and maximize it, append this to `init.lisp`:
```
(lem-sdl2/frame:maximize-frame)
```

cxxxr:
That's good.
If I had to say something, I'd like to put it in the same package as toggle-frame-fullscreen and unify the interface.
I also want to make it compatible with other than SDL2.

https://github.com/lem-project/lem/pull/831/files
Wouldn't it be nice to define the interface like the PR of the link destination?

sakurawald:
> I tried it on a macos, but it seems that the process will end abnormally if you don't use sdl2:in-main-thread.
> 
> I was able to get it working on macOS by doing the following
> 
> ```commonlisp
> (sdl2:in-main-thread ()
>     (sdl2:maximize-window (lem-sdl2/display::display-window (lem-sdl2/display:current-display))))
> ```

Hah, sorry, i forget about this. 
It works well in `linux`, i am not very sure about it in other platforms.

I have modified the source, and you can help me modify it if anything is wrong.

cxxxr:
Thank you.
Can I merge it and then make changes?
I'll move the command definitions to lem-core/commands/frame.

sakurawald:
> Thank you. Can I merge it and then make changes? I'll move the command definitions to lem-core/commands/frame.

Sure!

sakurawald:
One more question: I create the `frame.lisp` file in `frontends/sdl2/frame.lisp`, it's because the usage of `sdl2` package.
I didn't define the commands in `src`, because i don't know if it's not sdl2 version.

PR #1627, opened 2024-11-21
Vindarel/detective see macros

vindarel:
M-x detective-all didn't suggest macros

we also add M-x detective-macros


vindarel:
(rebased on main)

Issue #1621, opened 2024-11-20
feature: add the missing binding for `~` character in `vi-mode`.


sakurawald:
In `vim edtiro`, the `~` character is bind to `swapcase and forward next character`.

In `lem editor`, there is a command named `lem-vi-mode/binds::vi-swapcase`, however, this command requires a motion action, which doesn't fit the need.

There should be a new command to do something like `swap-case on cursor, and forward character`.

And add the missing binding:
```
(define-key lem-vi-mode:*normal-keymap* "~" '<the new command name>)
```

cxxxr:
If you like, please make a PR.

PR #1629, opened 2024-11-22
feature: new command `vi-swapcase-and-forward-char` for vi-mode.

sakurawald:
To close the issue: https://github.com/lem-project/lem/issues/1621

The command `vi-swapcase-and-forward-char` is used as a **shortcut** to swap-case on cursor and forward char in **vi-normal mode**.
The `vi-swapcase` is an **vi-operator**, which requires an extract **motion action**, to decide the region of swap-case, which may be an over-kill for some simple-case.

If you only want to swapcase for a consecutive characters, just simply press the `~` key, and this new command will swap the case of character, move the cursor to forward character, repeatly.

### DIFF
--- a/extensions/legit/legit-commit.lisp
+++ b/extensions/legit/legit-commit.lisp
@@ -27,6 +27,9 @@ Future:
   ;; no syntax highlihgt in fact.
   (setf (variable-value 'enable-syntax-highlight) t))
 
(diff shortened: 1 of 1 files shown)


sakurawald:
For some reason, the file is re-formatted by the formatter in my env.
I don't know if this is correct. I am using the default code formatter provided by lem editor.

PR #1632, opened 2024-11-22
feature: add vi-swapcase-and-forward-char command for vi-mode

sakurawald:
To close the issue: https://github.com/lem-project/lem/issues/1621

The command `vi-swapcase-and-forward-char` is used as a **shortcut** to swap-case on cursor and forward char in **vi-normal mode**.
The `vi-swapcase` is an **vi-operator**, which requires an extract **motion action**, to decide the region of swap-case, which may be an over-kill for some simple-case.

If you only want to swapcase for consecutive characters, just simply press the `~` key, and this new command will swap the case of character, move the cursor to forward character, repeatly.

### DIFF
--- a/extensions/vi-mode/commands.lisp
+++ b/extensions/vi-mode/commands.lisp
@@ -90,6 +90,7 @@
            :vi-upcase
            :vi-downcase
            :vi-swapcase
+           :vi-swapcase-and-forward-char
            :vi-undo
            :vi-redo
            :vi-record-macro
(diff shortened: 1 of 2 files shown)


extensions/vi-mode/commands.lisp
```diff
@@ -639,6 +640,12 @@ Move the cursor to the first non-blank character of the line."
         (apply-visual-range #'swapcase-region)
         (swapcase-region start end))))
 
+(define-command vi-swapcase-and-forward-char () ()
+  (let*  ((start (copy-point (current-point)))
+          (end (character-offset (current-point) 1)))
+    (vi-swapcase start end (current-state)))
```
cxxxr, line 646: If you don't use copy-point in conjunction with delete-point, it will cause a memory leak, so it is better to use with-point.

If it were me, I would write it like this
```common-lisp
(with-point ((start (current-point))
             (end (current-point)))
  (character-offset end 1)
  ...)
```

cxxxr (commented):
Thanks for the PR.
I've made a comment, so please check it.

extensions/vi-mode/commands.lisp
sakurawald: Please help me modify it, i just learn that.

Issue #1590, opened 2024-11-08
Feature: sync the `mark region` between `vi-visual-mode` and the mark region selected by `mouse-clicking`.


sakurawald:
I am using `vi-mode` in `sdl2 version lem`.
If you use `mouse` click, to select a region.
And then using `keyboard` to use `vi commands`, the `selected region` still be slected, and is not count as the `vi-visual mode`.

To let the `vi-mode` works normally, you have to type `escape` or click the mouse again, to un-mark the selected region.
However, i want to know if there is a way to **un-mark the selected region selected by mouse-click automatically** while entering the vi-mode.

![image](https://github.com/user-attachments/assets/9ff57ff2-11a3-4acd-8d70-c3ba17fd6766)

sakurawald:
I didn't find a universal way to defien `escape key sequence` for `vi-mode`, so i do this:
```
;; use jk to escape vim. (use v key to escape in visual mode)
(define-key lem-vi-mode:*insert-keymap* "j k" 'lem-vi-mode/commands:vi-normal)
(define-key lem-vi-mode:*ex-keymap* "j k" 'lem-vi-mode/commands:vi-normal)

(define-key lem-vi-mode:*normal-keymap* "Space j k" 'escape)
```

sakurawald:
That's annoy, since if you use `mouse click` to select a region, then the `vi-mode` still in `NORMAL` instead of `VISUAL`.

sakurawald:
Maybe one solution is to, write 2 hooks:
1. on region marked, switch to vi-visual-mode
2. on region un-marked, swtich to vi-normal-mode

However, i don't know how to do this, and wheter there is existing hooks for this purpose.

PR #1633, opened 2024-11-22
feature: sync mark region for vi-mode.

sakurawald:
Related issue: https://github.com/lem-project/lem/issues/1590

To sync the mark-region between `vi-mode` and `mouse-clicking`:
1. Ensure the `vi-visual-char` mode is entered, when selecting a region using mouse.
2. Sync the `*start-point*` for `vi-visual-mode`.

This pr contains some cons:
1. The `visual-overlay` may not be cleared correctly: The `visual-overlay` will be cleared after you pressed `y` key in `vi-visaul mode`, if `(current-point)` > `*start-point*` (In right-side). It will not be cleared if `(current-point)` < `*start-point*` (In left-side).
2. Press `v` key again in `vi-visual-char` mode will not clear the `visual-overlay` selected by mouse.

Anyway, this pr:
1. Ensure the vi-mode will be `vi-visual-char` mode, if you use mouse to select a mark region.
2. Press `y` key in `vi-visual-char` mode can copy the `correct string` into `clip-board`. (Though it will not ensure the `visual-overlay` be synced.)

### DIFF
--- a/src/mouse.lisp
+++ b/src/mouse.lisp
@@ -215,6 +215,15 @@
     (funcall callback window point)
     t))
 
+(defun sync-mark-start-point-for-vi-mode (start-point)
+  (when (typep (lem-core:current-global-mode) 'lem-vi-mode:vi-mode)
+    ;; Ensure current main state is visual-char.
+    (unless (lem-vi-mode/visual:visual-char-p)
+      (lem-vi-mode/commands::vi-visual-char))
+    ;; Override the *start-point* in vi-visual mode.
+    (setf lem-vi-mode/visual::*start-point* start-point)
+    ))
+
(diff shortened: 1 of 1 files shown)


sakurawald:
This impl only work in `lem self-hack mod`, and failed to pass the compilation.

There are some tasks should be done:
1. Extract the `vi-mode` specific code from the `core code`, to avoid the access of `lem-vi-mode` package.
2. Maybe use a hook to listen the `mark region` changed event, and set the `*start-point*` for `vi-visual-char mode`.
3. The `visual-overlay in vi-mode` didn't disapper in some cases, like if the `(current-point) < *start-point*`, then the `visual-overlay` will not be cleared, and requries extra `esc key` to clear this overlay.

sakurawald:
Close this since it's done by another PR.

PR #1645, opened 2024-11-27
fix and refactor `kill-whole-line`

SequentialDesign:
this patch makes `kill-whole-line`, one, kill the entire line, including the new line at the end of a line if indeed it is there, and, two, allow for prefixes

|;; one one one
;; two two two
;; three three three
;; four four four
;; five five five

(| is point) if `5 kill-whole-line`, all lines are deleted.

;; one one one
;; two two two
;; three three |three
;; four four four
;; five five five

(| is point) if `0 kill-whole-line`, nothing happens.

;; one one one
;; two two two
;; three three |three
;; four four four
;; five five five

(| is point) if `-2 kill-whole-line`, ";; one one one" and ";; two two two" are deleted, and the point stays right before the last "three".  if you want to delete the previous "three"s, then you can use the `kill` command.

i'm curious if others agree with me

### DIFF
--- a/src/commands/edit.lisp
+++ b/src/commands/edit.lisp
@@ -212,13 +212,17 @@
            (let ((end (current-point)))
              (kill-region start end))))))))
 
-(define-command kill-whole-line () ()
-  "Kill the entire line and the remaining whitespace"
-   (with-point ((start (current-point))
-                (end (current-point)))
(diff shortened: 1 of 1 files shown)


cxxxr:
Thanks

Issue #1615, opened 2024-11-19
feature: set `read-only` for buffers that shows implementation source.


sakurawald:
When you execute `find-definition` command on a `built-in symbol`, you will go to the `buffer` with `implementation source`.

It's good to have a `protection` from modifying the source code accidentally.

PR #1655, opened 2024-11-30
feature: add impl-source-read-only editor-variable

sakurawald:
To close issue: https://github.com/lem-project/lem/issues/1615

**This option is enabled by default**, to disable it, eval the form in `init.lisp`
```
(setf (lem:variable-value 'lem-core::impl-source-read-only :global) nil)
```

The typical value of source dir for `sbcl implementation` are:
- `/usr/share/sbcl-source/src/code/` (installed via `yay -S sbcl`)
- `~/.roswell/src/sbcl-2.4.10/src/code/` (installed via `roswell`)

There may be a better function to decide if `(current-directory)` is part of `implementation source directory`.

image

### DIFF
--- a/src/highlight-line.lisp
+++ b/src/highlight-line.lisp
@@ -1,8 +1,14 @@
 (in-package :lem-core)
 
 (define-editor-variable highlight-line nil)
+(define-editor-variable highlight-line-color nil)
 
 (defun highlight-line-color ()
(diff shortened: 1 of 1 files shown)


sakurawald:
Another way to get the `source directory at runtime` is use the `SYS:SRC` logic path.

image

PR #1656, opened 2024-11-30
feature: add impl-source-read-only editor-variable

sakurawald:
To close issue: https://github.com/lem-project/lem/issues/1615

**This option is enabled by default**, to disable it, eval the form in `init.lisp`
```
(setf (lem:variable-value 'lem-core::impl-source-read-only :global) nil)
```

The typical value of source dir for `sbcl implementation` are:
- `/usr/share/sbcl-source/src/code/` (installed via `yay -S sbcl`)
- `~/.roswell/src/sbcl-2.4.10/src/code/` (installed via `roswell`)

There may be a better function to decide if `(current-directory)` is part of `implementation source directory`.

image

### DIFF
--- a/src/lem.lisp
+++ b/src/lem.lisp
@@ -303,3 +303,17 @@ See scripts/build-ncurses.lisp or scripts/build-sdl2.lisp"
                   (insert-string (buffer-point buffer) warning)
                   (insert-character (buffer-point buffer) #\newline))
                 (pop-to-buffer buffer)))))
+
+
+(defun impl-source-buffer-p (buffer)
+  (pathname-match-p (buffer-directory buffer)
+                    (second (assoc "SYS:SRC;**;*.*.*" (logical-pathname-translations "SYS") :test 'equal)))
+  )
+
+(add-hook *after-init-hook*
+          (lambda ()
+            (when (variable-value 'impl-source-read-only)
+              (add-hook *switch-to-buffer-hook* (lambda (buffer)
+                                                  (when (impl-source-buffer-p buffer)
+                                                    (setf (buffer-read-only-p buffer) t)))))
+            ))
--- a/src/window/window.lisp
+++ b/src/window/window.lisp
@@ -1,6 +1,7 @@
 (in-package :lem-core)
 
 (define-editor-variable line-wrap t)
+(define-editor-variable impl-source-read-only t)
 
 (defparameter *window-sufficient-width* 150)
 (defparameter *scroll-recenter-p* t)


sakurawald:
I think the form `(eval '(logical-pathname-translations "SYS"))` should be eval in the `target repl`, not the `host repl (the repl that running the lem editor)`.

sakurawald:
Still i don't know how to send the `eval form` to the `connected running repl`, or maybe there is some better solutions to this.

src/lem.lisp
```diff
@@ -303,3 +303,17 @@ See scripts/build-ncurses.lisp or scripts/build-sdl2.lisp"
                   (insert-string (buffer-point buffer) warning)
                   (insert-character (buffer-point buffer) #\newline))
                 (pop-to-buffer buffer)))))
+
+
+(defun impl-source-buffer-p (buffer)
+  (pathname-match-p (buffer-directory buffer)
+                    (second (assoc "SYS:SRC;**;*.*.*" (logical-pathname-translations "SYS") :test 'equal)))
+  )
```
cxxxr, line 311: > Still i don't know how to send the eval form to the connected running repl, or maybe there is some better solutions to this.

I have just changed micros and lem to allow this.

https://github.com/lem-project/lem/commit/86d9f36ae1c93d238ec67f2f1671ffa8984d4674
https://github.com/lem-project/micros/commit/25177e04a946365397a0848a6545399860d310d0


The source pattern is now retained for each connection, so that it can be written as follows.
```common-lisp
(dolist (pattern (connection-system-file-patterns (current-connection)))
  (when (pathname-match-p directory pattern)
    (return t)))
```

src/lem.lisp
```diff
@@ -303,3 +303,17 @@ See scripts/build-ncurses.lisp or scripts/build-sdl2.lisp"
...
+
+(add-hook *after-init-hook*
+          (lambda ()
+            (when (variable-value 'impl-source-read-only)
+              (add-hook *switch-to-buffer-hook* (lambda (buffer)
+                                                  (when (impl-source-buffer-p buffer)
+                                                    (setf (buffer-read-only-p buffer) t)))))
+            ))
```
cxxxr, line 319: I think it could be written a little simpler here.

for example
```common-lisp
(defun on-switch-to-buffer (buffer)
  (when (and (variable-value 'read-only-sources :default buffer)
             (read-only-source-p (buffer-directory buffer)))
    (setf (buffer-read-only-p buffer) t)))

(add-hook *switch-to-buffer-hook* 'on-switch-to-buffer)
```

src/window/window.lisp
```diff
@@ -1,6 +1,7 @@
 (in-package :lem-core)
 
 (define-editor-variable line-wrap t)
+(define-editor-variable impl-source-read-only t)
```
cxxxr, line 4: This definition is likely to be better located close to the point of use.

cxxxr (commented):
Thank you for the PR.
This suggestion for improvement is wonderful.
I am also grateful that you actually made the suggestion into a PR.

Also, I have some comments and suggestions regarding the changes.
Thank you.

## suggestion 1

What about separating the files?

I think it can be separated into a single function: make this particular directory read-only.

Then it seems to me that it would be better to separate each file/package.

```common-lisp
;; src/ext/read-only-sources.lisp
(uiop:define-package :lem/read-only-sources
  (:use :cl :lem)
  ...)
(in-package :lem/read-only-sources)

...

(add-hook *switch-to-buffer-hook* 'on-switch-to-buffer)
```

## suggestion 2

How about separating the definition of the rules for making it read-only from the hooks?

For example, the following macro defines a rule.
```common-lisp
;; lisp-mode.lisp
(define-read-only-source sbcl-source (directory)
  (dolist (pattern (connection-system-file-patterns (current-connection)))
    (when (pathname-match-p directory pattern)
      (return t))))

```

The definition of this macro is placed in the same place as the hook.
```common-lisp
;; src/ext/read-only-sources.lisp

(defvar *patterns* (make-hash-table :test 'equal))

(defun register-pattern (name function)
  (setf (gethash name *patterns*) function))

(defmacro define-read-only-source (name (directory) &body body)
  `(register-pattern ',name (lambda (,directory) ,@body)))

(defun read-only-source-p (directory)
  (maphash (lambda (name pattern)
             (declare (ignore name))
             (when (funcall pattern directory)
               (return-from read-only-source-p t)))
           *patterns*))

(defun on-switch-to-buffer (buffer)
  (when (and (variable-value 'read-only-sources :default buffer)
             (read-only-source-p (buffer-directory buffer)))
    (setf (buffer-read-only-p buffer) t)))
```

sakurawald:
The *Suggestion 1** would be good, to extract the feature into an extension, reducing the complexity of the core source.

The *Suggestion 2** makes the read-only a facility to support other purpose, looks nice.

You can close this issue, if you have implement the part of the source.

src/lem.lisp
sakurawald: Yeah, easier to read, and eliminates the un-necessary lambdas.

src/window/window.lisp
sakurawald: It's true. I didn't really understand the initialization process of lem, so i put the editor-variable in here.

cxxxr:
> You can close this issue, if you have implement the source in your side.

OK.
I have created the following PR

PR #1659, opened 2024-11-30
java-mode: Use c-mode indentation function to indent Java files

bolsen:
I found Lem today. :) I saw that I can't indent Java files. It looked like it was missing an indent function, so I am using the one from c-mode.


cxxxr:
Thank you.

PR #1665, opened 2024-12-02
change `insert-()` to `insert-()-or-wrap`

SequentialDesign:
change `insert-()` so that it wraps parens over a region if the mark is active and inserts () when it's not, in both cases, leaving the point right after the open paren.  this patch also updates the name of the function so that functionality is not hidden from the user


cxxxr:
Thank you.

PR #1666, opened 2024-12-02
change `move-over-)` to `move-over-)-or-wrap`

SequentialDesign:
change `move-over-)` to `move-over-)-or-wrap` so that if the mark is active, the region will be wrapped with parens and the point will be placed after the close paren, helpful if you don't need to add anything in the parens.  could also be helpful if you use pareto mode.  and the name is changed so that functionality is not hidden from the user

an odd command, but one that seems like a logical counterpart to `insert-()-or-wrap`

### DIFF
--- a/src/ext/language-mode.lisp
+++ b/src/ext/language-mode.lisp
@@ -93,7 +93,7 @@
 (define-key *language-mode-keymap* "M-," 'pop-definition-stack)
 (define-key *language-mode-keymap* "C-M-i" 'complete-symbol)
 (define-key *global-keymap* "M-(" 'insert-\(\)-or-wrap)
-(define-key *global-keymap* "M-)" 'move-over-\))
+(define-key *global-keymap* "M-)" 'move-over-\)-or-wrap)
 
 (defun beginning-of-defun-1 (n)
   (alexandria:when-let ((fn (variable-value 'beginning-of-defun-function :buffer)))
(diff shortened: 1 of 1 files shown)


cxxxr:
https://github.com/lem-project/lem/pull/1665/files
How is this different from the PR here?
Please fix it as there seems to be a conflict going on.

SequentialDesign:
change move-over-) to move-over-)-or-wrap so that if the mark is active, the region will be wrapped with parens and the point will be placed after the close paren, helpful if you don't need to add anything in the parens. could also be helpful if you use pareto mode. and the name is changed so that functionality is not hidden from the user
an odd command, but one that seems like a logical counterpart to insert-()-or-wrap

the difference between this patch and `insert-()-or-wrap` is that this one puts the point at the end, after the close paren.

(when (point> end |current-point)

(| is the point.)  if we put the region around "current-point" and then execute this command, we will get "(when (point> end (current-point))" and then point will be to the right of the close paren.

(when (point> end (current-point)|)

(| is the point) and then we can execute this command (the region will not be active), and we will skip over the close paren and go to the next line.

(when (point> end (current-point))
  |

(| is the point.)  seems that it could be useful.

cxxxr:
Ah, indeed.
I misread it and thought it was the same command change.
Thank you for fixing the conflicts, will merge.

PR #1670, opened 2024-12-03
feature: add `:new` ex-command for `vi-mode`

sakurawald:
The `:new` command in `vi-ex mode` is used to create a new empty buffer.

### DIFF
--- a/extensions/vi-mode/ex-command.lisp
+++ b/extensions/vi-mode/ex-command.lisp
@@ -259,3 +259,7 @@
 (define-ex-command "^pwd?$" (range argument)
   (declare (ignore range argument))
   (lem:current-directory))
+
+(define-ex-command "^new?$" (range argument)
+  (declare (ignore range argument))
+  (lem/frame-multiplexer:frame-multiplexer-create-with-new-buffer-list))


extensions/vi-mode/ex-command.lisp
```diff
@@ -259,3 +259,7 @@
 (define-ex-command "^pwd?$" (range argument)
   (declare (ignore range argument))
   (lem:current-directory))
+
+(define-ex-command "^new?$" (range argument)
+  (declare (ignore range argument))
+  (lem/frame-multiplexer:frame-multiplexer-create-with-new-buffer-list))
```
cxxxr, line 265: Isn't it different from the original behavior?
It seems that vim splits the screen.

sakurawald: > Isn't it different from the original behavior? It seems that vim splits the window.

Sorry, i mis-understand this.
I would add this feature back in the future. Now i am working to add the missing vim command: the `vi-jump-previous` to jump back and forth, useful to quickly jump to a location, and back.
The default binding is `''`. I found the similar commands like `C-o` and `C-o`, but the behaviour is a bit different.

PR #1671, opened 2024-12-03
fix: the `vi-paste-before` command in `line-mode` at `the last line` should open a newline before current line.

sakurawald:
This commit fix the `edge case` when vi-paste a line before at the last line.

Taken this example buffer
```
first
second
third1 th[i]rd2 third 3
```

If current-line is **at the last line**, then the `vi-paste-before` command in `line-mode` should `open a new line` above `current-line`.

---
The keypoint is that: **The last line** doesn't contains the **line-break character**, so it's an edge-case.

### DIFF
--- a/extensions/vi-mode/commands.lisp
+++ b/extensions/vi-mode/commands.lisp
@@ -568,7 +568,11 @@ Move the cursor to the first non-blank character of the line."
          (insert-character (current-point) #\Newline)))
       (t
        (when (eq type :line)
-         (line-start (current-point)))))
+         (if (last-line-p (current-point))
+             (progn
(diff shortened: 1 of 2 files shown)


sakurawald:
This also can't pass the `continous paste before case`.
If in `line-mode` paste, the `paste string` contains a `line-break character`, this `line-break character` introduces the difference between `the last line` and `not the last line`.

sakurawald:
The `vi-yank-line` command, behaves differently on `the last line` and `not the last line`. 
In `not the last line`, the `line-break character` is copied.
In `the last line`, the `line-break character` is not copied.

PR #1675, opened 2024-12-04
[DRAFT] add: the `:jumps` ex-command for vi-mode.

sakurawald:
The `:jumps` command for `vi-ex-mode`, which shows **current jumplist stack pointer** and the **jumplist**, useful to help people understand how the state of **jumplist**, and how it works.

I have checked that the current implementation of `jumplist` structure and the `print-jumplist` funciton does print **the same human readable string** as `vim`.

image
image

### DIFF
--- a/extensions/vi-mode/commands.lisp
+++ b/extensions/vi-mode/commands.lisp
@@ -123,6 +123,7 @@
            :vi-append-line
            :vi-open-below
            :vi-open-above
+           :vi-jumps
            :vi-jump-back
            :vi-jump-next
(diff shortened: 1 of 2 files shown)


sakurawald:
Maybe need another window to display the info, now the message window.

sakurawald:
It seems that the `(lem:message)` and `(lem:message-buffer)` will failed to close, if the `mouse focus` is on their window. Another pr will be need to fix this.

sakurawald:
I close this pr, and will include it in another pr.

PR #1682, opened 2024-12-05
feature: make `vi-paste-{before/after}` accept the universal args.

sakurawald:
In `vim`, it's possible to use `yy3p` to: yank current line, and paste it 3 times.


sakurawald:
The `github diff` seems strange. The changes are: wrap the `vi-paste-{before/after}` with `(dotimes (i n) ...)`, and modify the `function signature` with `(&optional (n 1))` and `(:universal)`

PR #1684, opened 2024-12-05
add two editor variables to line-numbers, line-number-format and current-line-value

SequentialDesign:
with these editor variables, the user can customize line-format and current-line value. for example :

```
(setf (variable-value 'line-number-format :global) "~2D ")
(setf (variable-value 'relative-line-numbers-current-line-value :global) 0)
 ```

this will make the line numbers narrow and make the current line show up as '0' in the line-number column

### DIFF
--- a/src/ext/line-numbers.lisp
+++ b/src/ext/line-numbers.lisp
@@ -1,6 +1,8 @@
 (defpackage :lem/line-numbers
   (:use :cl :lem)
   (:export :*relative-line*
+           :line-number-format
+           :relative-line-numbers-current-line-value
            :line-numbers-attribute
            :active-line-number-attribute
            :line-numbers
@@ -15,7 +17,15 @@
 (defvar *previous-relative-line* nil)
 
 (defvar *initialized* nil)
-(defvar *line-number-format* nil)
+
+(define-editor-variable line-number-format "~6D "
+  "Set to desired format, for example, \"~2D \" for a
+two-character line-number column.")
+
+(define-editor-variable relative-line-numbers-current-line-value 
+  '(line-number-at-point (current-point))
+  "Set to desired current-line value when relative line numbers
+are active, for example, 0 or \"->\".")
 
 (define-attribute line-numbers-attribute
     (t :foreground :base07 :background :base01))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/ext/line-numbers.lisp [hunk 1]
```diff
@@ -15,7 +17,15 @@
...
 (defvar *initialized* nil)
-(defvar *line-number-format* nil)
+
+(define-editor-variable line-number-format "~6D "
+  "Set to desired format, for example, \"~2D \" for a
+two-character line-number column.")
+
+(define-editor-variable relative-line-numbers-current-line-value 
+  '(line-number-at-point (current-point))
+  "Set to desired current-line value when relative line numbers
```
cxxxr, line 27: I think it would be better to make this a function rather than eval.

cxxxr: Before that, it seems like the API that specifies this behavior from the outside is not a good idea.

src/ext/line-numbers.lisp
```diff
@@ -15,7 +17,15 @@
 (defvar *previous-relative-line* nil)
 
 (defvar *initialized* nil)
-(defvar *line-number-format* nil)
+
+(define-editor-variable line-number-format "~6D "
```
cxxxr, line 21: I think this is good.

SequentialDesign:
sorry for the poor style, purely bad habits.  and thank you for the patience

i have changed it for the variable to be assigned inside the let with `funcall` and modified the name of the variable so that it's clear that it must be a function

src/ext/line-numbers.lisp
```diff
@@ -15,13 +17,21 @@
...
+  "Set to desired format, for example, \"~2D \" for a
+two-character line-number column.")
+
+(define-editor-variable current-line-display-function
+  (lambda () (line-number-at-point (current-point)))
```
cxxxr, line 26: compute-line holds the point variable.
Wouldn't it be better to take that as an argument?
It seems to me that using what is passed as an argument is a better style than thinking about what position the current-point is actually in.

```
#'line-number-at-point
```

SequentialDesign: okay, i think i see what you mean.  the thing is i want to be able to set the value to `0` or `->`, etc.  so i'm guessing you mean make a variable like `custom-current-line-indicator` and set it to nil, and if it is defined, then `compute-line` will call that.  otherwise, show the current line number.

that does make sense.  am i understanding that correctly ?

SequentialDesign:
alright, i think it's much better now.  `custom-current-line` is set to nil, and that will show absolute line.  if `custom-current-line` is set, then it will show its value

SequentialDesign:
one more change.  i put everything in the let to make it cleaner and proper

src/ext/line-numbers.lisp
cxxxr: Okay, I understand your intention.

Issue #1686, opened 2024-12-05
improve: can't close floating-window if mouse focus gained. (`"Can not delete this window"`)


sakurawald:
Steps tp reproduce:
1. Press `M-a` in a lisp symbol, to open a `floating window`.
2. Use mouse to click this window.
3. Press `Esc` to delete this window.

PR #1693, opened 2024-12-09
Ban `mouse click event` in `popup-window` maded by `popup-message` function.

sakurawald:
To close issue: https://github.com/lem-project/lem/issues/1686

This commit does:
1. Add a new parameter `clickable` to `window`. (When `clickable = nil`, the `mouse clicking` event will be ignore for this window. However, thw `mouse hover event` and `mouse scrolling event` can still be received for that window)
2. Made the `popup-window` maded by `popup-message` now `un-clickable`.

### DIFF
--- a/src/ext/popup-message.lisp
+++ b/src/ext/popup-message.lisp
@@ -33,6 +33,9 @@
                      :buffer buffer
                      :width width
                      :height height
+                     ;; NOTE: Simply ban the mouse-click event for a popup-window, to avoid the popup-window gain the mouse focus, being the only window as current-window in window-tree, causing lem unable to delete it.
+                     ;; However, since the popup-message is a window used to display string, you can still use mouse-scroll to scroll the text.
+                     :clickable nil
                      :style style)))
         (buffer-start (window-view-point window))
         (window-see window)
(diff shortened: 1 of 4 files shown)


jfaz1 (approved):
LGTM, maybe modify the comment to be a bit clearer:

```lisp
;; NOTE: Block mouse clicks on popup windows to prevent them from gaining focus.
;; If a popup window gains focus, it becomes the only active window in the window-tree,
;; which prevents Lem from being able to close it.
;; Mouse scrolling remains enabled since popup windows are used for displaying text content.
```

sakurawald:
> LGTM, maybe modify the comment to be a bit clearer:
> 
> ```commonlisp
> ;; NOTE: Block mouse clicks on popup windows to prevent them from gaining focus.
> ;; If a popup window gains focus, it becomes the only active window in the window-tree,
> ;; which prevents Lem from being able to close it.
> ;; Mouse scrolling remains enabled since popup windows are used for displaying text content.
> ```

Modified.

jfaz1 (approved):

cxxxr:
Thank you

PR #1694, opened 2024-12-09
feature: ignore-case for git-grep command by default

sakurawald:
For a `wide-scope grep` like `project-grep` or `directory-grep`, it's **more useful** to ignore-case in most of cases.

The search scope is large, it may be annoy to **miss some search result item** just because the case is sensitive.

### DIFF
--- a/src/ext/grep.lisp
+++ b/src/ext/grep.lisp
@@ -72,7 +72,7 @@
   (lem/peek-source:show-matched-line))
 
 (defvar *grep-command* "git grep")
-(defvar *grep-args* "-nHI")
+(defvar *grep-args* "-niHI")
 (defvar *last-query* (str:concat *grep-command* " " *grep-args* " "))
 (defvar *last-directory* nil)
 


sakurawald:
> I think that case-sensitive is the default for general browsers and editors.

Not all, the `modern IDEs` like `jetbrains` ignore the case by default.
image


The `firefox browser does ignore case by default.
image

The reason that the `chrome browser` only use the `exactly pure-string match, and case-sensitive`, is that the `chrome` is designed to be used by `non-tech user`, so it's good to have easy and simple search function.

sakurawald:
Sorry, i just learn that the `chrome` does `ignore-case` by default.

image

sakurawald:
> I think that case-sensitive is the default for general browsers and editors.

Okay, i document this in the user manual.
https://github.com/lem-project/lem-project.github.io/pull/48

cxxxr:
For example,
```
class User {
```
If you want to search for the User class,

```
const user = ...
```
it is not good to match the variable user in the line

sakurawald:
> For example,
> 
> ```
> class User {
> ```
> 
> If you want to search for the User class,
> 
> ```
> const user = ...
> ```
> 
> it is not good to match the variable user in the line

This case didn't make sense.

My consideration is, the `git grep` or `directory grep` is used to filter out `interesting strings` in a directory **where you don't know what files it contains and what is inside the files.** (Image you are just clone a new git repository, and and to `git grep` some interesting strings, so you don't care the `case` and don't want to miss some `interesting strings`. This is a wide-scope search, we just try and find some interesting strings.)

If you know there is a `class User`, you are possible know which file it is in (For example, the definition may written inside `User.js`), you can just use `project-open-file` to open that file.
You don't need to do a 2-step operation: 
1. run the `git grep` and select the only 1 interesting result item.
2. go to the file that contains the definition of `class User`.

You can just do:
1. open the `User.js` file using `project-open-file` command.

Or let's say you really want to find the definition of `class User` using a `project grep`, you can just grep the `user` and ignore the case, and in the `grep result window`, simply type `/` key to start a vi-isearch, this time we search `class User`, and you will find all you want.

Also, for those languages whose identifier is case-sensitive

sakurawald:
In this screen-shot, you can see:
1. Start a `project-grep` to search `defclass`. (ignore-case)
2. Start a `isearch` in the `grep result`. (ignore-case)

image

---

In this screen-shot, you can see:
1. Start a `project-grep` to search `defclass point` (ignore-case)
image

sakurawald:
> The variable 'user' is used in many places, so there is too much noise if I want to find the User class. Classes and variables can sometimes have the same name, but sometimes you only want one of them to be matched when searching, so I sometimes use it for that purpose.

I aware that. It may be useful to take case into consideration in some language that cares the case, like `JavaScrip`, `Java` and `CPP`. I myself use `git grep` as a broad-phase filter (And start a second reg-isearch in the grep result window, as a precies filter.), to find interesting strings in a new project, which i am not very familar with.

This is my work-flow:
1. For definition of symbols, i use `(define-key lem-vi-mode:*normal-keymap* "g d" 'lem/language-mode::find-definitions)` to goto the definition buffer of that symbol. (The `find-definitions` works well in `lisp-mode`, and also, i can use `find-references` later.). 
image

2. For text serach (occurrence of strings, case-insensitive.), i use `project-grep`.
The grep is really a powerful text processing tool, if you want to find out all occurrence of text/string `*magic-variable*` or `magic-class`, you can ignore the case, find all its occurrenec, and see where the text occurrence, and finally, try to figure out what it does.

PR #1695, opened 2024-12-09
feature: recenter line if jump target point is not inside window. (vi-mode)

sakurawald:
In `vim`, the `jump-back` and `jump-next` will `recenter jump target line` **if the line is not inside the window before jump**.

### DIFF
--- a/extensions/vi-mode/commands.lisp
+++ b/extensions/vi-mode/commands.lisp
@@ -1019,6 +1019,11 @@ on the same line or at eol if there are none."
                                     (line-start p))))))
     (move-to-column (current-point) column t)))
 
+(defun recenter-line-if-point-not-inside-window (point &optional (window (current-window)))
+  (when point
+    (unless (lem-vi-mode/window::point-inside-window-p point window)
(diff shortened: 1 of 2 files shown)


sakurawald:
The code is formatted by the `default formatter` provided by `lem`.
I didn't know why the `identation` is different from the original file. It's annoy, each time i have to reset the identation differs.

sakurawald:
Sorry, i didn't notice that 1 line code is deleted accidently.

sakurawald:
I remove the `target-point` binding in `let form`, and document the `return value` of `(jump-back)` and `(jump-next)`.
It looks clearer.

sakurawald:
I make this pr a draft, and considering make it a feature in lem.

A new `line positioning strategy` to `re-position where the line should the cursor in` after jumping the point. Or to say, a function `line-view-position-function` to be called after we `move-point` to a new `point`.

The strategy can be: `none`, `recenter-current-line-if-new-point-is-not-inside-original-window-view`, `scroll-line-to-top-of-window`, `scroll-line-to-bottom-of-window`.

PR #1697, opened 2024-12-09
Add a mechanism to perform conversions such as C-m/Return and C-i/Tab in a unified manner

cxxxr:
Until now, the conversion of keys was done for each frontend, but this has been changed so that it is done on the lem side.


sakurawald:
Tested, it works well.
It's a nice thing to unify the key-conversion in the core source. Making it easier to share the key-bindings between frontends. Not to mention a new way to emit the `Escape` key using `C-[`.

PR #1701, opened 2024-12-10
fix: the self-insert command in vi-mode should not insert un-printable char while flushing the pending-keys

sakurawald:
In pr: https://github.com/lem-project/lem/pull/1691, we flush the `pending-keys` for `self-insert` command in `vi-insert-mode`, so that we can define keymap like `(define-key lem-vi-mode:*insert-keymap* "j k" 'lem-vi-mode/commands:vi-normal)`.

The `effect` of this `define-key` is:
1. Press `j k` in `vi-insert-mode` will the `lem-vi-mode/commands:vi-normal` command.
2. Press `j <second-key>` in `vi-insert-mode` will self-insert `j` and self-insert `<second-key>`.

However, in the `vim` editor, if the `<second-key>` is NOT print-able key, then it will `quit the key pending state`, and will not self-insert the `<second-key>`.

**A case is: press `j` and `Backspace`.**
This pr fix this edge-case.

---
NOTE: 
This pr has some minor difference from the `vim`. The `named-key` is not identical to `print-able-key`, taken `Tab` for example.
I didn't add the function `printable-key-sym-p` or `un-printable-key-sym-p`, because it is hard to decide whether a key is print-able or not.

The difference is minor, and sufficient in daily use.

### DIFF
--- a/extensions/vi-mode/vi-mode.lisp
+++ b/extensions/vi-mode/vi-mode.lisp
@@ -77,7 +77,10 @@
     (when (and
            (typep command 'self-insert)
            pending-keys)
-      (dolist (key pending-keys) (self-insert 1 (key-to-char key))))
+      (loop :for key :in pending-keys
+            :until (named-key-sym-p (key-sym key))
+            :do 
+               (self-insert 1 (key-to-char key))))
     
     (when *enable-repeat-recording*
       (unless (or (and (typep command 'vi-command)
--- a/extensions/vi-mode/vi-mode.lisp
+++ b/extensions/vi-mode/vi-mode.lisp
@@ -76,20 +76,23 @@
     ;;   a. If `<second-key>` is print-able-key, taken `e` key for example, then will self-insert `j` and self-insert `e`.
     ;;   b. If `<second-key>` is un-print-able-key, taken `Backspace` key for example, then will cancel the self-insert `j` and remain in vi-insert-mode.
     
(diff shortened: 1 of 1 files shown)


sakurawald:
The `vi-pre-command-hook` calles the `pre-command-hook` function.
However, i would like to write this feature in `command-hook` and using `:around` method-combination to do the effect. (The `:around` gives the possibility to `cancel` the `self-insert j`, if the `<second-key>` is `un-printable-key`.)

sakurawald:
Another joint to handle the `pending-keys flushing for self-insert command` is to use `(defmethod execute :around (mode (c "" ommand self-insert) argument)`, so that we can `cancel` the executing of `self-insert` command, if the `<second-key>` is `un-printable-key`.
However, this `method` only support to specify the `major mode`. 
---
I think it's ready to be merged.

sakurawald:
Currently, for `self-insert` command **in vi-mode**, we `flush the keys in (cdr this-command-keys)` into the buffer. (We do this via the `post-command-hook` in `vi-insert-mode`)

Since the `emacs-mode` seldom use pure `alpha letters` to define-key. (Emacs define-keys using `alpha-letter` with `modifier keys`), and `vi-mode` does define some keys using pure alpha-letter like `define jk in vi-insert-mode).
I impl the `flushing function` in `post-command-hook of vi-mode`, not the global `executing :around` for `self-insert` command.

PR #1703, opened 2024-12-10
add `after-load-theme-hook`

SequentialDesign:
add `after-load-theme-hook` so that users may add functions that should run after a theme is loaded.  this could be useful in some cases.  in my specific case, it's to to maintain the color of the point since it is always very hard to see when the point is on an overlay like in search.

### DIFF
--- a/src/color-theme.lisp
+++ b/src/color-theme.lisp
@@ -1,5 +1,9 @@
 (in-package :lem-core)
 
+(defvar *after-load-theme-hook* '()
+  "For functions that should run after a theme is loaded,
+for example, to maintain an attribute like CURSOR.")
+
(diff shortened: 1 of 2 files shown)


sakurawald:
This is my config
```
(setf lem-vi-mode/core::*default-cursor-color* "#ffffff")
(setf (lem:variable-value 'lem-core::highlight-line-color :global) "#000066")
```

May be useful to color cursor.

SequentialDesign:
i only ever tried `(set-attribute 'cursor :background "magenta")` in my `init.lisp`, but because `init` runs `after-init-hook`, running `initialize-theme' which in turn runs `load-theme`, i didn't see another proper way to get my cursor to be what i wanted on startup

Issue #1612, opened 2024-11-19
The level in paired `open-parenthesis` and `closed-parenthesis` is not identical.


sakurawald:
If you `current cursor` is on `close parenthesis`, then the result of `M-x paredit-slurp <ret>` is as expected.
 
![image](https://github.com/user-attachments/assets/29e3ebbd-10d3-4a66-9b74-7026149ddaf1)

![image](https://github.com/user-attachments/assets/d342aa5c-8b6b-4759-876e-43ea1ac0428f)

However, if your cursor is on `open parenthesis`, then the result is different from the previous result.

What i expected is that: the cursor `on open parenthesis`, `on closed parenthesis` and `between open parenthesis and closed parenthesis` should mean **the same list**.

However, the actual result is that, if the cursor is `on close parenthesis` and `between open parenthesis and closed parenthesis`, then it means the same list. But if the cursor is on `open parenthesis`, then it means **the different list**.

sakurawald:
The case, if the `cursor` is `on open parenthesis`.

![image](https://github.com/user-attachments/assets/7c25e500-1e91-46ee-9c17-642b9400ee57)
![image](https://github.com/user-attachments/assets/a900ac6c-51e0-4a19-8530-f019ded5e423)

sakurawald:
A similar issue is here, also related to the `identity of open parenthesis and closed parenthesis`.

I guess this is something about the parenthesis matching algorithm, it seems the open parenthesis is simply ignored, and use the closed parenthesis to match it. Not very sure about this.

https://github.com/lem-project/lem/issues/1591

sakurawald:
I close this issue, because it seems this behaviour is the same in `smart-paren` extension in emacs.
It is treated as a feature, not a bug. (Not very understand).

PR #1709, opened 2024-12-11
fix: the paredit-splice should not backward 1 form if current point already at open-char

sakurawald:
This pr fixes the `matched level` of `open-char` and `closed-char`.
To fix cases like: https://github.com/lem-project/lem/issues/1612

Before this pr: the `paredit-splice` treat `string between open-char and closed-char` and `closed-char` as the `same form`, but treat the `open-char` as the `(form-offset (current-point) -1)`.

After this pr: the `paredit-splice` treat `open-char`, `string between open-char and closed-char` and `closed-char` as the `same form`.

---
For this test-case:
image

Before this pr:
image

After this pr:
image

### DIFF
--- a/extensions/paredit-mode/paredit-mode.lisp
+++ b/extensions/paredit-mode/paredit-mode.lisp
@@ -2,6 +2,7 @@
 link : http://www.daregada.sakuraweb.com/paredit_tutorial_ja.html
 |#
 
+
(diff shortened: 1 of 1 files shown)


sakurawald:
Closed as this is treated as a feature.

PR #1713, opened 2024-12-24
vi-mode: Adding additional C-w functionality (Window management)

LeeLaffan:
# Preface
Hi! I just want to start off by saying how much I am enjoying Lem. It has been great to work with the project.
I am a vim user, and wanted to add some of the functionality around window management to the Lem vi-mode.

This is my first public pull request, so please let me know if anything is not protocol. I am also new to common lisp and lisp in general, so I am very happy to take on feedback about the code. Same with the Lem specific stuff. I am not an expert yet. 

I would also be happy to break this up into smaller pull requests if that would help, let me know. Although there are sections like Keybinds and Ex-Commands, they use the same vi-* functions in a lot of cases.

Thank you for your time.
Lee

# Feature Breakdown
I tested what I could with a clean vim build to verify this functionality matched.
To each key where the docs specified(https://vimhelp.org/windows.txt.html), I added the alternative C-w h & C-w C-h. Some are not in vim like C-w C-c.
    
### C-w hjkl
Navigating splits.
To allow number prefix, eg. `2 C-w l`
#### New Commands
```lisp 2
   (vi-window-move-left)
   (vi-window-move-down)
   (vi-window-move-up)
   (vi-window-move-right)
```
    
### C-w s/v
Splits the window and moves to the split.
This now allows a number prefix, eg. `3 C-w v`
#### New Commands
```lisp 
  (vi-window-split-horizontally)
  (vi-window-split-vertically)
```
    
### C-w n
Split the window and creates a new window as per the vim docs here: https://vimhelp.org/windows.txt.html#opening-window
#### New Commands
```lisp 
  (vi-window-split-vertically-new)
```
    
### C-w c
Closes the current split. If there is one window, don't close it as in vi/vim.
#### New Commands
```lisp 
  (vi-close)
```
    
### C-w p
Uses Lem function to go to previous window.
```lisp 
  (define-key *motion-keymap* "C-w p" 'previous-window)
```

### C-w o
Uses Lem function to delete other windows.
```lisp 
  (define-key *motion-keymap* "C-w o" 'delete-other-windows)
```
    
## Ex-Commands
These link up with the above keybinds for the most part.

### :new
Creates a new buffer in a split below. Uses the `C-w n` command: (vi-window-split-vertically-new)
    
### :vne(w)
Creates a new buffer in a split to the right. Uses something similar to the above: (vi-window-split-horizontally-new)
There isn't a vanilla vim keybinding to match this functionality.
    
### :enew(w)
Opens the specified file name or creates a blank buffer.
There isn't a vanilla vim keybinding to match this functionality.
    
### :clo(se)
Closes the buffer, unless it's the last one. Uses the `C-w c` command: (vi-close)
    
### :on(ly)
Uses Lem function to delete other windows. Same as `C-w o`: (lem:delete-other-windows)

## Bug with :wa
There was a small bug in the matching for :wa(ll) and :wa(ll)!
They did not treat the (ll) as optional.
```code
  ^wa(?:ll)$ 

  should be:

  ^wa(?:ll)?$
```


cxxxr:
Thanks a lot!

LeeLaffan:
Seems I used a variable that is marked as ignored. This is cleaned up in this PR:

https://github.com/lem-project/lem/pull/1714

Issue #278, opened 2018-08-14
[vi-mode] Not implemented / unexpected features (since Aug 14th)


fukamachi:
- [x] `daw` `caw`: Delete a word at the cursor
- [x] `C-o` (and `C-i`) to move backward (forward) to the previous cursor position. (Vim's jumplist)
- [x] `q`: Keyboard macro
- [x] `I` and `A` in rectangle visual mode.
- [x] The point to move with `b` `B` is not compatible with `e` `W`

ref #205

- [x] `dG`
- [x] `dgg`
- [x] `yG` and `ygg` (probably `vi-delete` and `vi-yank` should be share those code)
- [x] Rectangle paste (#284)
- [x] `(` & `"` will be inserted even in COMMAND mode when using with paredit-mode
- [x] `Esc` to quit from modeline
- [x] Search repeat at the last result (#285)
- [x] Make `:e` allow to start completion with extra following spaces
- [x] Make `:e.` work to start completion
- [x] Completion for `:vs` `:sp`
- [x] `x` & `s` to delete region in VISUAL mode
- [x] `r` in VISUAL mode
- [x] `d%` deletes unexpected region for a multiple line form
- [x] Pasting in visual mode #286
- [x] Don't treat `Esc` + Something as `M-` command #287
- [x] `o` in VISUAL mode for moving to the opposite side of the region
- [x] `%s///` doesn't substitute all lines (`/g` does it, but it doesn't mean the same)
- [x] `/c` flag for substitution (`:s`)
- [ ] `/n` flag for substitution (`:s`)
- [x] Auto set regions with `:` in VISUAL mode (like `:'<,'>`)
- [x] `3$`
- [x] `2Y`
- [x] Cut the last word with `C-w` in INSERT mode
- [x] scrolloff option
- [x] #300 
- [x] `.` #302 
- [x] `:update` #299
- [x] `g^` `g0` `g$`
- [x] Make incremental search string global
- [x] Cursor position of `n` #314
- [x] Make selectable line end in visual mode (#315)
- [x] `:buffers` #318 
- [x] `:buffer` #319 
- [x] `Return` inserts a newline even in visual mode
- [x] https://github.com/lem-project/lem/issues/1376
- [x] https://github.com/lem-project/lem/issues/1191
- [x] https://github.com/lem-project/lem/issues/1188

kenoss:
#287 is done.

theangelperalta:
[...]

PR #1714, opened 2024-12-24
vi-mode: Add g^, g0, g$

LeeLaffan:
These use Lem's move-to-(beginning/end)-of-line. The equivalent for the regular non-g ^ or $ would be move-to-(beginning/end)-of-logical-line.

These are requested features in:
  [vi-mode] Not implemented / unexpected features (since Aug 14th) #278 

There is also a variable marked as ignored when it is used. From this PR: https://github.com/lem-project/lem/pull/1713

Even though it is a small commit, it's good to provide the best experience for users out of the box, some people might use these keybinds a lot.


cxxxr:
Thank you

PR #1715, opened 2024-12-24
Add micros keybinds to lisp-mode

jfaz1:
Starting/restarting slime (micros) is a common-enough task I think it makes sense for it to have default keybinds.

I used the `C-c m` prefix, so e.g. `C-c m s` to start micros, `C-c m r` to restart, etc. Originally wanted `C-c C-m` but `C-m` is reserved for return.

Side-note, we should rename the `slime-*` commands at some point to `micros-*`. I didn't want to include that here since that's a breaking change.

Also, merry christmas :christmas_tree: :santa:

### DIFF
--- a/extensions/lisp-mode/lisp-mode.lisp
+++ b/extensions/lisp-mode/lisp-mode.lisp
@@ -92,6 +92,10 @@
 (define-key *lisp-mode-keymap* "Return" 'newline-and-indent)
 (define-key *lisp-mode-keymap* "C-c C-j" 'lisp-eval-expression-in-repl)
 (define-key *lisp-mode-keymap* "C-c ~" 'lisp-listen-in-current-package)
+(define-key *lisp-mode-keymap* "C-c m s" 'slime)
+(define-key *lisp-mode-keymap* "C-c m r" 'slime-restart)
+(define-key *lisp-mode-keymap* "C-c m q" 'slime-quit)
+(define-key *lisp-mode-keymap* "C-c m c" 'slime-connect)
 
 (defmethod convert-modeline-element ((element (eql 'lisp-mode)) window)
   (format nil "  ~A~A" (buffer-package (window-buffer window) "CL-USER")
@@ -891,7 +895,7 @@
 
 (defun notify-change-connection-to-wait-message-thread ()
   (bt2:interrupt-thread *wait-message-thread*
-                       (lambda () (error 'change-connection))))
+                        (lambda () (error 'change-connection))))
 
 (defun message-waiting-some-connections-p (&key (timeout 0))
   (with-broadcast-connections (connection)


SequentialDesign:
i don't know what kind of keybind conventions LEM has, but at least in GNU Emacs, `C-c <letter>` is reserved for the user.  i think that is an important convention to have.  how about using `C-c C-s` as in slime/swank since micros is micro slime/swank ?

cxxxr:
I think it's good PR, thank you.
I also agree with changing the command name from slime to micros.
I'd like to do it someday, but first I have to make the existing slime commands deprecated and add micros commands, which is a bit of a hassle, so I haven't done it yet.

> i don't know what kind of keybind conventions LEM has, but at least in GNU Emacs, C-c <letter> is reserved for the user. i think that is an important convention to have. how about using C-c C-s as in slime/swank since micros is micro slime/swank ?

If C-c s is allowed, then C-c m should be fine, right? (Because lem uses micros instead of slime)

I thought that anything starting with C-c was a mode-specific definition.
This has the potential to conflict with user-specific definitions, so how about adopting a different prefix that is only used by the user?
For example, S-<letter>, but this can't be used on a terminal.

cxxxr:
And, although it was a little late, Merry Christmas!

SequentialDesign:
I think `C-c C-m` is a good place to put it, but that is currently `lisp-macrostep-expand` which i like.  i could also see `lisp-macrostep-expand` on `C-c C-x`.  i like `S-` for the user, but i disagree with using `C-c <letter>` since modes usually take `C-c C-<letter>`, and that should be plenty of keys to choose from without infringing on the user's safe space

jfaz1:
> i don't know what kind of keybind conventions LEM has, but at least in GNU Emacs, `C-c <letter>` is reserved for the user. i think that is an important convention to have. how about using `C-c C-s` as in slime/swank since micros is micro slime/swank ?

I generally agree which is why `C-c C-m` fits the best (but would break in the shell). That being said, there are other default binds like `C-c z` for switch to repl, `C-c g` for interrupt, `C-c i`/`C-c I` for inspect that don't follow the convention already, so I felt it's most likely fine for micros as well. Those binds are also in sly/emacs by default.

SequentialDesign:
ahh, i did not know about `C-c z`, `C-c g`, and `C-c I`.  good to know.  `C-c C-m` works for me in either sdl2 or the terminal, but it is `lisp-macrostep-expand` currently

i don't want to be pedantic about it, but i hope we can move those keys to `C-c C-<letter>` or at least a moratorium on adding new ones.  it's one of the conventions from GNU Emacs that i really like since it gives users an area that is just for them

jfaz1:
> ahh, i did not know about `C-c z`, `C-c g`, and `C-c I`. good to know. `C-c C-m` works for me in either sdl2 or the terminal, but it is `lisp-macrostep-expand` currently
> 
> i don't want to be pedantic about it, but i hope we can move those keys to `C-c C-<letter>` or at least a moratorium on adding new ones. it's one of the conventions from GNU Emacs that i really like since it gives users an area that is just for them

Isn't macroexpand `C-c M-m`? I also use `C-c Return`

Issue #1657, opened 2024-11-30
improve: the `vi"` and `va"` in `vi-mode` will freeze the lem editor inside an escaped string.


sakurawald:
To-reproduce:
1. Input `"\"` string.
2. Type `viw` or `vaw` in `vi-mode`
3. The lem editor stucked.

![image](https://github.com/user-attachments/assets/f9b10479-af3a-49c2-8f58-495fd297a482)

sakurawald:
The `vi(` result, it seems like the `escaped close character` is ignored.

![image](https://github.com/user-attachments/assets/1483298d-4c2e-49fa-b9b6-c50351db9d6e)

LeeLaffan:
I can reproduce this issue too. viw/vaw doesn't freeze it for me, but vi" and va" do cause the freeze. Need to restart Lem then.

LeeLaffan:
This is fixed now :)

PR #1716, opened 2024-12-26
vi-mode - Bug - Fix freeze when using _i"/_a" in escaped string

LeeLaffan:
# Change Description

The fix is for the bug in [Issue #1657](https://github.com/lem-project/lem/issues/1657).
This affects code in operator pending mode, so the editor will freeze when using:
 - v i/a "
 - c i/a "
 - d i/a "
 - y i/a "
    
This occurs as when the escape character is found, it does not move the character forward/back, causing an infinite loop of finding the escape char.

As shown in this seek backward example, I have changed it to skip over 2 chars (the escape and the quote char). I did the same for when seeking forward.
```common-lisp
  ;; Before
  ((and escape-char)
    (char= prev-char escape-char))
   
  ;; After 
  ((and escape-char 
            (character-at point -2) ;; Bound check
            (char= (character-at point -2) escape-char))
    (character-offset point -2))
```
    
This behaviour to ignore escape chars, which seems intended for Lem's vi-mode(just bugged), does match vanilla vim so good to have.

# Testing

I have updated the tests to account for this case.
This image shows the rendered test outputs which can help you decipher the string in the tests in you want:
Tests Passing

# Other Bugs

I found some more bugs that are existing already with this functionality, but no more editor freezes, just unexpected behavious in some cases.
I hope to fix them with a bit of changes to the logic around finding quotes specifically.
 - Both have different but unexpected behavior when they are before the first quote/after the last quote
 - Both do not gracefully fail in all cases(one "). This is particularly annoying when doing d_" or c_" where you don't have the visual feedback

### Feature
 - Both do not mimic vim seek ahead in line with operator pending mode

### DIFF
--- a/extensions/vi-mode/README.md
+++ b/extensions/vi-mode/README.md
@@ -39,9 +39,18 @@ Here's a list of all options currently implemented:
   * Default: `nil` (don't show)
   * Aliases: `nu`
 
-## Unittests
-
-```
+## Unit Tests
+    
+### Command Line
+    
+```bash
 qlot install
 .qlot/bin/rove extensions/vi-mode/lem-vi-mode.asd
+```
+    
+### In Lem
+    
+Use `C-c C-r` or:
+```common-lisp Example Test
+  (rove:run-test 'lem-vi-mode/tests/text-objects::word-object)
 ```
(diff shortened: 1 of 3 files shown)


cxxxr:
Thank you!

PR #1717, opened 2024-12-27
make makefile applicable.

snmsts:


jfaz1:
I agree with the changes, but `sudo make install` doesn't find qlot; is there an easy way to have it find it? Will probably have to update the docs.

snmsts:
If you intend to build on root account you should setup qlot on root account as well.
I believe it is not good idea to encourage suspicious activity(build) on privileged user.

how about.

```
make;sudo make install-bin
```

or 

```
make lem;sudo make install
```

jfaz1 (approved):
I'll add a disclaimer to the docs, but this makes sense, thanks!

PR #1721, opened 2024-12-30
Add vi-operator-surrounding-blanks editor-variable

Encryptoid:
# Current Behaviour
Currently, there is strange behaviour in vi-mode where when using `around` in operator pending mode, only with double quotes, where it will include spaces/tabs around the quotes.
This does not match the behaviour of vi/vim. I was going to make a bug Issue for this to fix it, but when I looked into the code and found this seemed to be intentional.

I feel a good goal for vi-mode should be to match vanilla vi/vim as much as possible except where:
  1. There is no default behaviour in vim. (eg. add `viq` for `vi"` as `viq` is not bound in vim)
  2. The same features can be provided with better Lem alternatives (`gd` uses lem's `go-to-def')
    
# Changes
    
In this spirit I kept the current functionality to do this when the editor variable `vi-operator-surround-blanks` is set to `t`. This was simple as this functionality was in a single function.

But the default for this is `nil`, in order to match vanilla vim. In future, many of these could be exposed as a list of vi-editor-variables, but there is only one now.
    
I updated both the README.md and the tests for vi-mode to reflect this.
    
## Readme
I recommend setting the variable as follows based on this macro: [var.lisp](https://github.com/Encryptoid/lem/blob/ad1e66bf418bd9aa20ad03052e26b08ce10ce6ea/src/common/var.lisp#L87)
```common-lisp To enable this variable:
(setf (variable-value 'lem-vi-mode/text-objects:vi-operator-surrounding-blanks :global) t)
```
Is this recommended? I couldn't find examples in anyones configs.
    
I have tested this personally changing back and forth, and the variable can be set at runtime and the behaviour will change.
    
## Tests
Updated the tests using the testing editor-variables macro: [(defmacro with-global-variable-value)](https://github.com/Encryptoid/lem/blob/ad1e66bf418bd9aa20ad03052e26b08ce10ce6ea/src/common/var.lisp#L121)

Github seems to have some unfortunate formatting on the comparison, I haven't changed the existing tests(other than wrapping then in with-global-variable-value). I just added 2 new tests. Locally(in LazyGit) it looks much better:

image

### Tests Passing:
Tests Passing


cxxxr:
Thank you for the PR.
I think it's good, so I'll merge it.
Also, I think @fukamachi knows more about vi-mode than I do.

Issue #1611, opened 2024-11-19
feature: the `undefine-key` function to delete an existing key in a keymap.


sakurawald:
I have the folloing bindings:
```
(defun define-paired-commands ()
  (define-key lem-vi-mode:*normal-keymap* "[ b" 'previous-buffer)
  (define-key lem-vi-mode:*normal-keymap* "] b" 'next-buffer)

  (define-key lem-vi-mode:*normal-keymap* "[ w" 'previous-window)
  (define-key lem-vi-mode:*normal-keymap* "] w" 'next-window)

  (define-key lem-vi-mode:*normal-keymap* "[ t" 'lem/frame-multiplexer::frame-multiplexer-prev)
  (define-key lem-vi-mode:*normal-keymap* "] t" 'lem/frame-multiplexer::frame-multiplexer-next)

  (define-key lem-vi-mode:*normal-keymap* "[ s" 'backward-sexp)
  (define-key lem-vi-mode:*normal-keymap* "] s" 'forward-sexp)

  (define-key lem-vi-mode:*normal-keymap* "[ l" 'up-list)
  (define-key lem-vi-mode:*normal-keymap* "] l" 'down-list)

  (define-key lem-vi-mode:*normal-keymap* "[ f" 'lem/language-mode::beginning-of-defun)
  (define-key lem-vi-mode:*normal-keymap* "] f" 'lem/language-mode::end-of-defun)
  
  (define-key lem-vi-mode:*normal-keymap* "[ i" 'lem-vi-mode/commands::vi-search-backward-symbol-at-point)
  (define-key lem-vi-mode:*normal-keymap* "] i" 'lem-vi-mode/commands::vi-search-forward-symbol-at-point)
  )
```

The `paredit` mode will override the vi-mode bindings starts with `[` and `]`
```
  (lem-paredit-mode:paredit-mode t)
```

I would like to un-define the bindings that defined by `paredit` mode. Is there an easy way to do this ?

sakurawald:
> undefine-key is not yet implemented. could you use override its commands with define-key?

It seems like the `paredit` defines the `[` and `]` keys in `paredit-mode-keymap`, it is conflict with `vi-normal keymap` and `vi-insert keymap`.

https://github.com/lem-project/lem/blob/17639f26c3ed5d7c0f34a85798e79e25829f6478/extensions/paredit-mode/paredit-mode.lisp#L600-L622

I don't know how to override this.

sakurawald:
[...]

PR #1727, opened 2025-01-01
Add (undefine-key) functionality

Encryptoid:
# Overview

I noticed this here, and wasn't going to look into it. But I started getting into paredit myself, so I encountered the same issue. Now it's personal..

Issue: [feature: the undefine-key function to delete an existing key in a keymap.](https://github.com/lem-project/lem/issues/1611)
    
# Change Description

Add public function: `undefine-keys` and marco `undefine-keys`. These are very much in the same style as their `define` counterparts.

# Testing

I tested both functionality in my own config:
```common-lisp
(undefine-key *paredit-mode-keymap* "C-k")

(undefine-keys *paredit-mode-keymap* ("C-k")
  ( "C-L"))
```
    
I also tested attempting to undefine keys that don't exist in the map. This is fine with no errors.

For example with paredit, you can define your own keys before or after undefining the paredit keys, as it will just remove from the map, your keys and paredit will not interfere.

You can undefine keys to your hearts content:

image

### DIFF
--- a/src/internal-packages.lisp
+++ b/src/internal-packages.lisp
@@ -459,6 +459,8 @@
    :*global-keymap*
    :define-key
    :define-keys
+   :undefine-key
+   :undefine-keys
    :keyseq-to-string
    :find-keybind
    :insertion-key-p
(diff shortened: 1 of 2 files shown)


cxxxr:
Thank you very much!

Issue #1726, opened 2024-12-31
improve: [line-numbers-mode] - Sync line format with lines that are displayed


Encryptoid:
# Description
When using a lower number format than the default(6), each line only displays the length that matches that line, as shown in the image: 

![image](https://github.com/user-attachments/assets/de4fc592-3f5c-4cea-a124-05717d936373)

It would be good to match most other editors by having the actual format sync with the lines that are being displayed, so all lines-numbers have the same width. This means when showing 2 digit lines, and you scroll down to a 3 digit line, the 2 digit lines will expand their width to match the 3 digits.

I imagine this is why a high default of 6 was chosen, but it produces a very large block like this:
![image](https://github.com/user-attachments/assets/c6202861-9cc0-4dff-bcbc-fd652dfb0e9a)
It would be good to reduce the default after fixing.

# Steps to reproduce
Set a lower format like this: (or open a file with 1M+ lines!)
```common-lisp
(setf (lem:variable-value 'lem/line-numbers:line-number-format :global) "~1D ")
```

# Recommendations
I am not sure if Lem has a concept yet of which lines are currently being displayed, but this might be a lot easier after that.

It also might not be directly in the handling of the string format, but I think you can also pass a function to work out the line number.

PR #1732, opened 2025-01-04
Fix: [line-number-mode] - Wrapped line attributes & Uniform number width

Encryptoid:
This change is fixes 2 bugs with line-numbers-mode. They are broken up into 2 commits if you wish to view the changes individually. Line numbers are looking a lot smoother now.

# Wrapped lines have no font attributes
Currently, wrapped lines do not have the same font attributes as the line numbers and it looks very strange.

I moved the computation of the wrapped left area to a generic and removed the implementation(not needed when no line numbers). line-numbers-mode then implements this generic to allow using it's attributes to highlight the blank spaces.

I could have just exposed the attribute, but this is now matching with the unwrapped lines, and it allows more flexibility in extensions(adding icons to wrapped lines, etc.)

Before:
image

After:
image

# Line numbers format match max buffer lines
This is a fix for my own issue: [improve: [line-numbers-mode] - Sync line format with lines that are displayed](https://github.com/lem-project/lem/issues/1726)

I looked at a few editors, and most of the handle this by making the line number width for all lines match the amount of digits in the last line of the buffer. Some base it of the visual lines on the screen(if you can see a higher digit), but it's fully customizable via a minor mode now so people can change it if they wish.

Before:
image

After:
image

### DIFF
--- a/src/display/base.lisp
+++ b/src/display/base.lisp
@@ -16,6 +16,10 @@
 (defgeneric compute-left-display-area-content (mode buffer point)
   (:method (mode buffer point) nil))
 
+(defgeneric compute-wrap-left-area-content (left-side-width left-side-characters)
+  (:method (left-side-width left-side-characters)
+    nil))
+
 (defvar *in-redraw-display* nil
   "T if the screen is currently being redrawn by `redraw-display`.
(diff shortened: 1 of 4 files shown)


cxxxr:
Thank you!

PR #1737, opened 2025-01-05
Implemented improved Python indent function.

funk443:
The new indent function for Python mimics the behaviour of indent function as seen in Emacs, where hitting tab key will cycle through the following positions:

- Same indent level as last line.
- +1 indent level relative to last line.
- -1 indent level relative to last line.

There're still a flaw in this function, that I currently have no idea how to fix:

```python
         v Cursor.
foo = bar|
```

after `newline-and-indent`:

```python
foo = bar
    |
    ^ It should not have indented, since the last line is just a
      top-level declaration.
```

### DIFF
--- a/extensions/python-mode/python-mode.lisp
+++ b/extensions/python-mode/python-mode.lisp
@@ -91,10 +91,21 @@
            (column (point-column (back-to-indentation point)))
            (next-indent-column (+ last-line-indent-column tab-width))
            (previous-indent-column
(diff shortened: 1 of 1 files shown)


funk443:
I've did some little tweak, so now it also indents last line ends with `:`, `(`, `[` or `{`.

cxxxr:
I know it's a difficult problem, but I have high hopes for it.

Issue #1698, opened 2024-12-10
improve: git grep gets `SDL Error (#<SDL-SURFACE {#X00000000}>): Text has zero width` or `crash the lem`.


sakurawald:
To reproduce:
1. Run `git clone git@github.com:lem-project/lem-project.github.io.git`
2. Reset the `HEAD` to `d5e3548aef9eadf2ed36fb274c4105a9344b39e6`.
3. Open the project using `lem`.
4. Run `lem/grep::project-grep` with arg `git grep -niHi 'grep'`.
5. Get the error.

sakurawald:
It seems like the big js file causing the error, or some problems about the string escape.

![image](https://github.com/user-attachments/assets/4e9033af-610b-4eff-8c62-7cbdc239b0a8)

sakurawald:
Another interesting command: Run `lem/grep::project-gre` with `git grep -nHI 'webpack'`.
And the `lem` will crash directly without any error message.

funk443:
> To reproduce:
> 
>     1. Run `git clone git@github.com:lem-project/lem-project.github.io.git`
> 
>     2. Reset the `HEAD` to `d5e3548aef9eadf2ed36fb274c4105a9344b39e6`.
> 
>     3. Open the project using `lem`.
> 
>     4. Run `lem/grep::project-grep` with arg `git grep -niHi 'grep'`.
> 
>     5. Get the error.

This should be fix by #1741.

> Another interesting command: Run `lem/grep::project-gre` with `git grep -nHI 'webpack'`. And the `lem` will crash directly without any error message.

I believe this is because Lem cannot handle so much super long lines.

ryan-carpenter:
I get this error when I attempt to open the dashboard, maybe because of the `<U+E908>` `<U+F0C9>` and `<U+F011>` characters. Is this different than the issue reported here and the one meant to be fixed by #1741?

https://github.com/lem-project/lem/blob/8d5c80dfbea05b0997b21bb061c8c370f157667e/extensions/lem-dashboard/default-dashboard.lisp#L57

SBCL 2.5.0-1.2-suse
8d5c80dfbea05b0997b21bb061c8c370f157667e

PR #1739, opened 2025-01-09
Lem SDL2 can now properly display zero-width characters.

funk443:
Previously, if someone tried to open a file that contains zero-width characters, such as `U+200B`, `U+200C`, `U+200D`, and `U+FEFF`, Lem SDL2 will just refuse to display any content, leaving the user with a blank window.  This is probably because SDL2 does not like zero-width characters.

So now I try to fix this problem by adding another `char-type` dedicated for zero-width characters, and display them with a special character.

I've also created a new attribute for zero-width characters and control characters, so users won't confuse them with real contents.

### DIFF
--- a/frontends/sdl2/display.lisp
+++ b/frontends/sdl2/display.lisp
@@ -147,13 +147,13 @@
         ((eq type :braille)
          (display-braille-font display))
         (bold
-         (if (eq type :latin)
-             (display-latin-bold-font display)
-             (display-cjk-bold-font display)))
+         (case type
+           ((:latin :zero-width) (display-latin-bold-font display))
+           (otherwise (display-cjk-bold-font display))))
         (t
-         (if (eq type :latin)
-             (display-latin-font display)
-             (display-cjk-normal-font display)))))
+         (case type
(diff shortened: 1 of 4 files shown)


sakurawald:
I think this may be a relative issue, would you like to test it?
https://github.com/lem-project/lem/issues/1698

funk443:
> I think this may be a relative issue, would you like to test it? #1698

Interesting, it seems Lem just hang instead of giving out that SDL error while reproducing this issue.  Maybe there are other weird characters in these documents which cause these problems.

image

funk443:
So I just checked if Lem can now open the file that I wanted to open, turns out it still gives out the same zero-width error from SDL2.  So I went to see what kind of characters this document has, it turns out it does not contain any zero-width character AT ALL, then it must be some other character causing the issue.

The document: https://raw.githubusercontent.com/dannyvankooten/1brc/refs/heads/main/README.md

funk443:
I think the problem is cause by Lem not recognising `🏎️`, which in terms of unicode characters is `\u1F3CE\uFE0F`

PR #1741, opened 2025-01-09
Fix zero-width characters and control characters attribute issue.

funk443:
Since 1e42f96b44e4c2eb664100858c482a480526a4b4, if you put the cursor on the different line than the line those special characters are on, Lem will signal an error.  This issue is fixed in this PR.


funk443:
Also we should take a look on how we should treat Unicode variation selectors, as 87deab863b99a5b1e8e98336cdb333a45cf282b4 is only a workaround.

PR #1744, opened 2025-01-13
Add coalton-mode for basic features.

fukamachi:
* Code highlighting for `.coal` files.
* Indentation support.
* Show autodoc for Coalton functions.
  * Its arguments can be different from definitions since the information is from Swank, just like Common Lisp.

Note that the indentation code is copied from lem-lisp-syntax since it doesn't take any customized rules.


cxxxr:
Thank you

PR #1745, opened 2025-01-13
Lem intelligence

cxxxr:


erickisos:
Hey @cxxxr!

What do you think about integrating something like eca.dev instead into the editor? The main advantage  being that it's agnostic  of the LLM Model/Providers while still getting some default cool features.

PR #1764, opened 2025-02-02
support mark-(next|previous)-like-this

Prikaz98:
There is a similar [PR](https://github.com/lem-project/lem/pull/1517), but I didn't use isearch to find the next marks.

mark-(next|previous)-like-this aims to work like [multiple-cursors.el](https://github.com/magnars/multiple-cursors.el) This means that if there is nothing to select, it marks the next or previous line.
Handling offsets is a bit complex because we need to maintain the same starting or ending position of the region when calling mark-next or mark-prev.

Demo:
lem-multicursor

TODO:
- [x] Figure out how to write the test
- [x] Fix the bug where duplicate marks appear when spamming the commands add-cursors-to-(next|prev)-line (previously handled by clear-duplicate-cursors; I believe we can address this within the function that determines the next cursor position)


Prikaz98:
Also add additional thing with C-g double click behavior (first cancel mark region, second clean multicursors)
lem-multicursor-double-c-g-behavior

Prikaz98:
I have a problem with test running.  I read an instruction from [this page](https://github.com/lem-project/lem/wiki/How-to-hack-on-lem-itself#running-the-tests)
and checked [rove](https://github.com/fukamachi/rove) project but unfortunately I've got the same result. 
Or this
```bash
.qlot/bin/rove lem-tests.asd

Testing System lem-tests

✓ 0 tests completed

Summary:
  All 0 tests passed.
```
no test executed

or 
```bash
.qlot/bin/rove tests/cursors.lisp

Unhandled ASDF/FIND-COMPONENT:MISSING-COMPONENT in thread #<SB-THREAD:THREAD tid=1070475 "main thread" RUNNING
                                                             {1003F80003}>:
  Component "lem-tests/cursors" not found
````

If someone know what wrong, please let me know :pray:

cxxxr:
lem-tests is not a package-inferred-system, so it is not possible to run tests on a file-by-file basis from rove.
If you want to test individually, how about running 'run-test' from the REPL?

Prikaz98:
I think that's enough for this PR. If you see any issues in my code, feel free to leave comments :)

Prikaz98:
Oh, yes, it is a similar function. I didn't know that this function already exists in isearch module. Let's close this PR and I just keep it in my personal build.

Also it will be great to add mention in Lem doc in [multiple cursors](https://lem-project.github.io/usage/usage/#multiple-cursors) description. Because when I read it I thought: "Oh, there is no my most useful function mc mode. I must implement this"

cxxxr:
Yeah, I'm sorry there are no documents. That's something we need to improve.
Also, since the functions of isearch are incomplete, I would be happy if they could be integrated with this PR.

Prikaz98:
Merged prior changes with the existing implementation in the isearch module. Hopefully, it turned out pretty nice.

lem-multicursors-isearch

cxxxr:
Thank you so much!
it can't support the behavior of having multiple cursors on the same line that was supported before this change, but I will merge this PR itself.

PR #1766, opened 2025-02-07
Fix grammar of window movement commands

solanav:
Changed "move to the window on the up" and "move to the window on the down" to "move to the window above" and "move to the window below" respectively.


cxxxr:
Thank you

PR #1767, opened 2025-02-08
recenter-top-bottom (emacs-mode)

Prikaz98:
Emacs default behavior of (C-l recenter-top-bottom)

lem-recenter-top-bottom

### DIFF
--- a/src/window/virtual-line.lisp
+++ b/src/window/virtual-line.lisp
@@ -28,7 +28,6 @@ If cursor is on top then move move WINDOW to the bottom."
   (let* ((line (window-cursor-y window))
          (window-height (window-height-without-modeline window))
          (middle (floor window-height 2))
-         (bottom (window-height-without-modeline window))
          (scrolloff (min (floor (/ (window-height window) 2)) 0))
          (top 0))
     (cond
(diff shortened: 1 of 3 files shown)


cxxxr:
Thank you for the PR.
BTW since lem is not emacs, I don't think it's necessary to match the emacs default.
Is it possible to add just the command and change it in the config?

For example, I would like to have options to change the behavior of recenter in the future (such as the behavior of JetBrains C-l).

Prikaz98:
>  Is it possible to add just the command and change it in the config?
Yes, I don't mind. I believe it will be better to return default binding (C-l 'recenter) because changing behavior can affect some user who already got used to 'recenter and 'recenter-top-bottom will be just an option to configure, for users like me who like this feature.

src/window/virtual-line.lisp
```diff
@@ -21,6 +21,21 @@ If FROM-BOTTOM is T, start counting from the bottom."
...
       n)))
 
+(defun window-recenter-top-bottom (window)
+  "In first call recenter WINDOW to the middle line.
+If cursor is already in the middle of WINDOW then move cursor in the top position.
+If cursor is on top then move move WINDOW to the bottom."
+  (let* ((line (window-cursor-y window))
+         (window-height (window-height-without-modeline window))
+         (middle (floor window-height 2))
+         (bottom (window-height-without-modeline window))
```
cxxxr, line 31: sblint gives a warning.

cxxxr (commented):
Thank you so much!
I made one comment, but I think it's good.

PR #1770, opened 2025-02-15
fix: scala-mode calc-indent-function

Prikaz98:
I found a bug in `scala-mode` `calc-indent-function`. `lem-js-node::js-calc-indent` breaks the indents in some cases. I tried to fix it in this PR.

In the following screen record I've launched two instance of Lem (without fix and with fix) and indent a file .scala that shared between them

lem-scala-mode-indent


cxxxr:
Thank you!

PR #1777, opened 2025-03-01
Use gitignore when finding project files

jfaz1:
This PR takes into account .gitignore when finding project files in git repos by default. Toggle with `*respect-gitignore*`.

Besides having more accurate autocomplete (e.g. ignoring .git/node_modules), `project-find-file` is actually usable on large projects now (e.g. autocomplete Linux kernel would take ~5 seconds to show up, now shows up instantly).

### DIFF
--- a/src/commands/project.lisp
+++ b/src/commands/project.lisp
@@ -99,7 +99,7 @@
                        ;; (try to) handle non-unix platforms path separators.
                        (add-directory-separator marker) pathname)
         if (member subpath subdirectories :test #'equal)
-          return t
+        return t
         finally (return)))
 
 (defvar *recurse* t)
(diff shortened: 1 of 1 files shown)


src/commands/project.lisp
```diff
@@ -136,22 +137,69 @@
         (find-root (parent-directory pathname) :recurse recurse :recursing t)
         pathname)))))
 
+(defvar *respect-gitignore* t
+  "If non-nil, project-find-file will respect .gitignore files.")
+
+(defun git-repository-p (pathname)
+  (let ((subdirectories (list-subdirectories pathname)))
+    (member (uiop:merge-pathnames*
```
vindarel, line 145: what about `(uiop:directory-exists-p "xxx/.git")`?

src/commands/project.lisp
```diff
@@ -136,22 +137,69 @@
...
          (project-root (find-root cwd))
          (root (or project-root cwd)))
     (uiop:with-current-directory (root)
-      (let ((filename (prompt-for-files-recursively))
-            buffer)
+      (let ((filename 
+              (if (and *respect-gitignore* (git-repository-p root))
+                  ;; In a git repo and respecting gitignore, use git ls-files
+                  (let ((files (list-project-files-git root)))
+                    (if files
```
vindarel, line 185: not a fan of nested IFs personally. We have no style rule for this.

src/commands/project.lisp
jfaz1: Ah yeah that's much better I had something else here before :smile:

src/commands/project.lisp
jfaz1: Makes sense, I consolidated them

vindarel:
wait! Now this command will list all files including all `node_modules/` that are normally ignored.

How are they ignored? Automatically by a call to the `fdfind` program. Which also respects .gitignore, .ignore (and .fdignore) by default. You didn't know about that, surely?

For a full Lisp solution, very welcome, we should ignore node_modules too :S What other big repositories should we ignore?

src/commands/project.lisp
```diff
@@ -136,22 +137,61 @@
         (find-root (parent-directory pathname) :recurse recurse :recursing t)
         pathname)))))
 
+(defvar *respect-gitignore* t
+  "If non-nil, project-find-file will respect .gitignore files.")
+
+(defun list-project-files-git (directory)
+  "List all files in a git project using git ls-files."
+  (handler-case
+      (let ((cmd "git ls-files --cached --others --exclude-standard"))
```
vindarel, line 146: this command should be tweakable -> global parameter.

vindarel:
I'm sorry, I'm adding a section in the Install doc to mention it earlier.

But wait! again, because if `fd` isn't present the project command relies on a Lisp function that *doesn't have the same behavior of ignoring files* (with a slight "IIRC" prudency touch, details to be double-checked). So your code would be useful for the fallback case.

jfaz1:
> I'm sorry, I'm adding a section in the Install doc to mention it earlier.

You're good, I'm the one that missed it :laughing: 

> But wait! again, because if `fd` isn't present the project command relies on a Lisp function that _doesn't have the same behavior of ignoring files_ (with a slight "IIRC" prudency touch, details to be double-checked). So your code would be useful for the fallback case.

Yeah, if `fd` isn't present it uses the function that doesn't ignore them which is why I made the PR. I'm on a clean install which is why I noticed the .gitignored files showing up, I had `fd` in my previous setup, just didn't realize that was why. I think maybe instead of using this as a fallback, we could add `fd`/`fd-find` packages to the install instructions? It makes for a better experience in general anyways.

Issue #1522, opened 2024-09-07
In vi mode `:'<,'>s/^/;;/` leads to a crash.


g-gundam:
## Steps to Reproduce

1. `M-x vi-mode`
2. Select a range of lines using `V`
3. Hit `:` and then type `s/^/;;/`
    - The purpose of this regexp substitution is to put ";;" at the beginning of every line in the region.

## Expectation

The lines in the region should have ";;" prepended to them.

## Actual Result

Lem is unresponsive to input, and I saw this when I switched back to the terminal that I started Lem on.

```
Heap exhausted during garbage collection: 0 bytes available, 64 requested.
Gen  Boxed   Code    Raw  LgBox LgCode  LgRaw  Pin       Alloc     Waste        Trig      WP GCs Mem-age
 2   20760      0      0      0      0      0    5   339727520    404320   214536266       0   1  1.3998
 3   38774      1     73      0      0     24   68   636059232    819616     2000000   18797   0  0.4859
 4       0      0      0      0      0      0    0           0         0     2000000       0   0  0.0000
 5       0      0      0      0      0      0    0           0         0     2000000       0   0  0.0000
 6    3556     13   1762    350      0    222    0    94756640   1958112     2000000    5608   0  0.0000
           Total bytes allocated    =    1070543392
           Dynamic-space-size bytes =    1073741824
GC control variables:
   *GC-INHIBIT* = true
   *GC-PENDING* = true
   *STOP-FOR-GC-PENDING* = false
fatal error encountered in SBCL pid 2070099 tid 2070132:
Heap exhausted, game over.

Welcome to LDB, a low-level debugger for the Lisp runtime environment.
ldb> 
```

Encryptoid:
Hi, this is now fixed with: https://github.com/lem-project/lem/pull/1725

For a brief overview, the editor was getting stuck in an infinite loop while trying to highlight regex that evaluated to an empty string, like "^". The same was true for other substitutions like "$", etc.

g-gundam:
Nice.  Thanks for fixing it.

PR #1782, opened 2025-03-15
Fix query-replace to highlight all candidates

masatoi:
Hello everyone.

I have found a small bug and am trying to fix it.
When I run `query-replace`, only the first hit word is highlighted and I can't see where the word is after the second.

Looking at the current `highlight-region` implementation, only one search result is added to isearch-overlays.
It seems that the parentheses of `unless` have the wrong correspondence and it is not used to determine whether to exit the loop.

The following images show the before and after behaviour.

SS 2025-03-15 11 28 19

### DIFF
--- a/src/ext/isearch.lisp
+++ b/src/ext/isearch.lisp
@@ -212,9 +212,10 @@
       (when (string= search-string "")
         (return-from highlight-region nil))
       (with-point ((p start-point))
-        (loop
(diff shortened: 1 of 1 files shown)


masatoi:
This change seems to reproduce [this issue](https://github.com/lem-project/lem/issues/1522).

cxxxr:
Thank you so much!

Issue #413, opened 2019-08-26
Things I miss (a lot!) in Lem


svetlyak40wt:
Today I've tried to use Lem for work on one of my pet projects. Here is the list of things I was lacking to be productive:

# Table of Contents

1.  [Things I miss (a lot!) in Lem](#org2e0092d)
    1.  [A ChangeLog.md](#org1027185)
    2.  [Bindings](#org346f295)
        1.  [C-c ~ (or C-c v) repl-sync to syncronize current package](#orga92f4e9)
        2.  [C-w to delete previous word](#org3e1d0c5)
        3.  [Undo is not bound to C-/](#orgc2b7936)
        4.  [Paredit should indent line after the new-line](#orgc9008fc)
        5.  [Something like C-u C-Space or pop-global-mark](#org57c4295)
        6.  [C-s C-w - to search thing a point](#org0f768fe)
        7.  [Double C-s should start searching a previous term](#org69d841a)
        8.  [C-c x to export symbol at point (as in SLY)](#org8678739)
    3.  [Gotchas](#org7ce0859)
        1.  [Unable to start Lem's lisp repl when you already connected to other Lisp process and have a REPL](#org5ac5348)
        2.  [Lisp-mode should kee history between restarts](#org090e4b0)
        3.  [Grep does not work like expected](#org6f11ebc)
        4.  [C-u C-c C-c should compile defun with (optimize (debug 3) (safety 3) (speed 0))](#org54c9663)
    4.  [Modes](#org4a40ffc)
        1.  [Package Manager](#org8aa48cb)
        2.  [Lispy-mode](#org1658d7c)
        3.  [Bookmarks](#orgd539d40)
        4.  [Org-Mode](#orga2983bb)
        5.  [Magit](#org2987cb2)
        6.  [Helm](#org4c5adc9)
        7.  [Projectile (with helm integration)](#org3ea02b5)
        8.  [YaSnippet](#orga62b355)

<a id="org2e0092d"></a>

# Things I miss (a lot!) in Lem

<a id="org1027185"></a>

## A ChangeLog.md

<a id="org346f295"></a>

## Bindings

<a id="orga92f4e9"></a>

### C-c ~ (or C-c v) repl-sync to syncronize current package

<a id="org3e1d0c5"></a>

### DONE C-w to delete previous word

-   State "DONE"       from              <span class="timestamp-wrapper"><span class="timestamp">[2019-08-26 Mon 09:27]</span></span>

(Can be bound to backward-delete-word

[...]

PR #1785, opened 2025-03-30
Implement mark-ring

apr3vau:
Integrate https://github.com/lem-project/lem/issues/413#issuecomment-2746231770 into main repo

Notes:

- Ring rotating function needs to check whether a location is "empty", so it needs extra changes to generalize it. To minimize changes, I embedded the specialized version into `pop-buffer-ring`.
- Added a hook to `*set-location-hook*`, so that commands like `move-to-end-of-buffer` will also push location to mark-ring, same with Emacs do.


jfaz1 (approved):
LGTM and works well, very useful :+1:

cxxxr:
Thank you for the PR.
I hadn't noticed it because I hadn't checked in a while, but I think it's good, so I'll merge it.

Issue #1700, opened 2024-12-10
improve: the `type` of `paste-yank` fell through ECASE expression.


sakurawald:
To re-produce:
1. Launch `lem` with `vi-mode` enabled.
2. Press `p` key to paste from yank.
3. Get the error.

![image](https://github.com/user-attachments/assets/d103d472-196a-4a57-8a48-e89437e383d0)

PR #1794, opened 2025-04-16
Fix broken paste in vi-mode

not-omer:
### Overview
Fix broken paste in vi-mode. (https://github.com/lem-project/lem/issues/1700)

- We are not correctly handling the case when `(get-clipboard-data)` returns nil.
- `(get-clipboard-data)` was erroneously returning a SB-SYS:INT-SAP alongside the clipboard string which was being passed into `vi-paste-before/after` instead of a `:type`.

### Testing
- `make`
- Open new buffer in vi-mode
- Press `p` to yank
- No more backtrace


cxxxr:
Thank you for your PR.
I apologize for the delay in responding.

PR #1795, opened 2025-04-17
Add filer-delete-select for filer mode

lu4nx:
Hello, I'm adding the filer-delete-select (C-c C-d) command to delete selected file.

### DIFF
--- a/src/ext/filer.lisp
+++ b/src/ext/filer.lisp
@@ -13,7 +13,7 @@
 
 (define-key *global-keymap* "C-x d" 'filer)
 (define-key *filer-mode-keymap* "Return" 'filer-select)
-(define-key *filer-mode-keymap* "C-c C-d" 'filer-delete-select)
+(define-key *filer-mode-keymap* "D" 'filer-delete-select)
 
 (defclass item ()
   ((pathname :initarg :pathname
--- a/src/ext/filer.lisp
+++ b/src/ext/filer.lisp
@@ -202,7 +202,7 @@
       (when (and file
                  (prompt-for-y-or-n-p (format nil "Do you really want to delete \"~A\"?" file)))
         (handler-case
-            (delete-file file)
+            (lem/directory-mode/file:delete-file* file)
           (sb-ext:delete-file-error (e)
             (editor-error (format nil "~A" e))))
         (when (not (uiop:file-exists-p file))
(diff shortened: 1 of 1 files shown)


src/ext/filer.lisp
```diff
@@ -190,3 +191,21 @@
...
+  (let ((item (text-property-at (back-to-indentation (current-point)) :item)))
+    (unless item
+      (editor-error "No file selected")
+      (return-from filer-delete-select))
+    (let ((file (item-pathname item)))
+      (when (and file
+                 (prompt-for-y-or-n-p (format nil "Do you really want to delete \"~A\"?" file)))
+        (handler-case
+            (lem/directory-mode/file:delete-file* file)
+          (sb-ext:delete-file-error (e)
```
vindarel, line 206: I suggest adding a `#+sbcl` feature flag: even though SBCL is the main implementation we use, there has been work to run Lem on others.

(and so have a `#-sbcl` for other impls too)

vindarel:
Nice +1

About naming: this command looks to be for one file, the file at point. "filer-delete-select" sounds vague to me: is it about select(ed) file(s)? Did you consider other command names? like "filer-delete-file-at-point"

src/ext/filer.lisp
```diff
@@ -13,6 +13,7 @@
 
 (define-key *global-keymap* "C-x d" 'filer)
 (define-key *filer-mode-keymap* "Return" 'filer-select)
+(define-key *filer-mode-keymap* "C-c C-d" 'filer-delete-select)
```
vindarel, line 16: Also what about `D` (capital d) for the keybinding?

vindarel: just like in directory-mode. See src/ext/directory-mode/commands.lisp and keybindings.lisp (directory-mode-delete-files)

lu4nx: OK, I agree.

lu4nx (commented):
Change keymap and add the #+sbcl

lu4nx:
> Nice +1
> 
> About naming: this command looks to be for one file, the file at point. "filer-delete-select" sounds vague to me: is it about select(ed) file(s)? Did you consider other command names? like "filer-delete-file-at-point", "filer-delete-file"…

I choose filer-delete-file :-)

vindarel:
hey don't you think you need to handle the non-SBCL case with `#-sbcl`?

lu4nx:
> hey don't you think you need to handle the non-SBCL case with `#-sbcl`?

Fixed

cxxxr:
I think it's good.

PR #1796, opened 2025-04-20
Bugfix: peek source window remains stuck in the view

tomaneo:
If "peek source" window was already open when a new one was opened then the first one was never removed from the view and it was not possible to close it afterwards.

For example second M-x grep before previous grep window is closed can reproduce this bug.


cxxxr:
Thank you!

PR #1806, opened 2025-04-27
Improved parenthesis highlighting in vi insert mode

tomaneo:
Matching parenthesis highlighting was not very useful in vi insert mode, because it didn't highlight the opening parenthesis for just inserted closing parenthesis. Only after switching back to normal mode you could see the matching pair highlighted.

This patch changes insert mode behaviour to use parenthesis before the cursor as the highlighted pair in insert mode. This way it is also similar to emacs mode.

Before the patch:
parens-original

After the patch:
parens-improved


vindarel:
Thank you!! This was making editing harder.

PR #1810, opened 2025-05-05
Add `make install` targets

jfaz1:
Currently `make install` always installs the `sdl2-ncurses` variant. 

This PR allows users to also `make install-ncurses`, `make install-sdl2` (default), or `make install-sdl2-ncurses`. This mirrors the standard make process better (i.e `make sdl2`) and also lets users only install the variant they want.


cxxxr:
Thank you!

PR #1811, opened 2025-05-05
Add meta key to modifier text input (sdl2-compat fix)

### LINKED ISSUE #1800: Bug: Meta modifier only works on alphanumeric keys in combination with Control
Fairly weird issue, but: 

Using SDL2 frontend, Meta key gets detected in two cases only

1. The "main" key pressed isn't alphanumeric (for example, `M-F1` and `M-Delete` work)
2. Control is also held at the same time (for example, `M-C-x` works)

This issue doesn't seem to affect the ncurses frontend

Lem version: Git, commit: `a392523`
OS: Linux, 6.12.24-1-lts
Desktop: Hyprland (Wayland), version 0.48.1

Will be glad to provide any other information, because my CL skills weren't enough to debug this myself beyond figuring out that the modifier is actually accounted for at the lem-sdl2/keyboard level

jfaz1:
Closes #1800 

It seems that when meta  is pressed with an alphanumeric key, SDL2-compat still generates TEXTINPUT events (unlike native SDL2). This PR makes meta key combinations work again with SDL2-compat.

### DIFF
--- a/frontends/sdl2/keyboard.lisp
+++ b/frontends/sdl2/keyboard.lisp
@@ -132,7 +132,8 @@
 
 ;; linux
 (defun modifier-is-accept-text-input-p (modifier)
-  (not (modifier-ctrl modifier)))
+  (and (not (modifier-ctrl modifier))
+       (not (modifier-meta modifier))))
 
 (defmethod handle-text-input ((platform lem-sdl2/platform:linux) text)
   (handle-text-input-unix text))


nassir90:
This resolved the issue

SequentialDesign:
nice.  meta key works again

jfaz1:
I thought I tried on standard SDL2, but looks like the one I built wasn't used by Lem and it just used my system one :face_exhaling: 

In other words, not suitable to merge.

I think the question is whether it's a bug with SDL2-compat and we should open an issue there, or whether it's something on our end or we can make a hotfix. In the meantime I'll leave this PR up so those on SDL2-compat have a way to deal with it.

Pinging President @cxxxr for his opinion :saluting_face:

cxxxr:
Let's deprecated lem-sdl2

mantacid:
If it means anything, I was able to fix the key issue by replacing `sdl2-compat` with `sdl2`.

novemberist:
Why was this fix never merged? As far as I can tell it works as intended with sdl2-compat.

But I guess it doesn't matter anymore since sdl2 was deprecated.

ethanxxxl:
I think this PR should be merged, it fixes the issue (I tested on arch linux with and without sdl2-compat).  Even though SDL2 is deprecated, if some motivated individual wants to play around with it, we might as well have known working code merged.

perhaps @cxxxr would reconsider this merge?

cxxxr:
Indeed, I think merging would be fine.

Issue #1813, opened 2025-05-07
Github Actions - Unit Tests not running


theangelperalta:
It seems like the unit tests aren't running in Github Actions. I noticed this because I'm working on a Pull Request and wanted to run some new tests: https://github.com/lem-project/lem/actions/runs/14846558131/job/41681686735#step:4:547.

![Image](https://github.com/user-attachments/assets/84f17d92-2aa7-49ea-b616-bf9bb43e6d6b)

This seems to be related to ASDF version bundled with sbcl: https://github.com/fukamachi/rove/issues/71

theangelperalta:
Fixed in https://github.com/lem-project/lem/pull/1814

PR #1814, opened 2025-05-09
[FIX] Unit tests not running due to asdf issue

theangelperalta:


ryukinix:
@theangelperalta 
That would be interesting to get back to work properly. I want to help.

I see that reading #1813, some bugs/errors were introduced since this, reading the `CI / build` logs, 3 tests are broken: 

```
    - lem-tests/isearch
    - lem-tests/self-insert-command
    - lem-tests/scala-mode
 ```
 
In my view, it would be better to use `skip` of rove on the broken tests and accept this PR instead of allowing more PRs to be merged without tests. This behavior can get worse, and the main branch can degrade over time.

File: Makefile [hunk 1]
```diff
@@ -51,6 +51,7 @@ install-sdl2-ncurses:
 test:
 	qlot install
 	.qlot/bin/rove lem-tests.asd
+	.qlot/bin/rove extensions/vi-mode/lem-vi-mode.asd
```
ryukinix, line 54: There is no way to this be tested in a single rove call? I think that lem-tests system should take care of it, otherwise, we'll have two reports in the ci

extensions/vi-mode/tests/visual.lisp
```diff
@@ -77,14 +77,20 @@
         (cmd "v3iw")
         (ok (buf= #?"<foo ba[r]>\n"))))
     (testing "va\""
+        (skip "Skipped temporarily...")
```
ryukinix, line 80: Particularly, I would create a new issue only about fixing these skipped tests and attach the issue id or even the link here

ryukinix (changes requested):
A few suggestions

Makefile
theangelperalta: It seems like the two packages are mutually exclusive; therefore, two separate calls are need in the current setup. I've actually noticed this is the case for most, if not all, the lem extensions: https://github.com/lem-project/lem/blob/main/.github/workflows/test.yml#L46-L47

ryukinix: Ok! If there is no a easy option, I think we can improve this later, if we find a way further.

ryukinix (approved):

PR #1817, opened 2025-05-16
Unify vi-mode visual state with lem buffer marks

### LINKED ISSUE #1590: Feature: sync the `mark region` between `vi-visual-mode` and the mark region selected by `mouse-clicking`.

tomaneo:
Modified the visual states in vi mode to use lem buffer mark as starting point and initiator of visual mode regions. To support synchronisation some small additions to core lem implementation was needed, mainly hooks for enabling and disabling the buffer mark.

This allows activating visual mode from any of the "mark-xxx" commands available on emacs mode side, including mouse operations.

Vi states are now buffer specific, not global. This fixes several small glitches and annoyances when switching between windows and buffers

Closes #1590

### DIFF
--- a/extensions/vi-mode/commands.lisp
+++ b/extensions/vi-mode/commands.lisp
@@ -380,7 +380,7 @@ Move the cursor to the first non-blank character of the line."
 (define-operator vi-substitute (beg end type) ("<R>")
     (:motion vi-forward-char)
   (vi-delete beg end type)
-  (change-state 'insert))
+  (setf (buffer-state) 'insert))
 
 (define-operator vi-delete-next-char (beg end type) ("<R>")
     (:motion vi-forward-char)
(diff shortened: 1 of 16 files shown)


tomaneo:
Needs improvements.

cadrgtsecond:
I would also like to see this merged in, since it would help in porting other modal text editing styles, like Kakoune

Sasanidas:
I think this feature is really interesting, ping @cxxxr

cxxxr:
Sorry for the late reply.
Thank you for the PR, I'll merge it.

PR #1818, opened 2025-05-17
Bugfix: abbrev command errors with subsequent calls

### LINKED ISSUE #1802: Vi-mode error
Pressing ctrl-p in insert mode when no completions can be found causes an error.

To replicate

- Enter vi-mode
- Open a file
- Enter insert mode
- Begin typing a random string that has no completions in the current file
- Attempt pressing ctrl-p twice

First press of `ctrl-p`
Image
Second press of `ctrl-p`
Image

tomaneo:
If abbrev was called second time after the first call didn't have matches it tried to insert nil string.

Fixes #1802


Sasanidas:
Indeed, that solves the bug, LGTM

PR #1820, opened 2025-05-23
Fixed eternal loop when searching with regexp $

tomaneo:
Vi-mode ex command s/$/whatever/ triggered the issue.


Sasanidas:
LGTM

PR #1833, opened 2025-07-23
Make extension-commands.lisp dependent on if quicklisp is present

Fuglesteg:
The extension commands are dependent on quicklisp and should only be loaded if quicklisp is available in *features*

### DIFF
--- a/lem.asd
+++ b/lem.asd
@@ -215,7 +215,7 @@
                              (:file "deepl")
                              (:file "themes")
                              (:file "detective")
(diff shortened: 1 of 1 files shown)


cxxxr:
Thank you for the PR.
I see a difference between the PR description and the change details.
How about splitting the addition of vue-mode into a separate PR?

Fuglesteg:
Apologies, I pushed the vue-mode commit to the wrong branch 😅

PR #1834, opened 2025-07-23
Add vue-mode

Fuglesteg:
Add `vue-mode` a major mode based off `js-mode` to augment the LSP to work with the [Vue framework](https://vuejs.org).

Minor change made to the `define-language-spec` macro in order to support subclassing language specs.

The mode exposes the variable `*vue-language-server-location*` so that the user can specify alternative installation paths for the Vue language server. In case the user installs packages in a non-default location.

Further work would be to change the js-grammar so that it can better highlight the keywords and other oddities of the vue template syntax.


extensions/lsp-mode/lsp-mode.lisp
```diff
@@ -1789,12 +1789,12 @@
 (defun enable-lsp-mode ()
   (lsp-mode t))
 
-(defmacro define-language-spec ((spec-name major-mode) &body initargs)
+(defmacro define-language-spec ((spec-name major-mode &optional super-spec) &body initargs)
```
cxxxr: It might be better to use keyword.

Fuglesteg:
@cxxxr I have a weird bug when I changed the parent-mode of vue-mode from language-mode to js-mode. The LSP no longer gets set up right, seems like the initialization-options don't apply. Maybe you know what that is about?

Fuglesteg:
> @cxxxr I have a weird bug when I changed the parent-mode of vue-mode from language-mode to js-mode. The LSP no longer gets set up right, seems like the initialization-options don't apply. Maybe you know what that is about?

Nevermind, I think I understand it. The hook for starting the LSP for js also gets triggered in vue-mode, then because I didn't set a language-id the vue-mode hook will just connect to the js workspace instead of starting a new LSP.

However, if I add a language-id it will now start two LSPs one for javascript, one for vue, and give an error to the user that the 'vue' workspace was not found, because now the language-id of the js-spec, which starts from the hook, doesn't match with the vue-spec.

I think the best solution for now is to just set the parent-class to language-mode, but for the future it would make it much easier to extend modes if parent-modes didn't trigger hooks for LSP. Maybe LSP initialization could be tied to a method instead of the mode hooks.

cxxxr:
I don't remember much about this anymore, so I can't really give any advice right away.
Please let me know when the merge is OK.

Fuglesteg:
No worries! It should be fine to merge now! 😊

vindarel:
aaand a doc page: https://lem-project.github.io/modes/vue/

PR #1848, opened 2025-07-31
Modifying file completion so that backspace will go up a directory

cianoc:
This changes file completion (e.g. for find-file) so that if the user's cursor is after '/', then if the user presses backspace, then instead of deleting this, it will instead go up a directory.

(In the following examples I'm using '|' as the cursor position)

E.g. 
/src/extension/file/|
[backspace]
/src/extension/|

/src/extension/file|
[backspace]
/src/extension/fil|

/|
[backspace]
/.

(The last example is to show that it does preserve existing behavior).

### DIFF
--- a/src/ext/prompt-window.lisp
+++ b/src/ext/prompt-window.lisp
@@ -397,7 +397,7 @@
           *special-paths*
           :initial-value path))
 
-
+
 (defun prompt-file-completion (string directory &key directory-only)
   (replace-prompt-input (normalize-path-input string))
   (flet ((move-to-file-start (point)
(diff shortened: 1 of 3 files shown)


cianoc:
I tried to write this so that it would work even if the user changed backspace, but due to how the completion-mode works, this would have required a lot of changes. So this is just on backspace, as it seems unlikely that many users will change this in the file-prompt-mode.

src/ext/prompt-window.lisp
```diff
@@ -466,3 +466,44 @@
 (setf *prompt-file-completion-function* 'prompt-file-completion)
 (setf *prompt-buffer-completion-function* 'prompt-buffer-completion)
 (setf *prompt-command-completion-function* 'prompt-command-completion)
+
+(defvar *file-prompt-keymap* (make-keymap :name '*file-mode-prompt-keymap*))
+(define-key *file-prompt-keymap* "Backspace" 'file-prompt-delete-previous-char)
+
+(define-command file-prompt-delete-previous-char (n) (:universal)
```
vindarel, line 473: Would you add a docstring?

vindarel:
I didn't try yet, but doesn't it break a current behavior?

- TAB-complete a directory, type BACKSPACE to remove the "/", add a few letters and a "/", Enter: Lem suggests to create this new directory. That's nice.

Did you know that Control-backspace works fine to erase words?

C-backspace isn't bound to anything… could it be used to get this behavior?

cianoc:
No that makes sense. I'll make those changes.

cianoc:
So now C-backspace is used to get this behavior, and so the current behavior should be preserved.

Though having slept on it, maybe if it's on C-backspace it would be better if the behavior was:
- If the cursor is after the 'forward slash' then it goes up a directory
- Otherwise the cursor goes back to the forward slash.

E.g.
/one/two/three/|
-> C-backspace
/one/two/|

/one/two/thre|
-> C-backspace
/one/two/

/one/two/three.html|
/one/two/

That feels a lot nicer. Thoughts @vindarel ?

cianoc:
Closing this, as I want to submit a cleaned up version of this.

PR #1849, opened 2025-08-01
Add a keybinding/command for file prompt that allows the user to jump to the parent folder

cianoc:
Continued from original pull request:
https://github.com/lem-project/lem/pull/1848

In the file prompt by clicking C-Backspace the user can jump to the parent folder.

Some clarifying examples '|' is the user cursor.
```
Find File: /folder1/folder2/|
[C-backspace]
Find File: /folder1/|
```

```
Find File: /folder1/folder2/index.html|
[C-backspace]
Find File: /folder1/folder2/|
```

This preserves the existing behavior, while adding useful additional behavior.

### DIFF
--- a/src/ext/prompt-window.lisp
+++ b/src/ext/prompt-window.lisp
@@ -397,7 +397,7 @@
           *special-paths*
           :initial-value path))
 
-
+
 (defun prompt-file-completion (string directory &key directory-only)
   (replace-prompt-input (normalize-path-input string))
   (flet ((move-to-file-start (point)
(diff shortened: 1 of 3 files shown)


vindarel:
LGTM! 

It worked on Lem's GUI (not ncurses because of keybinding being caugth by the terminal).

src/ext/prompt-window.lisp
```diff
@@ -466,3 +466,41 @@
...
+(define-key *file-prompt-keymap* "C-Backspace" 'file-prompt-parent-folder)
+
+(define-command file-prompt-parent-folder () ()
+  "In file prompt jump the parent folder and show the completion results for that folder."
+  (when (char= (character-at (current-point) -1) #\/)
+      (delete-previous-char))
+  (with-point ((end (current-point)))
+    (let ((point (search-backward (current-point) "/")))
+      (when point
+        (call-command 'forward-char nil)
```
cxxxr: It is better to avoid calling call-command unnecessarily.
This is because it increases the possibility of unexpected hooks being called.

If you just want to move the cursor one character, I think `(lem:character-offset point 1)` is good.

cxxxr: Other than that, I think it's good.
Thank you for the nice PR.

cianoc: Yes that makes sense. I'll try to remember that in future.

vindarel:
Great! Now if you want to document it… https://github.com/lem-project/lem-project.github.io/

PR #1851, opened 2025-08-05
Lem webview frontend

cxxxr:
## Usage
```
make webview
./lem
```

When separating the server and client
```
make server
make client

./lem-server --mode websocket --port 12345
./lem-client --url http://localhost:12345
```

## implementation
Lem can already be used from a browser.
https://github.com/lem-project/lem/pull/1571

This is achieved by launching a jsonrpc/websocket server from sbcl and communicating with a jsonrpc client written in JavaScript in the browser.

In this PR, by connecting to a locally launched server from webview, it is now possible to use lem as if it were a native GUI application.

In webview, this is possible by using `webview_navigate(‘http://localhost:xxxxx’)`.

### icon
Icon settings differ depending on each OS, so we need to handle them individually.

### JavaScript required :(
However, it is possible to create a framework for manipulating webview from Common Lisp and gradually replace it.

## TODO
- Setting icons for each OS
- Supports image display
- Other features supported by lem-sdl2
- macos
- windows


vindarel:
@cxxxr We can not run the ncurses front-end if we don't have the correct libwebview version. 

What about, in webview.lisp:

```lisp
(handler-case
    (use-foreign-library libwebview)
  (error (c)
    #-lem-ncurses
    (format *error-output* "Webview is not available: ~a" c)))
```

now I can run Lem again.

PR #1855, opened 2025-08-06
Persist M-x commands

vindarel:
I think it's good for review but I'd appreciate more tests by more people.

Now:

- by default, persist M-x commands on disk, or set `*persist-M-x-commands*` to NIL.
- 10 (default) of these commands are showed at the top of the M-x completion
 - they also show their keybindings, and they are correctly filtered on user input.

Now the core function `prompt-command-completion` accepts a key parameter `candidates`. If it is present, we build the completion items like the others and we show them first.

### DIFF
--- a/src/commands/other.lisp
+++ b/src/commands/other.lisp
@@ -1,6 +1,9 @@
 (defpackage :lem-core/commands/other
   (:use :cl :lem-core)
-  (:export :nop-command
+  (:export :*persist-M-x-commands*
+           :*history-limit*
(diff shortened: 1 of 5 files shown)


cianoc:
There seems to be duplication:
image

asm-mode is appearing in two places.

vindarel:
> There seems to be duplication

oh yeah, on it, thanks.

vindarel:
> There seems to be duplication

thanks a lot! I wouldn't have noticed the deprecation. Fixed.

PR #1859, opened 2025-08-07
Added shift arrow key selection

cianoc:
Added commands that allow the user to select regions using the shift and arrow keys.

If no mark is set, then shift + arrow-key will set the mark at point and move the cursor in the direction specified.
If the mark is set then it acts identically to the arrow key.

### DIFF
--- a/src/commands/edit.lisp
+++ b/src/commands/edit.lisp
@@ -32,7 +32,11 @@
            :transpose-characters
            :undo
            :redo
-           :delete-trailing-whitespace)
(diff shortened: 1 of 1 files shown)


vindarel:
Nice. Shouldn't they be bound to keys?

cianoc:
Key bindings are for cowards :)

I'll fix and I'll also add documentation for the commands.

cianoc:
We should be good now.

PR #1863, opened 2025-08-07
New Feature: toggle display borderless (no-titlebar).

li-yiyang:
Introduced
a new lem-interface protocol:
+ lem-if::display-borderless-p

a new lem-core function:
+ lem-core::display-borderless-p

a new lem command:
+ toggle-display-borderless

Tested on:
macOS Sequoia 15.5, SBCL 2.5.6, SDL2 build from source.

### DIFF
--- a/frontends/sdl2/main.lisp
+++ b/frontends/sdl2/main.lisp
@@ -499,3 +499,45 @@
 (defun char-width () (display:display-char-width (display:current-display)))
 (defun char-height () (display:display-char-height (display:current-display)))
 (defun current-renderer () (display:display-renderer (display:current-display)))
+
+(defmethod lem-if:display-fullscreen-p ((implementation sdl2))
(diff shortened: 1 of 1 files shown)


cianoc:
I'll try to test this on my mac today.

cxxxr:
Thank you for your PR.
However, I am currently considering this (https://github.com/lem-project/lem/issues/1867),
so I am unsure whether to merge it or not.
What do you think?

li-yiyang:
I think webview is a better choice than SDL. And I understand that changing the `lem-implementation` protocol may conflict with other implementation. 

So I think this PR should wait, or just be left as a user custom patch. 

If have to achieve borderless effects in webview, maybe we could refer to [webui frameless](https://github.com/webui-dev/webui/blob/main/examples/C/frameless/main.c) (I think there's some platform specific tricks to do so, not sure). I can help with mac version webview porting if I got more free time.

PR #1866, opened 2025-08-07
New `lem-core:exec-path` for PATH in macOS

li-yiyang:
Set PATH properly in macOS (Discussion #1864)


cianoc:
Nice! I'll try to test this today, but I may not get to it before the weekend.

PR #1870, opened 2025-08-07
change font api

cxxxr:
`lem-if:set-font-name` -> `lem-core:set-font-name`


ryukinix:
that makes more sense, thx  @cxxxr

PR #1878, opened 2025-08-09
fix nix package

mahmoodsh36:
installation through the nix flake was failing because of:
1. wrong hash output
2. old revision of `lem-extension-manager` which had references of the package `ql-http` and caused compilation issues.

this isnt a complete fix as now running `nix run github:lem-project/lem` gives `SDL Error: EGL not initialized`, but the compilation errors were fixed.


mahmoodsh36:
sems that the `SDL Error: EGL not initialized` error is an issue with wayland

running `DISPLAY=:0 nix run .#lem-sdl2` works. (well in this case its `DISPLAY=:0 nix run github:mahmoodsh36/lem#lem-sdl2`).

Issue #1869, opened 2025-08-07
vi-mode + M-x instant completion -> read-only


monkeyjunglejuice:
> Show the completion list instantly
> 
> The completion pop-up, like the one we get with M-x, wants us to press TAB before we get the completion candidates. If you want to open it right away, add this to your init file:
> 
> (add-hook *prompt-after-activate-hook*
>           (lambda ()
>             (call-command 'lem/prompt-window::prompt-completion nil)))
> 
> (add-hook *prompt-deactivate-hook*
>           (lambda ()
>             (lem/completion-mode:completion-end)))
> 
> You can start typing and the list will narrow down.

--> Doesn't work. Lem complains that the minibuffer is read-only (macOS, both in sdl2 and ncurses, showing the completion popup at the bottom). So there's no typing possible, selection via arrow keys only. Built from main branch.

cianoc:
I'm working on adding an option that would do completion candidates automatically (the add-hook thing is kind of a hack that doesn't work perfectly).

I'll look at this as part of that work and see if I can work out what the issue is.

Is this on the latest version?

monkeyjunglejuice:
No it's not from v2.3.0, but I built Lem just from the main branch yesterday. I've updated Lem (last commit a3e28c2) and built again, but the "problem" persists (well if that's just a hack anyway, it's hardly a "problem" then, but a matter of the manual). The hack doesn't work on v2.3.0 either btw., minibuffer stays read-only.

cianoc:
> The hack doesn't work on v2.3.0 either btw., minibuffer stays read-only.

Odd - it's been working fine for me on MacOS. Maybe there's something in your config?

Hopefully I should have a PR soon that will render that hack unnecessary.

monkeyjunglejuice:
Hm, not sure. The rest of my `init.lisp` is just that. Haven't tried using the "instant completion" hack without pinning the completion popup to the bottom yet. 
```common-lisp

(in-package :lem-user)

;; Start in vi-mode
(lem-vi-mode:vi-mode)

;; Start the Lisp REPL in vi insert mode.
(add-hook lem-lisp-mode:*lisp-repl-mode-hook* 'lem-vi-mode/commands:vi-insert)

;; Show completion popup at the bottom
(setf lem-core::*default-prompt-gravity* :bottom-display)
(setf lem/prompt-window::*prompt-completion-window-gravity* :horizontally-above-window)
(setf lem/prompt-window::*fill-width* t)
```

[...]

PR #1884, opened 2025-08-10
Configurable Automatic Completions

cianoc:
This adds a user configurable option:
`lem:prompt-window:*automatic-tab-completion*`

This is set to true by default (let me know if you want this set to false), and when set to true - the completion window opens automatically when the user opens a text prompt.

This also fixes issue: #1869

### DIFF
--- a/src/ext/prompt-window.lisp
+++ b/src/ext/prompt-window.lisp
@@ -14,7 +14,10 @@
 (defconstant +min-width+   10)
 (defconstant +min-height+  1)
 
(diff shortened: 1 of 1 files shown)


vindarel:
@cianoc well done, it's great! 

Did you document it ? :D https://github.com/lem-project/lem-project.github.io/

cianoc:
Not yet. It uncovered up some other problems that I'm trying to slowly fix. I'll document it when I'm happy with all of those fixes. I'll add an issue in the documentation repo.

PR #1886, opened 2025-08-10
Change Escape so that it closes non temporary overlays in the current buffer

cianoc:
I've got annoyed by overlays not always closing when I press escape.

This minor change fixes that by calling 'clear-overlays'.

Currently it only clears the overlays for the current buffer. I could change it so that it does it for all buffers.

I also wasn't sure if this was appropriate behavior, as maybe there are some overlays that should never be closed?


cxxxr:
Overlays should be closed by the owner, and closing them arbitrarily may cause malfunctions.

Issue #1441, opened 2024-07-15
fuzzy completion improvement


cxxxr:
- [ ] https://github.com/lem-project/lem/issues/926
- [ ] Highlight matchies

vindarel:
Still relevant. I've been updating https://github.com/vindarel/fuzzy-match/ for it to easily work with Lem.

It's working fine, I only have a display glitch at one point, in the completion pop-up.

vindarel:
@cxxxr In my experiments, I accept a string input with a space, but I get this little glitch when pressing TAB.

below, after "proj sw" TAB:

<img width="337" height="92" alt="Image" src="https://github.com/user-attachments/assets/98847097-0946-496d-aa93-5cc0c4e31fb4" />

my item completion candidates are `completion-item` objects. I don't have the glitch with completion candidates as strings.

I looked at existing functions in prompt-window.lisp: `replace-prompt-input`, `current-prompt-start-point`… but I couldn't debug this yet.

Does that raise memories?

cxxxr:
https://github.com/lem-project/lem/blob/60469dd8987e21c498c30463f6435d84259f5f9c/src/ext/completion-mode.lisp#L234
maybe, this one?
Whether the character is a symbol-char

jcs090218:
I've ported emacs [flx](https://github.com/lewang/flx) to [cl-flx](https://github.com/the-flx/cl-flx) a while ago. Would like to have this implemented to lem. :)

PR #1915, opened 2025-08-13
Improve fuzzy completion

vindarel:
WIP

The fuzzy-match algorithm has good results (see tests, not yet ported, in https://github.com/vindarel/fuzzy-match),

```lisp
  (lem/fuzzy-match:fuzzy-match "proj ws" (list of completion items) :key #'lem/completion-mode:completion-item-labe)
;; => "project-switch" item
```

the integration in Lem is not working properly, I have the visual glitch described in #1441 

"proj sw" (with space) ->

Selection_325

I didn't find why yet, despite your indication.

### DIFF
--- a/src/completion.lisp
+++ b/src/completion.lisp
@@ -28,28 +28,37 @@
               :while pos))))
 
 (defun completion (name elements &key (test #'search) separator key rank)
-  "Perform completion on ELEMENTS matching NAME. Returns matching elements, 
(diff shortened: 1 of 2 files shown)


cianoc:
https://github.com/the-flx/cl-flx

Would it be simpler to use this, as it is pretty battle tested as algorithms go.

PR #1916, opened 2025-08-13
[WIP] smart-parens-mode

daninus14:
not currently working, this is just a draft


vindarel:
(following Discord:) I pull the branch, start Lem, start the mode with `M-x smart-parens-mode` and typing a `(` inserts the pair. It works! (except it inserts too many closing parens)

daninus14:
Hm. `smart-parens-mode` is not showing up for me. What is your workflow? I was doing a `make` build after each change in the code. Are you running it from slime? How do I run it locally without building?

cxxxr:
I tried it, and make was successful.
The feature in this mode are also working.

cxxxr:
[Screencast from 2025-08-14 01-39-12.webm](https://github.com/user-attachments/assets/3d7a11f7-de7e-44ca-a3e4-c940ef3d6854)

extensions/smart-parens-mode/smart-parens-mode.lisp
```diff
@@ -0,0 +1,99 @@
...
+
+(in-package :smart-parens-mode)
+
+(define-minor-mode smart-parens-mode
+    (:name "smart-parens-mode"
+     :keymap *smart-parens-keymap*))
+
+;;; From paredit
+(defun bolp (point)
+  (zerop (point-charpos point)))
```
cxxxr: It can use `(start-line-p point)` instead.

extensions/smart-parens-mode/smart-parens-mode.lisp
```diff
@@ -0,0 +1,99 @@
...
+
+(defun eolp (point)
```
cxxxr: It can use `(end-line-p point)` instead.

extensions/smart-parens-mode/smart-parens-mode.lisp
```diff
@@ -0,0 +1,99 @@
...
+  (editor-insert-pair #\" #\"))
+
+(define-command smart-parens-insert-single-quote () ()
+  (editor-insert-pair #\' #\'))
+
+(define-key *smart-parens-keymap* "\"" 'smart-parens-insert-double-quote)
+(define-key *smart-parens-keymap* "'" 'smart-parens-insert-single-quote)
+(define-key *smart-parens-keymap* "(" 'smart-parens-insert-paren)
+(define-key *smart-parens-keymap* "[" 'smart-parens-insert-bracket)
+(define-key *smart-parens-keymap* "{" 'smart-parens-insert-brace)
```
cxxxr: One problem with this implementation is that it implicitly relies on C-like syntax.
It would be better to refer to the syntax-table for each mode.

for example
```
CL-USER> (lem:syntax-open-paren-char-p #\()
(#\( . #\))
CL-USER> (lem:syntax-open-paren-char-p #\[)
(#\[ . #\])
CL-USER> (lem:syntax-open-paren-char-p #\{)
(#\{ . #\})
```

extensions/smart-parens-mode/smart-parens-mode.lisp
```diff
@@ -0,0 +1,99 @@
+(defpackage :smart-parens-mode
+  (:use :cl :lem)
+  (:export :smart-parens-mode))
+
+(in-package :smart-parens-mode)
```
cxxxr: The package name may conflict with others.
We use the prefix “lem-”.

extensions/smart-parens-mode/smart-parens-mode.lisp
```diff
@@ -0,0 +1,99 @@
+(defpackage :smart-parens-mode
+  (:use :cl :lem)
+  (:export :smart-parens-mode))
+
```
cxxxr: Personally, I think it is better not to have a space between defpackage and in-package.

extensions/smart-parens-mode/smart-parens-mode.lisp
daninus14: Hm, do you mean that instead of using `define-key` I should use some function which gets the currently inserted character like this:

```common-lisp
(defun close-paren (given-char)
  (when (lem:syntax-open-paren-char-p given-char)
    (apply #'editor-insert-pair (lem:syntax-open-paren-char-p given-char))))
```

That seems like it would only work for the parens and not the `"` and `'` quote characters.

I don't know lem well enough. What function or macro would I pass this `close-paren` function to?

extensions/smart-parens-mode/smart-parens-mode.lisp
daninus14: just did it

extensions/smart-parens-mode/smart-parens-mode.lisp
daninus14: ok

extensions/smart-parens-mode/smart-parens-mode.lisp
daninus14: just did it

extensions/smart-parens-mode/smart-parens-mode.lisp
daninus14: ok

extensions/smart-parens-mode/smart-parens-mode.lisp
cxxxr: https://github.com/lem-project/lem/blob/4072439a3bfeee50f8a637a99bc8429f651e3692/extensions/c-mode/c-mode.lisp#L13

syntax-table is defined here in C, for example.

cxxxr: for example, like this
```
(defmethod execute ((mode smart-parens-mode) (command self-insert) argument)
  (let ((c (get-self-insert-char)))
    (alexandria:when-let (pair (syntax-open-paren-char-p c))
      (editor-insert-pair (car pair) (cdr pair))
      (return-from execute))
    (when (syntax-string-quote-char-p c)
      (editor-insert-pair c c)
      (return-from execute))
    (call-next-method)))
```

daninus14:
Ok, I realize I doubt I'll ever get to change this code. This is a stylistic issue for an optional mode, where the code works. It's based on code of existing modes. While I understand it's better to have your desired style in the code base, considering this is not even part of the main code base and it's an additional mode, following the style of other modes, I think it's reasonable to ask to accept the PR.

PR #1924, opened 2025-08-15
Synchronize the position and size of the prompt-window and completion-window

cxxxr:
image


vindarel:
I don't really get what this improves but it's a beautiful screenshot, thanks.

cxxxr:
https://github.com/lem-project/lem/pull/1924/commits/d0314c7d7d87656682eeacaf47b737f18e647de8
When the completion window is displayed, the position of the prompt window is now calculated including the completion window.
This is because even when center was specified, the display was shifted to the right when the completion window was displayed.

PR #1927, opened 2025-08-18
Add an attached buffer and window

cxxxr:
This feature attaches “attached-window” to the bottom of a specific window.
attached-window has attached-buffer.
When you attach an attached-window to a specific window, that window becomes the parent window from the perspective of the attached-window.
The attached window remains visible even when the parent window is scrolled.
It's like a comment input form on Slack or Discord.


vindarel:
great, don't hesitate to add a screenshot and I'll report it into Lem's website.

cxxxr:
@vindarel 
image

I made this for claude code, so it doesn't seem to make sense on its own.

vindarel:
https://lem-project.github.io/development/displaying/#attached-buffers

PR #1939, opened 2025-08-23
Refactor

jean0t:
changing the backticks to (list ...) just as instructed in the style guide
in the file src/lem.lisp

### DIFF
--- a/src/lem.lisp
+++ b/src/lem.lisp
@@ -90,10 +90,10 @@
 
 (defun initialize-source-registry ()
   (asdf:initialize-source-registry
-   `(:source-registry
+   (list :source-registry
      :inherit-configuration
      (:also-exclude ".qlot")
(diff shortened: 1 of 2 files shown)


src/lem.lisp
```diff
@@ -90,10 +90,10 @@
 
 (defun initialize-source-registry ()
   (asdf:initialize-source-registry
-   `(:source-registry
+   (list :source-registry
      :inherit-configuration
      (:also-exclude ".qlot")
-     (:tree ,(asdf:system-source-directory :lem)))))
```
cxxxr, line 96: This should cause an error.
Please build lem and try again.

Backquotes might be better here.

@vindarel The text in that section of STYLEGUIDE.md was probably correct in its original context, but it does not seem to be universally correct.

PR #1942, opened 2025-08-26
Automatic prompt completion with flex matching, history and refactored interface.

cianoc:
This PR does the following:
- Automatic completion
  - The default is always at the top if set (so kill buffer will work as expected, with the current buffer the default).
  - Historical items are shown below default and above rest of completions.
- Flex matching by default using cl-flx
- Changes to key commands
  - Tab inserts selected item into prompt, unless there are multiple partial matches, in which case the current behavior is preserved (e.g. if there are two matches 'fred-flex' and 'fred-exact' - then 'fred' will be inserted into prompt, and a subsequent tab will insert the selected item).
  - Return selects and executes currently highlighted item.
  - control-backspace clears prompt (unless this is overridden. For file/directory the user will be taken up a folder).

The reason for the key-command changes is that this makes the prompt a little bit more efficient, while still supporting expected functionality. This means that we lose the ability to go up/down candidates with tab - but I think this change is worth it. Instead for most uses of the prompt you can select and execute with one keystroke, while for 'find-file', you can quickly go up folders with tab, down with command-backspace and select a directory with 

Refactoring changes are still in progress (hence draft status), but when finished:
- separation of candidates from matching filter. This means that when adding new prompt interfaces, you only need to worry about how to generate candidates.
- default and history are (optionally) passed in, and prompt-window (in open-prompt function) handles sorting, history, default, etc for you. This greatly simplifies adding new prompt completion functions and reduces the amount of code required.


cianoc:
What still needs to be done:
- [ ] Moving all prompt functions that use prompt-aux to this interface, and cleaning up prompt functions so they are fully refactored.
  - In particular removing 'candidate-function from the prompt-line, prompt-aux functions, etc.
- [x] Adding code so description is part of completion results (e.g. keybindings for commands).
- [ ] Comments...
- [ ] More testing
- [ ] Cleaning up user config of matching, so it is fully configurable
- [ ] General clean up
- [ ] Thorough testing.

For review - this is a big change, and it's possible some of these ux changes won't be quite right. Also the refactoring might be ugly. I suggest people try it while I finish these changes, and respond to any feedback on the code.

Currently the new code only works properly with the prompt for buffer and command (M-x and C-x b). All other prompt commands use the existing code.

src/ext/prompt-window.lisp
```diff
@@ -161,11 +177,55 @@
...
+                             (append
+                              (funcall filter-fn (when default (list default)) prompt-string)
+                              (loop for item in (funcall filter-fn history prompt-string)
+                                    collect item)
+                              (loop for item in (funcall filter-fn candidates prompt-string :sort t) 
+                                    collect item))
+                             :test #'equal
+                             :from-end t)))
+                (loop for item in items
+                      collect (lem/completion-mode:make-completion-item :label (car item)o
```
cxxxr: `(car item)o`

src/ext/popup-menu.lisp
```diff
@@ -1,4 +1,4 @@
-(defpackage :lem/popup-menu
+q(defpackage :lem/popup-menu
```
cxxxr: "q"

cianoc: Weird that it compiled with both of these typos. Fixed now.

cianoc:
So having reflected a bit more on tab completion, I think there should be two styles. The current style (where return inserts the selection in the prompt, and tab moves down) and a new style (where return executes the selection in the prompt, and tab inserts the selection in the prompt). Not sure what the right name for the variable should be. I'll see what the various Emacs libraries do here.

cianoc:
Added a flag *prompt-tab-completion*, which is true by default.

When true, tab and return work identically to today. When set to false, tab inserts in the prompt and return will execute the highly completion selection.

cianoc:
Added optional annotation function. Reused existing code for commands, and added annotations for buffers.

src/completion.lisp
```diff
@@ -2,6 +2,21 @@
...
+      (mapcar (lambda (x) (cons (car x) 
+                                (mapcar 
+                                 (lambda (pos) (cons pos (+ pos 1)))
+                                 (cddr x))))
+              (let ((results (loop for candidate in candidates
+                                   for score = (flx:score candidate elt) 
+                                   when (car score)
+                                   collect (cons candidate score))))
+                (if sort 
+                    (sort results (lambda (x y) (> (cadr x) (cadr y))))
```
vindarel: `(sort results #'> :key #'cadr)` :)   

I'm also in favor in using `#'second`.

qlfile
```diff
@@ -10,3 +10,4 @@ git cl-sdl2-image https://github.com/lem-project/cl-sdl2-image.git
 git jsonrpc https://github.com/cxxxr/jsonrpc.git
 git lem-extension-manager https://github.com/lem-project/lem-extension-manager.git
 git webview https://github.com/lem-project/webview.git
+git cl-flx https://github.com/the-flx/cl-flx.git
```
vindarel, line 13: For such a central feature, I strongly support to either copy and include cl-flx sources into Lem's codebase, or fork it into lem-project/cl-flx.

cianoc: I wondered about that. I would rather fork it into the code base, as I can imagine wanting to change it over time in ways that better fit Lem.

src/completion.lisp
cianoc: Oh that's a neat trick.

qlfile
vindarel: However I am not a fan of cl-flx codebase… too big functions. Not enough tests. It seems to define Elisp-specific functions (memq? capital-p? bigger-sublis doesn't use a :key function, etc). I'm not the maintainer but I looked at these completion features as a Lem developer so you hear me : )

cianoc: Where do you think it should go. It feels a little specialized for src/ext - but maybe I'm overthinking it.

cianoc: > However I am not a fan of cl-flx codebase… too big functions. Not enough tests. It seems to define Elisp-specific functions (memq? capital-p? bigger-sublis doesn't use a :key function, etc). I'm not the maintainer but I looked at these completion features as a Lem developer so you hear me : )

Yes that's fine. I would want to rewrite it first anyway if we were to bring it in.

cianoc: And if for whatever reason people don't want to use that algorithm, I'm fine with that.

cianoc:
I'm closing this as I'm going to break this into multiple PRs.

PR #1949, opened 2025-09-02
build legit binary

vindarel:
Now, we can call a `legit` binary from anywhere in the terminal.

Note: I tried creating an alias but `lem --eval '(lem/legit::legit-status)'` fails, "the value NIL is not of type NUMBER", "directory does not exist: … /lem/(lem/. Create?" etc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,13 @@ webview:
 
 lem: sdl2
 
+.PHONY: legit
+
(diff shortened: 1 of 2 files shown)


Sasanidas:
LGTM, legit is a very nice tool to use indeed :rocket:

PR #1951, opened 2025-09-03
Documentation fixes and revision

tautologico:
The main motivation for this PR is to fix the documentation for `line-offset` in `src/buffer/internal/basic.lisp`; the previous documentation was wrong. I then took the opportunity to revise the documentation of a few other functions in the same file, trying to keep the style consistent and clarifying the descriptions.


vindarel:
thanks!

Issue #1953, opened 2025-09-04
Lem 404 on MacOS


blackeuler:
<img width="1022" height="765" alt="Image" src="https://github.com/user-attachments/assets/07b92c46-1068-4f35-9828-065893d1583d" />

I believe the issue stems from 

`frontends/server/main.lisp`
```
(defun clack-handler (env)
  (unless (wsd:websocket-p env)
    (let ((path (getf env :path-info)))
      (cond ((string= "/" path)
             `(200 (:content-type "text/html")
                   ,(find-dist-by-path "frontend/dist/index.html")))
```

But the build has this structure, doesnt contain the needed frontend assets that are being looked for by the server
```
lem.app
└── Contents
    ├── Info.plist
    ├── MacOS
    │   ├── lem
    │   ├── libasyncprocess.so
    │   ├── libcrypto.3.dylib
    │   ├── libcrypto.dylib
    │   ├── libssl.3.dylib
    │   ├── libssl.dylib
    │   ├── libwebview.dylib
    │   ├── libzstd.1.5.7.dylib
    │   └── terminal.so
    └── Resources
        └── lem.png
```

vindarel:
Hello, if I followed Discord correctly you now have another analysis?

blackeuler:
yes the real issue was not in the assets not being in the build but this function
```
(defun find-dist-by-path (path)
  (asdf:system-relative-pathname :lem-server path)
  #+(or)
  (cdr (assoc path *dist* :test #'equal)))
```
the first line will use asdf to look for the actual path instead of relying on the defvar that builds with the image.

tautologico:
I believe this was closed by #1954, but I don't have a recent mac to test it.

blackeuler:
I was going to wait until the nightly release builds to test it and then I'll close out this issue. I was able to get the build script to run but just want to be sure that this fully addresses the problem from the release

blackeuler:
Closing this as it has been resolved by #1954 just tested it

PR #1954, opened 2025-09-04
remove the asdf check from find dist path

blackeuler:
#1953 
I am a little new to lisp images and lisp in general or I should say it's been a while. 
What I believe was happening here is at runtime, the get-dist-by-path function was using asdf to look for needed asset files for the server. As a result of asdf not being available to the runtime it would never find it thus resulting in the 404. I tested this locally and was able to get the editor working again on my Mac Silicon. 

Would love some eyes on this so that others can just get started by using the binary. Let me know if there is anything I need to change or a reason why this is a bad change. Still learning .

### DIFF
--- a/frontends/server/main.lisp
+++ b/frontends/server/main.lisp
@@ -48,8 +48,6 @@
 (defvar *dist* (dump-files))
 
 (defun find-dist-by-path (path)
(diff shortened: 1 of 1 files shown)


cxxxr:
Ah, it's my debugging code. Thank you.

PR #1955, opened 2025-09-09
Create the default C-x keymap.

sethaldini:
Creates the lem-core:*global-prefix-keymap* and binds to it "C-x". This allows users to rebind the keymap to a different prefix. New "C-x" bindings in the lem-core:*global-keymap* should be bound in this new keymap instead.


vindarel:
That sounds great, thank you.

Does that break backward compatibility with existing "C-x …" bindings to the global keymap in one's init file? (I guess not?) 

Can you also send a PR to the documentation on https://github.com/lem-project/lem-project.github.io/ too ?

Thanks!

sethaldini:
After doing some more testing there is an issue if there is an "C-x" binding in the init file. I'll see if I can get that resolved and then I'll update the documentation too.

vindarel:
if you threw a good, understantable error message that would help users migrate, that would be good too.

if both options are too difficult, alright.

vindarel:
hey @sethaldini , what's up with this good PR?

cxxxr:
I think this PR is great, but is there a reason it can't be merged?

vindarel:
I tried the PR and it works fine. I have no issue with my init file where I define C-x bindings in `*global-keymap*`.

vindarel:
I tried it more: my custom keybindings work fine, but I can't use `C-x C-f` and other built-in keybindings : S

I'd like to re-open this PR but we can't.

vindarel:
and so, this is the issue with existing C-x bindings in one's init file. If I don't load my init file, everything is OK.

If I have one define-key such as

    (define-key *global-keymap* "C-x i" 'imenu)

in my init file, then I can't use the default C-x keybindings.

PR #1958, opened 2025-09-11
Update lem-extension manager version

Sasanidas:
Change the version to reflect the changes on the lem-extension-manager package (https://github.com/lem-project/lem-extension-manager)


vindarel:
Hello, nice to hear from the author. Did you happen to see https://github.com/lem-project/lem-extension-manager/issues/5 ? We get an error when trying to use the package manager with organ-mode.

PR #1965, opened 2025-09-22
Fixed bugs in lem/link package and added link properties in Markdown Mode

ethanxxxl:
I tried to open a link using the `link-open` command in markdown mode, but the links did not open.
After some investigation I found a few issues:
1. markdown mode was not setting a `:link` property on the highlighted link text
2. the `lem/link::link-at` function was searching for `'link` and not `:link`
3. in `lem/link::search-url-link` and `lem/link::search-file-link`, link searches were not obeying the LIMIT parameter.  I added a check to ensure that the returned URL is bounded by LIMIT.

I fixed these issues and added code to the Markdown Mode syntax parser to add the `:link` text properties.
Lem Markdown mode adds the correct text properties to links, so they can be opened by the lem/link package.

### DIFF
--- a/extensions/markdown-mode/syntax-parser.lisp
+++ b/extensions/markdown-mode/syntax-parser.lisp
@@ -133,32 +133,23 @@
 
                     (move-to-column match-start start)
                     (move-to-column match-end end)
-
                     (put-text-property match-start match-end
                                        :attribute 'document-link-attribute)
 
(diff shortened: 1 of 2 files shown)


ethanxxxl:
I overlooked this symbol name change, it's fixed now.

vindarel:
It works now,

thanks a lot!

PR #1971, opened 2025-10-04
Fix `right-window`: Find windows where `left.x + left.width == right.x`

### LINKED ISSUE #1970: Webview – `window-move-right` not working.
In lem-webview, when the program is split horizontally into two windows, running `window-move-right` from the left window does nothing.
With 3 horizontal windows, running it from the left window skips to the right window.
The other three directions seem to work normally.
This doesn't happen on SDL2.

I believe this difference is caused by JsonRPC setting `:window-left-margin 0` on the class `jsonrpc` in `frontends/server/main.lisp`.
This means that `window-width` of the left window is equal to `window-x` of the right window.
So, `lem-core::right-window` doesn't find it.

The left-margin was set in edeb05a42, I think because setting it to 1 causes a gap between modelines.

markmurphydev:
Fixes #1970.
I tested the movement manually. Everything seemed to work normally.
It looks like `lem-core::right-window` is only called by `lem-core/commands/window::move-right`, so it shouldn't mess with anything else.
I think it's logically fine that window `{ left: 0, width: 100, ... }` has window `{ left: 100, ... }` to its right.

### DIFF
--- a/src/window/window.lisp
+++ b/src/window/window.lisp
@@ -999,7 +999,7 @@ You can pass in the optional argument WINDOW-LIST to replace the default
   (unless (floating-window-p window)
     (first (sort (min-if #'window-x
                          (remove-if-not (lambda (w)
-                                          (> (window-x w)
+                                          (>= (window-x w)
                                              (+ (window-x window)
                                                 (window-width window))))
                                         (window-list)))


vindarel:
Thank you!

PR #1972, opened 2025-10-12
Add English translations for comments in `extensions/`.

markmurphydev:
I translated all the JP comments I could find under `extensions/`.
I tried to check these translations carefully, but I'm not an expert on these packages, or totally fluent in technical Japanese. If anyone's able to double-check these, I'd be grateful.


cxxxr:
I think it's a good change.
You can delete the Japanese version if you like.

markmurphydev:
I went ahead and deleted them.

PR #1974, opened 2025-10-20
Add a simple python lsp config using python-lsp-server

Sasanidas:
So, this is a simple file similar to those in `js-mode` and `ts-mode` that add lsp configuartion to the python mode.


vindarel:
cool, thanks.

Issue #1989, opened 2025-11-05
Add webview-ncurses make target


ryan-carpenter:
The webview interface should have a webview-ncurses make target (See https://github.com/lem-project/lem/discussions/1980#discussioncomment-14824320).

- [ ] Add webview-ncurses to `lem/Makefile`
- [ ] Create `lem/scripts/build-webview-ncurses.lisp`
- [ ] Create `lem/scripts/install/lem-webview.desktop` and `lem-webview-ncurses.desktop`
- [ ] Update docs (See https://github.com/lem-project/lem-project.github.io/issues/63)

PR #1992, opened 2025-11-06
Add make webview-ncurses

ryan-carpenter:
Fix https://github.com/lem-project/lem/issues/1989. Changes default variant from sdl2 to webview and adds `webview` and `webview-ncurses` scripts/.desktop files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,15 +25,16 @@ With Docker (terminal version):
 
 ## Build
 
-### Lem(terminal version)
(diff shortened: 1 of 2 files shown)


ryan-carpenter:
This pull request is a best effort from a non-developer, so it may be incomplete or incorrect. I did not use AI. Following the precedent from sdl2 and sdl2-ncurses, I made separate-but-identical desktop files for webview and webview-ncurses. I added two installation scripts, but did not make any changes related to system-wide installation.

PR #1997, opened 2025-11-14
Fix completion popup not closing, Improve lsp completion

### LINKED ISSUE #1751: Completions Selection List Stuck Open
Somehow I made Lem forget it has this open

Image

and I don't know how to get rid of it without restarting.

Selecting an item produces this result:

Image

Fuglesteg:
Fix #1751

- Sort and filter completions from LSP using `completion-strings`
- Fix completion window popup not closing on async completions
- Add spinner on async completions to give instant feedback to the user

Before:

lem-completion-before

After:

lem-completion-after

### DIFF
--- a/src/ext/completion-mode.lisp
+++ b/src/ext/completion-mode.lisp
@@ -206,11 +206,13 @@
   (completion-refresh))
 
 (define-command completion-next-line () ()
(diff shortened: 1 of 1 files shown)


cxxxr:
Very good!
Thank you!

PR #2002, opened 2025-11-21
Improve popup menu with title for list-buffers

Fuglesteg:
This change improves the popup menu when used with a header, like in the `list-buffers` menu.
Adds keybinds to make it usable in vi-mode,
Improve look when header is active in popup menu and there are no items, so the header is still visible

- Fix list-buffers in vi-mode by adding keybinds to multi-column-list
- Fix focus-overlay in popup-window when only header

Before:

lem-buffer-list-before

After:

lem-buffer-list-after

### DIFF
--- a/src/ext/popup-menu.lisp
+++ b/src/ext/popup-menu.lisp
@@ -35,6 +35,9 @@
   (make-line-overlay point focus-attribute))
 
 (defun update-focus-overlay (popup-menu point)
++  "Refresh the focus highlight so it tracks POINT in POPUP-MENU.
++Deletes any previous overlay, clears stray overlays, and creates a new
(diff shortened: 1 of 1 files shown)


src/ext/popup-menu.lisp
```diff
@@ -35,9 +35,12 @@
   (make-line-overlay point focus-attribute))
 
 (defun update-focus-overlay (popup-menu point)
```
cxxxr, line 37:
```suggestion
(defun update-focus-overlay (popup-menu point)
+  "Refresh the focus highlight so it tracks POINT in POPUP-MENU.
+Deletes any previous overlay, clears stray overlays, and creates a new
+focus overlay unless POINT is on the header line."
```

PR #2006, opened 2025-11-24
Adds a new bookmark package.

mychris:
As requested, adds my bookmark package to core.

See the description of the package for more information.

For now, I decided to not name the package `lem-bookmark`. It seems this is the naming convention for internal packages, but I didn't want to break users of the package, which might already have it configured. If you would like me to change it, please tell me, I have no strong feelings about it. If you do change it yourself, please also update the documentation, which points to `(find-package "BOOKMARK")`.

I decided to not load/save bookmarks automatically. It should be easy enough for a user to add a hook to their config if they wish to do so:
```
(add-hook *after-init-hook* (lambda ()
                              (bookmark:bookmark-load)))
```
Not sure if there is a hook for shutdown? I personally would prefer bookmarks not to be saved automatically, since I usually also add temporary bookmarks while working with a bigger project and wouldn't want them to be saved automatically. If you want to change this behaviour in the future, please think about adding a configuration variable for saving.

I didn't test the integration thoroughly. Building lem from main worked for me, bookmarks got loaded and I could jump to them.

I don't care about ownership and gladly give up all rights so this package can be easily contributed to the core.

### DIFF
--- a/extensions/bookmark/bookmark.lisp
+++ b/extensions/bookmark/bookmark.lisp
@@ -2,7 +2,7 @@
 ;;;; Bookmark package for lem
 ;;;;
 
-(defpackage :bookmark
+(defpackage :lem-bookmark
   (:use :cl :lem)
   (:export
    ;; customization variables
@@ -30,7 +30,7 @@
    :bookmark-relocate
    :bookmark-jump))
 
-(in-package :bookmark)
+(in-package :lem-bookmark)
 
 (setf (documentation *package* t)
       "Bookmarks for the lem editor.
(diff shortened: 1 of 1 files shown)


vindarel:
This is great, thank you! 

minor feedback, what if delete-all prompted for user y/n confirmation? "Do you want to delete [n] bookmarks?"

LGTM and I will merge anyways.

mychris:
Sure thing, added.
What do you think about the package name?
On a second thought, maybe it is not a bad idea to change the package name to `lem-bookmark` now. Existing users already need to change their init file, because they need to remove the existing dependency. And I also think there aren't that many users right now. Better than changing it later when it's in core and there are way more users.
What do you think?

vindarel:
yes, I think it's good to be consistent in naming and rename it now.

mychris:
Done. Let me know if there is anything else I should do right now.

vindarel:
Thank you!

PR #2009, opened 2025-11-28
fix: Dockerfile now includes webkit for webview

malbertzard:
The Dockerfile did not build on my nixos machine. After looking at it for a bit it seems like due to webview being the default frontend there needs to be the libs for webview available even on the ncurses frontend used in docker. After testing this seems to work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ WORKDIR /app
 
 COPY . .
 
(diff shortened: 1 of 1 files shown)


sakurawald:
I download the binary from https://github.com/lem-project/lem/releases/download/nightly-latest/Lem-x86_64.AppImage

It seems like this issue is not fixed in my env.
image

PR #2019, opened 2025-12-09
qlot-free nix setup

tani:
Lem when used with Nix relies heavily on Nix's Common Lisp library collection. We have no dependency on qlot.

To launch Lem, you can simply run `nix run .#lem`. If you want to use the SDL2 version, append the flag: `nix run .#lem -- -i sdl2`.

Installation is also possible through `lem.${system}.packages.lem` or by setting up an [overlay](https://flake.parts/options/flake-parts-easyoverlay#warning).

Remember to run `nvfetcher` followed by `nix flake update` to keep your dependencies up to date.


psionic-k:
Hello, I'm just curious and not a reviewer.  IIRC the package built with nix could not load extensions in the past.  Has that situation changed?

cxxxr:
Thank you so much!

tani:
> Hello, I'm just curious and not a reviewer. IIRC the package built with nix could not load extensions in the past. Has that situation changed?

Yes, we can now load extensions.
​By using the function buildASDSystem to compile the extension into a Nix package, and then adding it to lispLibs using lem.overrideLispAttrs, users should be able to add extensions.
​I am still a light user of Lem, so I haven't added any extensions yet, but when I get the chance to do so, I will come back here to remember this process and write documentation.

PR #2035, opened 2025-12-16
CI for Nix

tani:
This enables CI testing for the Nix recipe.
GitHub Actions test `lem-{ncurses,webview,sdl2}` on macOS and Linux with two archtectures `arm64` and `x86_64`. Note that this test fails currently because `lem-sdl2` especially SDL2 for Common Lisp does not work on macOS. If you wishes to drop `lem-sdl2` on macOS, I'll update script to skip test for this issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,72 @@
+name: Nix CI
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  # Trigger manually to optionally include heavier/unstable targets
+  workflow_dispatch:
+    inputs:
+      include_optional:
+        description: 'Include optional targets (SDL2, Intel Mac)'
+        type: boolean
+        default: false
+
+jobs:
+  # 1. Job to generate the matrix dynamically based on inputs
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      os: ${{ steps.set-matrix.outputs.os }}
+      backend: ${{ steps.set-matrix.outputs.backend }}
+    steps:
+      - name: Generate Matrix JSON
+        id: set-matrix
+        uses: actions/github-script@v7
+        with:
+          script: |
+            // Default configuration (Stable & Lightweight targets)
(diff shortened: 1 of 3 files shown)


cxxxr:
That's great. I feel like we can replace everything related to lem with nix.
For actions that are currently failing, how about triggering them explicitly with the Run workflow button?

tani:
@cxxxr 

When running CI with the following combinations, both `lem-sdl2` and `macos-15-intel` often fail. Furthermore, both lem-sdl2 and the Intel version of macOS are nearing deprecation. Therefore, I propose removing them from CI. At that time, I would like to apply the method you previously suggested of selectively running only deprecated combinations. Could you please tell me how to do that?

      matrix:
        os:
          - ubuntu-24.04
          - ubuntu-24.04-arm
          - macos-15
          - macos-15-intel
        backend:
          - lem-ncurses
          - lem-sdl2
          - lem-webview

cxxxr:
@tani 
This method uses workflow_dispatch to build items with a checkbox (which evaluates to true via variable reference).
```
  on:
    push:
      branches:
        - main
    pull_request:
    workflow_dispatch:
      inputs:
        # OS
        ubuntu_x64:
          description: 'Ubuntu x64'
          type: boolean
          default: true
        ubuntu_arm:
          description: 'Ubuntu ARM'
          type: boolean
          default: true
        macos_arm:
          description: 'macOS ARM (Apple Silicon)'
          type: boolean
          default: true
        macos_intel:
          description: '[Deprecated] macOS Intel'
          type: boolean
          default: false
        # Backend
        lem_ncurses:
          description: 'lem-ncurses'
          type: boolean
          default: true
        lem_webview:
          description: 'lem-webview'
          type: boolean
          default: true
        lem_sdl2:
          description: '[Deprecated] lem-sdl2'
          type: boolean
          default: false

```


Since it's just one suggestion, if it unnecessarily complicates things, I think it's fine to comment it out.

tani:
@cxxxr Gemini taught me the magical script. Basically it is the same idea of you. Please check it out 😉

cxxxr:
I think it's good. Thanks. I'll merge it.

PR #2074, opened 2026-01-02
fix -i command line argument

ethanxxxl:
Webview was using it's own main function that didn't have logic for selecting the interface based off the -i argument.  I changed the build scripts to use `lem:main` in webview, and added some CLOS logic logic to ensure that the lem-webview setup code is run when using the webview interface.

I also made some changes to `lem-core::get-default-implementation`.  It falls back to webview, ncurses, or sdl2 if it can't find whatever the user tried to select.

### DIFF
--- a/lem.asd
+++ b/lem.asd
@@ -21,7 +21,6 @@
                "alexandria"
                "trivial-gray-streams"
                "trivial-types"
-               "trivial-backtrace"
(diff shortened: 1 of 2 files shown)


vindarel:
Thank you and well done :+1: :+1:

PR #2078, opened 2026-01-07
feat(input): add meta-prefix-keys for flexible Meta input

cxxxr:
## Summary
- Add `meta-prefix-keys` editor variable to treat keys as Meta prefix without timeout
- When enabled, pressing a prefix key followed by another key produces Meta+key immediately
- Pressing the same prefix key twice produces the prefix key itself
- Implemented in core layer, works across all frontends

## Supported configurations
```lisp
;; Use Escape as Meta prefix (simplest)
(setf (lem:variable-value 'lem:meta-prefix-keys :global) t)

;; Use a specific key as Meta prefix
(setf (lem:variable-value 'lem:meta-prefix-keys :global)
      (lem:make-key :ctrl t :sym ";"))

;; Use multiple keys as Meta prefix
(setf (lem:variable-value 'lem:meta-prefix-keys :global)
      (list (lem:make-key :sym "Escape")
            (lem:make-key :ctrl t :sym ";")))
```

## Behavior
| Input | Output |
|-------|--------|
| Escape x | M-x |
| Escape Escape | Escape |
| Escape C-c | M-C-c |

## Test plan
- [x] Unit tests for keys-equal-p, meta-prefix-key-p, add-meta-modifier, maybe-convert-to-meta
- [x] Test: `t` (Escape) as prefix
- [x] Test: Custom key as prefix  
- [x] Test: Multiple prefix keys
- [x] Test: Double-press produces the prefix key itself
- [x] Test: Non-prefix keys pass through unchanged
- [x] All 12 tests passing

### DIFF
--- a/src/input.lisp
+++ b/src/input.lisp
@@ -1,9 +1,21 @@
 (in-package :lem-core)
 
-(define-editor-variable escape-as-meta-prefix nil
-  "When non-nil, treat Escape key as Meta prefix without timeout.
-Pressing Escape followed by another key produces Meta+key.
-Escape Escape produces a single Escape.")
+(define-editor-variable meta-prefix-keys nil
+  "Keys to treat as Meta prefix without timeout.
+When non-nil, pressing a prefix key followed by another key produces Meta+key.
+Pressing the prefix key twice produces the prefix key itself.
+
+Value can be:
+- nil: disabled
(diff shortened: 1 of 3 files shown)


cxxxr:
@code-contractor-app

PR #2092, opened 2026-01-11
added doc strings to popup menu functions

ethanxxxl:
I was tinkering with the popup menu code.  It took me a while to understand what was going on, so I felt it might help someone down the line if I added some documentation here.

There also seems to be some git artifacts, presumably from a merge.  I went ahead and removed them.

### DIFF
--- a/extensions/process/process.lisp
+++ b/extensions/process/process.lisp
@@ -25,7 +25,7 @@
     :reader process-output-callback-type)))
 
 (defun run-process (command &key name output-callback output-callback-type directory)
-  "Starts runs an external program as a separate thread.
+  "Runs an external program as a separate thread.
 - `command`: can simply be a string naming the command to run.  If you need to
   pass arguments to the command, then `command` must be a list of strings.  Each string will be its own
   arg.  These args are not parsed by a shell, so you cannot combine tokens, such as a flag and its value.


vindarel:
thanks! Always good.

Do we leave the "TODO"??

vindarel:
BTW I am not familiar to the "::" syntax, I don't think it is used elsewhere, just use bullet points and a single ":" maybe?

ethanxxxl:
Yeah maybe TODO isn't the best thing to put there.  I didn't look into that parameter, so I don't feel confident documenting it though.

The `::` is a habit from emacs org-mode, we can use single colons :)

ethanxxxl:
rebased and snuck in documentation for `run-process` as well

extensions/process/process.lisp
```diff
@@ -25,6 +25,25 @@
     :reader process-output-callback-type)))
 
 (defun run-process (command &key name output-callback output-callback-type directory)
+  "Starts runs an external program as a separate thread.
```
vindarel: repetition:  "starts runs…"

vindarel:
Thanks! Just a typo and we're good.

ethanxxxl:
good catch, thanks!

PR #2119, opened 2026-03-14
The function `completion-symbol-async` always calls the `then` that is passed to it.

myaosato:
When the cursor was at a position where the function `lem/buffer/internal:symbol-string-at-point` would return `nil`, such as immediately after a closing parenthesis, calling `indent-line-and-complete-symbol` in `lisp-mode` caused an issue where completion candidates were not passed to the callback, and the spinner did not disappear.

Therefore, we changed `completion-symbol-async` to always call `then`.

### DIFF
--- a/extensions/lisp-mode/lisp-mode.lisp
+++ b/extensions/lisp-mode/lisp-mode.lisp
@@ -856,26 +856,27 @@
 (defun completion-symbol-async (point then)
   (check-connection)
   (let ((string (symbol-string-at-point point)))
-    (when string
-      (remote-eval-from-string
-       (current-connection)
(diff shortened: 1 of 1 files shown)


vindarel:
Hello, do you think you can add a unit test?

(If you confirm this fix works fine, I shall merge)

vindarel:
catsquotl feedback on Discord:

---

When in my old lem I open a lisp file and write 

(defun foo ()

 immediately followed by a tab I can reproduce the bug. 
I build another webview lem with the cherry picked pr above and the spinner seems to disappear when it needs to. so It does look like a fix.

vindarel:
@myaosato thanks!

PR #2121, opened 2026-03-20
improve prompt-window completion keys

mahmoodsh36:
previously:
- return key did both either complete or choose the current candidate, it was ambiguous.
- tab and shift-tab behaved just like C-n/C-p for navigating candidates in the completion, which is redundant imo
- M-n/M-p also behaved like C-n/C-p

now:

- return key selects the current candidate and exits immediately. `*completion-select-hook*` was introduced to make this work
- M-n/M-p navigate history in minibuffer completion prompt
- C-n/C-p navigate candidates
- tab completes the prompt to the currently focused candidate

### DIFF
--- a/src/ext/completion-mode.lisp
+++ b/src/ext/completion-mode.lisp
@@ -8,7 +8,8 @@
            :run-completion
            :completion-end
            :completion-mode
-           :completion-refresh)
+           :completion-refresh
+           :*completion-select-hook*)
   #+sbcl
   (:lock t))
 (in-package :lem/completion-mode)
@@ -17,6 +18,7 @@
 
 (defvar *completion-context* nil)
 (defvar *completion-reverse* nil)
+(defvar *completion-select-hook* nil)
 
 (defclass completion-context ()
   ((spec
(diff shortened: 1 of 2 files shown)


mahmoodsh36:
C-n/C-p worked before this commit though, my changes didnt touch them anyway.

vindarel:
I like Tab and shift-Tab to cycle through the completions,
so I tried your patch and this feels weird (actually unusable!).

If there's pressure or good reasons to change it, then so be it…

I like Enter's new behavior to directly select the candidate, but I fear it is not suitable in all contexts, in all types of completions:

- Alt-x … RET to choose a command: OK (and better)
- Alt-x project-find-file RET choose file RET => we might want to edit the selected file name
  - with your changes, we can complete the file name with Tab and edit it
- or imagine another command that prompts for strings but would accept an edited string. If Enter validated the prompt we could not edit it. Or, how?

Thoughts?

ps: would you enable C-n right after a Alt-x to start the completion? Currently it does nothing, only Tab opens the completion buffer.

mahmoodsh36:
> I like Tab and shift-Tab to cycle through the completions, so I tried your patch and this feels weird (actually unusable!).
> 
> If there's pressure or good reasons to change it, then so be it…
> 
> I like Enter's new behavior to directly select the candidate, but I fear it is not suitable in all contexts, in all types of completions:
> 
>     * Alt-x … RET to choose a command: OK (and better)
> 
>     * Alt-x project-find-file RET choose file RET => we might want to edit the selected file name
>       
>       * with your changes, we can complete the file name with Tab and edit it
> 
>     * or imagine another command that prompts for strings but would accept an edited string. If Enter validated the prompt we could not edit it. Or, how?
> 
> 
> Thoughts?
> 
> ps: would you enable C-n right after a Alt-x to start the completion? Currently it does nothing, only Tab opens the completion buffer.


this was the idea, to make return less ambiguous and use tab for selecting the current candidate for furhter editing. so return is for "choose this and exit" and tab is for "choose this".
i understand that this might be unintuitive for people that expect tab to also behave like C-n, perhaps we can have a variable for it like `*prompt-tab-for-complete*` or something? or maybe just bind this "complete without exiting" behavior to another key?

about the latter idea, im not sure i know what you mean, but wasnt `*automatic-tab-completion*` added for this purpose?

PR #2127, opened 2026-04-01
use uiop:parse-native-namestring instead of pathname

mahmoodsh36:
using `pathname` was causing errors when i tried to serve a file and view it from the webview client because the file had special characters (specifically it had `[` and it caused issues). the function `uiop:parse-native-namestring` knows to handle arbitrary chars in the filepath.


cxxxr:
Thank you!

PR #2130, opened 2026-04-08
Fixed a typo in lem-core:enable-clipboard-p.

RhysWheater:
Missing '*' at the end of variable name on line 40.


vindarel:
thanks!

PR #2134, opened 2026-04-13
[webview] Fix height of tab bar for macOS

theangelperalta:
# Summary

Header windows were previously hardcoded to a height of 1 character cell. This PR introduces a `header-window-height` generic function that subclasses can specialize to request more vertical space, and updates all layout calculations to respect it.

# Motivation

The server frontend's tabbar needs more than a single row to render properly (e.g., taller tab styling, padding). With a fixed height of 1, there was no way for a header window subclass to claim additional space without breaking the layout of editor windows below it.

Before:
Screenshot 2026-04-13 at 19 29 45
After:
Screenshot 2026-04-13 at 19 24 03

# Changes

Core (4 files):

- `header-window.lisp` — Add `defgeneric header-window-height` with a default method returning 1. Update `initialize-instance` to compute the y-position by summing existing header heights (using `loop :sum`) and to set height from the generic.
- `internal-packages.lisp` — Export `:header-window-height`.
- `frame.lisp` — `topleft-window-y` now sums header heights with `loop :sum` instead of `length`.
- `window.lisp` — `adjust-all-window-size` passes `(header-window-height window)` instead of 1.

Server frontend (1 file):

- `tabbar.lisp` — Specialize `header-window-height` to 2 for `tabbar-window`. Add `overflow-y: hidden` to prevent vertical scroll in the tab container.

### Backward compatibility

The default method returns 1, so all existing header window subclasses behave identically without changes.

### Style

Height summation uses `loop :for w :in ... :sum` per `contract.yml` `loop_keywords_rule`.

### DIFF
--- a/frontends/server/tabbar.lisp
+++ b/frontends/server/tabbar.lisp
@@ -85,6 +85,7 @@ html, body {
     linear-gradient(180deg, #1c1c1c, var(--bg)) /* subtle top gradient */;
   border-bottom: 1px solid var(--border);
   overflow-x: auto;         /* horizontal scroll for multiple tabs */
+  overflow-y: hidden;        /* prevent vertical scroll */
   white-space: nowrap;      /* no line wrap */
   scrollbar-width: none;    /* Firefox: hide scrollbar */
   position: relative;
--- a/frontends/server/tabbar.lisp
+++ b/frontends/server/tabbar.lisp
@@ -27,6 +27,9 @@
    (buffer-list
     :accessor tabbar-window-buffer-list)))
 
+(defmethod header-window-height ((window tabbar-window))
+  2)
+
 (defvar *tabbar* nil)
 
 (defun tabbar-init ()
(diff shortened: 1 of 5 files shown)


vindarel:
Hey, that's great, I see your generic function is documented, but this PR has even more information. Can you not include more info in the source? In the defgeneric, in comments, in `topleft-window-y` (no docstring) maybe… yeah I like docs and comments :imp:  WDYT?

theangelperalta:
> Hey, that's great, I see your generic function is documented, but this PR has even more information. Can you not include more info in the source? In the defgeneric, in comments, in `topleft-window-y` (no docstring) maybe… yeah I like docs and comments 👿 WDYT?

I just pushed some changes to align more with the style guide and added additional documentation. Let me know what you think @vindarel. I appreciate the review 🙏

vindarel:
Is it OK for you if I squash to 1 commit? (it slows you down for the best contributors' race…)

theangelperalta:
All good, I'm more interested in the improvements 😈

PR #2137, opened 2026-04-17
Fix vi-mode cursor entering read-only prompt area in REPL

theangelperalta:
## Problem

In listener-mode (REPL) with vi-mode enabled, switching from insert mode to normal mode allows the cursor to move into the read-only prompt region via backward movement commands (`h`, `b`, `0`, etc.). The prompt is marked with `:field` and `:read-only` text properties, but vi normal mode cursor movements don't respect the field boundary.

## Fix

Add `fall-within-field` in `commands/utils.lisp` — analogous to the existing `fall-within-line` which clamps the cursor at end-of-line. When the cursor lands within a `:field` region, it is moved past the field boundary via `next-single-property-change`.

This is called from the normal mode `post-command-hook` (before `fall-within-line`), so it applies after every command.

Navigation to previous lines (viewing REPL history) remains unaffected — only field regions on the current line are clamped.

## Changes

- `extensions/vi-mode/commands/utils.lisp`: Add `fall-within-field` function and export it
- `extensions/vi-mode/vi-mode.lisp`: Call `fall-within-field` in normal mode `post-command-hook`

### DIFF
--- a/extensions/vi-mode/commands/utils.lisp
+++ b/extensions/vi-mode/commands/utils.lisp
@@ -39,6 +39,7 @@
   (:export :bolp
            :eolp
            :fall-within-line
+           :fall-within-field
            :operator-pending-mode-p
            :read-universal-argument
            :*cursor-offset*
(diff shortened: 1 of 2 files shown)


theangelperalta:
closed because was the fix didn't cover other modes

PR #2141, opened 2026-04-18
fix: smooth scrolling in webview/server frontend

theangelperalta:
## Problem

Scrolling in the webview (and server-based) frontend is jerky or completely non-functional on trackpads, and coarse on mouse wheels.

### Root cause

The wheel event handler in `frontends/server/frontend/editor.js` converted browser pixel deltas with:

```javascript
wheelX: -Math.round(event.deltaX * 0.01),
wheelY: -Math.round(event.deltaY * 0.01),
```

This has three compounding issues:

1. **Trackpad events are silently dropped** — typical trackpad `deltaY` values are 1–30 pixels. After `* 0.01` and `Math.round()`, they become **0**, so the event is lost entirely.
2. **`deltaMode` is ignored** — the browser can report deltas in pixels (mode 0), lines (mode 1), or pages (mode 2). The old code treated all modes the same.
3. **No frame batching** — every single wheel event fires a WebSocket RPC round-trip, flooding the server with redundant messages that each trigger a full display redraw.

### Delta normalization

A new `normalizeWheelDelta()` function converts raw browser deltas to **line units** based on `deltaMode`:
- **Mode 0 (pixels):** divides by `fontHeight` to get fractional lines
- **Mode 1 (lines):** pass through as-is
- **Mode 2 (pages):** multiplies by 20 as a rough line estimate

### Sub-line accumulation

An `extractWholeLines()` function splits the accumulated float value into an integer scroll count and a fractional remainder. Only whole-line counts are sent over the wire; the remainder carries forward to the next frame. This means small trackpad gestures are never lost — they accumulate until they reach a full line.

### Frame-level coalescing

Multiple wheel events within a single animation frame are merged via `requestAnimationFrame`. Only one RPC call is made per frame at most, reducing WebSocket traffic and avoiding redundant redraws.

### Code structure

The logic is split into pure helper functions and a `makeWheelHandler(editor)` factory that encapsulates the minimal mutable state:

| Function | Purpose |
|----------|---------|
| `normalizeWheelDelta(deltaX, deltaY, deltaMode, fontHeight)` | Pure — converts raw deltas to line units |
| `extractWholeLines(accX, accY)` | Pure — splits float accumulator into integer + remainder |
| `cursorPosition(event, editor)` | Pure — derives pixel/character coords from a wheel event |
| `makeWheelHandler(editor)` | Factory — returns the event callback with isolated state |

## Files changed

- `frontends/server/frontend/editor.js` — wheel handler rewrite
- `frontends/server/frontend/dist/assets/index.js` — rebuilt production bundle

## Testing

- Verified build completes without errors (`npx vite build`)
- Trackpad: small gestures now accumulate and scroll correctly
- Mouse wheel: discrete notch events produce consistent 1-line scrolls
- Horizontal scroll (shift+wheel / trackpad swipe) works through the same path


vindarel:
:rocket: 

I think you have the opportunity to add high-level comments / design choices.

theangelperalta:
> 🚀
> 
> I think you have the opportunity to add high-level comments / design choices.

I added some high level documentation here. There is definitely opportunity to make the scrolling performance even better, but I won't let the perfect be the enemy of the good 🫡

PR #2147, opened 2026-04-24
fix(webview): raise cursor overlay z-index above floating windows

Catsquotl:
The .lem-cursor overlay had z-index: 100, but floating windows use z-index: 200 via zIndexTable in editor.js. This caused the floating window canvas to render on top of the cursor overlay, making the cursor invisible in floating windows such as Legit's peek and source panes.

Fix: raise .lem-cursor z-index to 300 in both editor.css and the bundled dist/assets/index.css.

Verified on Fedora 43 with WebKitGTK 2.52.1.


vindarel:
Thank youuuuuu

PR #2163, opened 2026-05-02
image-buffer: set buffer-filename when opening an image

### LINKED ISSUE #871: Bug that (buffer-%directory buffer) returns nil for image-viewer-mode but works for other files
I have this nifty command which I like to use to open the directory-mode in the current directory of the buffer and it works quite well. But if an image is open, `(lem:buffer-directory)` returns `(uiop:getcwd)` which is incorrect, meaning that `(buffer-%directory buffer)` is `nil` for `image-viewer-mode`. How would one fix that?

```lisp
(define-command dima-directory-to () ()
  (lem-core/commands/file:find-file
   (lem:buffer-directory)))
```

Maybe it also fails for other buffers, I just noticed this for images.

nathan-casabieille:
## Description
`open-image-buffer` was setting `:directory` on the buffer but never
setting `buffer-filename`. This caused two issues:

- `(buffer-filename)` returned `nil` on image buffers
- `C-x C-j` (find-file-directory) failed with "No file at point"

The fix sets `buffer-filename` to the image path, and calls
`update-changed-disk-date` right after so Lem doesn't see the
newly-set filename as a file that changed on disk - which would
trigger an unwanted revert and an encoding error on the binary.

## Related Issues
Fixes #871

### DIFF
--- a/src/ext/image-buffer.lisp
+++ b/src/ext/image-buffer.lisp
@@ -53,6 +53,8 @@ function resetImage() {
                   :html
                   (mustache:render* *html-template*
                                     `((image . ,(namestring pathname)))))
+    (setf (lem:buffer-filename buffer) (namestring pathname))
+    (lem:update-changed-disk-date buffer)
     (lem:change-buffer-mode buffer 'image-viewer-mode)))
 
 (defclass find-file-executor (lem:find-file-executor) ())


theangelperalta:
@nathan-casabieille Thank you 👏

PR #2172, opened 2026-05-14
perf(core): memoize parse-color, keybindings, and version lookups

theangelperalta:
## Summary

Memoize three pure-but-hot lookups called every frame:

- **`parse-color`** — `EQUAL` hash table with a nil sentinel for invalid input; `clear-parse-color-cache` invoked from `attribute.lisp` on theme change so cached colors don't outlive the active theme.
- **`get-keybinding-map`** — generation-counter cache; `invalidate-keybinding-cache` runs on `define-key` / `undefine-key`, so all subsequent lookups hit the cache until the next rebinding.
- **ASDF version** — resolve `asdf:component-version` once at load time and cache in `src/version.lisp`; `src/modeline.lisp` reads the cached string instead of walking system metadata on each redraw.

## Files

| File | Change |
|---|---|
| `src/common/color.lisp` | parse-color memoization + cache-clear hook |
| `src/attribute.lisp` | call clear-parse-color-cache on theme change |
| `src/keymap.lisp` | generation-based keybinding cache |
| `src/modeline.lisp` | use cached version string |
| `src/version.lisp` | cache asdf:component-version at load time |

~100 lines changed.

## Test plan

- [ ] Smoke-test: open Lem, exercise common edits, switch themes — no regressions
- [ ] Confirm theme change invalidates parse-color cache (re-parses on next access)
- [ ] Rebind a key, confirm new binding takes effect

### DIFF
--- a/src/attribute.lisp
+++ b/src/attribute.lisp
@@ -120,7 +120,11 @@
 
 (defun clear-all-attribute-cache ()
   (dolist (attribute *attributes*)
-    (setf (get attribute '%attribute-value) nil)))
+    (setf (get attribute '%attribute-value) nil))
+  ;; Also clear the parse-color memoization cache, since color theme
+  ;; changes may reassign the meaning of named colors or invalidate
+  ;; cached results.
+  (lem/common/color:clear-parse-color-cache))
 
 (defun get-attribute-cache (attribute &rest args &key background)
   (declare (ignore background))
--- a/src/common/color.lisp
+++ b/src/common/color.lisp
@@ -8,6 +8,7 @@
            :color-equal
            :light-color-p
            :parse-color
+           :clear-parse-color-cache
            :rgb-to-hsv
            :hsv-to-rgb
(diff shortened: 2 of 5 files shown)


theangelperalta:
Addressed both code-contractor violations in e1c0b4c9:

**`file_structure_rule` (modeline.lisp:133)** — moved `*cached-lem-version-string*` defvar up next to `*modeline-status-list*`, before the function definitions.

**`functional_style_rule` (keymap.lisp:575)** — moved the per-keymap binding cache *table* onto a new `binding-cache` slot on the `keymap` class itself, so each keymap owns its own cache. The remaining `*keybinding-cache-generation*` defvar is now relocated to the top of the file and documented as an editor-wide modification clock (analogous to `*current-buffer*`) — the rule's stated exception for editor-wide dynamic context. Lifting that counter to a slot would require parent back-pointers, since a `define-key` on a child keymap has to invalidate any parent that has indexed it.

PR #2173, opened 2026-05-14
perf(display): reduce structural allocations in redraw loop

theangelperalta:
## Summary

Trim per-frame allocation in the core redraw path (`src/display/physical-line.lisp`):

- **Destructive `drawing-object-merge`** — mutate the input list rather than allocating a new one.
- **Drop `copy-list` in `reduce-list`** — the caller no longer needs the original.
- **`clip-objects-to-display-range`** — compute width via the monospace formula and emit zero letter-objects when the row is fully clipped.
- **`make-object-with-type` with literal class names** — so SBCL's constructor caching kicks in (instead of looking up the class each call).
- **Line fingerprint cache for `redraw-logical-line-when-horizontal-scroll`** — unchanged lines short-circuit re-redraw.

## Files

- `src/display/physical-line.lisp` — ~150 lines

## Test plan

- [ ] Smoke-test scrolling and editing under both ncurses and sdl2 frontends
- [ ] Horizontal scroll on a long line, confirm no rendering glitches
- [ ] Verify clipped rows don't emit visible artifacts at scroll edges

### DIFF
--- a/src/display/physical-line.lisp
+++ b/src/display/physical-line.lisp
@@ -149,12 +149,14 @@
   drawing-object-1)
 
 (defmethod drawing-object-merge ((drawing-object-1 text-object) (drawing-object-2 text-object))
-  (make-instance 'text-object
-                 :string (str:concat (text-object-string drawing-object-1)
-                                     (text-object-string drawing-object-2))
-                 :attribute (text-object-attribute drawing-object-1)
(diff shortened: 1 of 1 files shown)


vindarel:
Hi, did you sort of measure the perf improvements?

PR #2189, opened 2026-05-17
removed code that conflicts with qlot asdf environment

ethanxxxl:
# Overview 
qlot establishes an ASFD environment to find and use project-local packages.  As I have been tinkering with Lem, I have also been messing with Lem's dependencies as well.  The best method I have found to incorporate my changes locally is to fork/clone the code of the dependency, then in the `qlfile` change the dependency from `git [system] [url]` to `local [system] [pathname]`. 

For a while now, this has allowed me to compile Lem with my changes baked in.  There has always been weird behavior when trying to add packages in after compiling, however.  This is especially a problem when adding new dependencies that no extensions or subsystems depend on yet.  Even if I specified a local installation in the qlfile, asdf and quicklisp would reach out and download the existing version if I tried to load it after starting lem.

The problem seems to be that qlots configuration is being overwritten at some point.  I tracked down two places where this was happening:

## src/lem.lisp

the function `initialize-source-registry` is called during end of the build.  I'm not exactly sure of the details of initializing the source registry, but it had the effect of clearing the configuration done by qlot.

## extensions/lisp-mode/lisp-mode.lisp

the `self-connect` function, called when opening a new lisp-repl, cleared the `asdf:source-registry`.  This line was written before Lem was using qlot, I think the problem it was trying to solve is no longer relevant.

### DIFF
--- a/extensions/lisp-mode/lisp-mode.lisp
+++ b/extensions/lisp-mode/lisp-mode.lisp
@@ -236,11 +236,6 @@
       (connect-to-micros *localhost* port)
       (update-buffer-package)
 
-      ;; XXX:
-      ;; Systems added after lem initialization are not visible from within this process and must
-      ;; be re-initialized.
-      (asdf:clear-source-registry)
-
       (setf *self-connected-port* port))))
 
 (defun self-connection ()
--- a/src/lem.lisp
+++ b/src/lem.lisp
@@ -47,20 +47,12 @@
       (unless (uiop:pathname-equal current-dir (user-homedir-pathname))
         (maybe-load (merge-pathnames ".lemrc" current-dir))))))
 
-(defun initialize-source-registry ()
-  (asdf:initialize-source-registry
-   `(:source-registry
-     :inherit-configuration
-     (:also-exclude ".qlot")
-     (:tree ,(asdf:system-source-directory :lem)))))
-
 (defun init-at-build-time ()
   "This function is called when an lem executable file is built.
 If a file named $HOME/.lem/build-init.lisp exists, it is loaded.
 The difference is that init.lisp loading is called when the editor is started,
 while build-init.lisp is called when the binary file is created.
 See scripts/build-ncurses.lisp or scripts/build-sdl2.lisp"
-  (initialize-source-registry)
   (let ((file (merge-pathnames "build-init.lisp" (lem-home))))
     (when (uiop:file-exists-p file)
       (load file))))


ethanxxxl:
@vindarel  can we review and merge these changes?

vindarel:
disclaimer: I am barely a Qlot user. Questions:

- does that change anything in how you start Lem / do you need to update the README / documentation?
- cutting the grass under asdf seems necessary (I've been caught by that kind of stuff). Can we however write out a test plan and do it after merging?

(thanks!)

theangelperalta (commented):
**Reviewed the change and tested the ASDF behavior directly (SBCL 2.5.10).** TL;DR: the diagnosis is right, the fix is sound, and removing the `clear-source-registry` call does **not** regress anything.

The key thing I wanted to confirm was the worry that deleting `(asdf:clear-source-registry)` from `self-connect` would stop the self REPL from discovering systems added after startup — since the old `XXX` comment claimed that's what it was for. So I set up a tree-based source registry (the Qlot-style setup), warmed the cache, dropped in a brand-new `.asd`, and tried each refresh path:

| Action | New `.asd` (added at runtime) discovered? |
|---|---|
| nothing (warm cache) | no |
| `asdf:clear-source-registry` | no |
| `asdf:clear-configuration` | no |
| `asdf:initialize-source-registry` (re-supply parameter) | yes |

Two takeaways:

1. **`clear-source-registry` was not doing what its comment claimed.** It clears the cache but the next lazy lookup just rebuilds the *same* directory listing — it does not rescan disk for new `.asd` files. Only a fresh `initialize-source-registry` forces a re-scan. So the deleted line never actually made "systems added after lem initialization" visible; removing it loses no real behavior.

2. **`clear-source-registry` preserves the existing programmatic config** (a previously-registered system still resolved after clearing — it reverts to the last `initialize-source-registry` parameter, i.e. Qlot's). So that call wasn't the thing clobbering Qlot's setup.

The real culprit is the `src/lem.lisp` `initialize-source-registry`, which *replaced* Qlot's parameter with one that `:also-exclude`d `.qlot` and re-rooted the tree — that's what hid the `local`-pinned deps. Removing it is the substantive, correct fix.

**On @vindarel's questions:**
- No change to how Lem starts; no doc change strictly required (though documenting the `local` qlfile workflow for hacking on deps would be a nice follow-up).
- A merge-then-test plan is fine. Suggested manual check: `local`-pin a dependency in the qlfile, build, and confirm your changes are baked in and resolved from the local path. The "drop in a brand-new system from the running self REPL" case is worth noting as *already* unsupported both before and after this PR — picking up a genuinely new system at runtime still requires `(asdf:initialize-source-registry …)` / `(ql:register-local-projects)`. If desired, exposing that as a small command would be a reasonable separate enhancement, not a blocker here.

LGTM.

theangelperalta:
Opened #2203 to track the follow-up: an explicit `lisp-mode` command to register/refresh newly-added ASDF systems at runtime (the working replacement for what the deleted `clear-source-registry` was meant to do). Not a blocker for merging this PR.

theangelperalta (approved):

vindarel:
Let's have 3 pairs of eyes (or more) on this now ;)

PR #1964, opened 2025-09-18
Remove binary files

mldulaney:
Open source software should, by definition, not be distributing binary files.  Given the xz backdoor, these increase distrust.

PR #2207, opened 2026-06-03
feat(nix): build and bundle terminal.so in the flake

theangelperalta:
## Summary

Makes the terminal extension work in Nix builds by compiling and bundling `terminal.so` in the flake. Until now the flake had **no libvterm input and no terminal build step**, so on Nix the terminal extension silently disabled itself.

## Why

`ffi.lisp` loads `terminal.so` best-effort (`ignore-errors`), and the flake builds Lem via `lisp.buildASDFSystem` with native deps declared in `nativeLibs` (ncurses, tree-sitter, `ts-wrapper`, webview) — but **never libvterm/terminal.so**. So:

- The terminal never worked under Nix, even before #2204: the committed `terminal.so` was dynamically linked against a libvterm that isn't in the Nix sandbox.
- Removing the committed binaries (#2204) didn't regress Nix — same end state — but left it with no path to a working terminal.

## Change

- New `terminal-so` derivation that compiles `extensions/terminal/terminal.c` against `pkgs.libvterm`, mirroring the existing `ts-wrapper` / `c-webview` C derivations.
- Added to `nativeLibs` of `lem-ncurses`, `lem-sdl2`, and `lem-webview` (both branches).

### Why dynamic (not static like #2205)

On Nix, dynamic linking is the idiomatic and more reliable choice:

- The stdenv records an **RPATH** to the pinned `libvterm` store path, so `terminal.so` resolves libvterm at runtime with no bundling or relinking — exactly how `ts-wrapper` resolves tree-sitter today.
- `nativeLibs` puts `terminal.so` itself on the library path that `ffi.lisp`'s `"terminal.so"` lookup uses, at both build time (so CFFI registers it for image-restore reload) and run time (baked into the `wrapProgram` `LD_LIBRARY_PATH`).
- Static linking would require a `libvterm.a` that nixpkgs doesn't ship by default.

The output keeps the `.so` suffix on both platforms (not `.dylib`), because `ffi.lisp` loads it by the literal name `terminal.so` even on macOS.

## Test plan

- [ ] **Nix CI** (`.github/workflows/ci.yml`) runs `nix flake check` + `nix build .#lem-ncurses/.#lem-webview/.#lem-sdl2` on Linux and macOS — this builds the `terminal-so` derivation and is the authoritative check. *(Could not run `nix` locally — no nix on this machine.)*
- [ ] Confirm `pkgs.libvterm` is the correct attr and its version compiles `terminal.c` (the `conceal`/`strike` cell-attr fields need libvterm ≥ 0.3; if nixpkgs' `libvterm` is too old or misnamed, fall back to `pkgs.libvterm-neovim`).
- [ ] Launch the Nix-built `lem`, open a `*terminal*` buffer, confirm a shell runs.

Refs #1964, #2060, #2204, #2205.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -353,8 +353,10 @@
               pkgs.ncurses
               pkgs.tree-sitter
               ts-wrapper
-              terminal-so
-            ];
+            ]
+            # libvterm is Linux-only in nixpkgs, so terminal-so (and thus the
+            # terminal extension) is only available on Linux Nix builds.
+            ++ pkgs.lib.optionals pkgs.stdenv.isLinux [ terminal-so ];
             # Add tree-sitter grammar paths (grammars don't have lib/ subdir)
             installPhase = ''
               runHook preInstall
@@ -380,14 +382,15 @@
                 sdl2-image
                 trivial-main-thread
               ]);
-            nativeLibs = with pkgs; [
+            nativeLibs = (with pkgs; [
               SDL2
               SDL2_ttf
               SDL2_image
               tree-sitter
               ts-wrapper
-              terminal-so
-            ];
+            ])
+            # libvterm is Linux-only in nixpkgs (see lem-ncurses).
+            ++ pkgs.lib.optionals pkgs.stdenv.isLinux [ terminal-so ];
             installPhase = ''
               runHook preInstall
               mkdir -p $out/bin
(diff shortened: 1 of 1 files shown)


haxfn:
Seems to work for me on NixOS:

image

PR #2218, opened 2026-06-04
Save cursor state when changing history in repl

JesperAxelsson:
Fixes a few issues with history navigation. And are probably the root cause to #2214.
- You can now step through multi line history items
- Fix issue where starting repl line were lost when trying to search past end of history

I split out a helper to share the logic of previous and next. Might we worth moving the 
searching to the history package at a later date.

```
[ THE BUG: Forcing an old position ]
You start with a longer string, cursor at the end.
Buffer:  cl-user> 'hello
Cursor:                 ^ (Old position saved)

Code replaces the text, then forcefully shoves 
the cursor back to that saved position.
Buffer:  cl-user> 'he
Cursor:                 ^ 💥 CRASH: Timer wakes up
                             and sees cursor in the void!

------------------------------------------

[ THE FIX: Safe linear offset ]
Buffer starts the same.
Buffer:  cl-user> 'hello

Code safely anchors to the start, then steps 
forward exactly `prefix-len`.
Buffer:  cl-user> 'he
Cursor:             ^ (Perfectly safe!)
==========================================
```

### DIFF
--- a/src/ext/listener-mode.lisp
+++ b/src/ext/listener-mode.lisp
@@ -183,7 +183,7 @@
     (when win
       (replace-textarea buffer str))))
 
-(defun %search-history-startwith (direction)
+(defun %search-history-startswith-input (direction)
   "Internal helper for history next and previous history navigation."
   (multiple-value-bind (step-fn rewind-fn)
       (ecase direction
@@ -219,13 +219,13 @@
   "Find the previous prompt starting with the current input.
 
   See also `listener-previous-input`."
-  (%search-history-startwith :previous))
+  (%search-history-startswith-input :previous))
 
 (define-command listener-next-startswith-input () ()
   "Find the next prompt starting with the current input.
 
   See also `listener-next-input`."
-  (%search-history-startwith :next))
+  (%search-history-startswith-input :next))
 
 (define-command listener-previous-input () ()
   "Get and insert the previous REPL input."
--- a/src/ext/listener-mode.lisp
+++ b/src/ext/listener-mode.lisp
@@ -185,35 +185,34 @@
 
 (defun %search-history-startswith-input (direction)
   "Internal helper for history next and previous history navigation."
-  (multiple-value-bind (step-fn rewind-fn)
-      (ecase direction
-        (:previous (values #'lem/common/history:previous-history
-                           #'lem/common/history:next-history))
-        (:next (values #'lem/common/history:next-history
-                       #'lem/common/history:previous-history)))
-    (let* ((buffer (current-buffer))
-           (point (buffer-point buffer))
-           (start (input-start-point buffer))
-           (prefix (points-to-string (input-start-point buffer) point))
-           (prefix-len (length prefix)))
-      (backup-edit-string (current-buffer))
-      (loop :for steps :from 0
-            :do (multiple-value-bind (str win)
-                    (funcall step-fn (current-listener-history))
-                  (if win
-                      (when (eql 0 (search prefix str :test #'string=))
-                        (replace-textarea buffer str)
-                        (lem-core:move-point point start)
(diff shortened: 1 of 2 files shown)


theangelperalta:
Really nice fix — the diagram in the description made the root cause click immediately. Anchoring to `input-start-point` and stepping by `prefix-len` instead of forcing a saved column is exactly the kind of change that makes the whole thing robust for multi-line items. 👏

I confirmed in `src/common/history.lisp` that `previous-history`/`next-history` only move the index on success and leave it untouched at the boundary — so your `steps`-counting + rewind logic lines up perfectly. That's a real improvement over the old path that walked the index to 0 and left it stranded.

A few small things I was curious about, mostly to make sure I understand the intent:

1. **Package qualifiers** — the file is `(:use :cl :lem)` and everything around your change uses unqualified `move-point` / `character-offset` / `buffer-end`. Was there a reason to reach for `lem-core:move-point` and `lem-core:character-offset` here specifically? Dropping the prefix would match the surrounding style (and the `contract.yml` preference for `lem:` over deeper packages).

2. **The rewind on the `:next` path** — since `restore-edit-string` already sets `index` back to `(length data)` unconditionally, does the `(dotimes (i steps) (funcall rewind-fn ...))` just before it ever have an effect when `direction` is `:next`? If I'm reading it right it only matters for `:previous` — would it be clearer to guard it or note that?

3. **Naming** — tiny one: the helper is `%search-history-startwith` while the commands are `…-startswith-input`. Intentional, or a stray missing `s`?

4. **A regression test?** — this fixes a crash with a lovely concrete repro (multi-line item, cursor past EOL). Would it be worth capturing that so it can't quietly come back? The cursor logic lives in listener-mode rather than the already-tested `common/history`, so I understand it's a bit more setup.

None of these are blockers from my side — the core fix is sound and I'd be happy to see it land. Thanks for splitting out the helper too; the previous/next duplication had been bugging me. 🙂

JesperAxelsson:
Appreciate the feed back!

> 
> A few small things I was curious about, mostly to make sure I understand the intent:
> 
>     1. **Package qualifiers** — the file is `(:use :cl :lem)` and everything around your change uses unqualified `move-point` / `character-offset` / `buffer-end`. Was there a reason to reach for `lem-core:move-point` and `lem-core:character-offset` here specifically? Dropping the prefix would match the surrounding style (and the `contract.yml` preference for `lem:` over deeper packages).
> 

No real reason. I was a bit unsure about the conventions, the history functions did use qualifier so I thought it were standard.
Removed the qualifiers now, and will remember to skip them for core lem functions.


>     2. **The rewind on the `:next` path** — since `restore-edit-string` already sets `index` back to `(length data)` unconditionally, does the `(dotimes (i steps) (funcall rewind-fn ...))` just before it ever have an effect when `direction` is `:next`? If I'm reading it right it only matters for `:previous` — would it be clearer to guard it or note that?
> 

True, changed it. Caused a bit of an avalanche of refactorings. The function do look cleaner now.

>     3. **Naming** — tiny one: the helper is `%search-history-startwith` while the commands are `…-startswith-input`. Intentional, or a stray missing `s`?
> 

Add the `-input` suffix. Pondered fixing the kebab casing but thought it better to match the command names. Easier to grep.

>     4. **A regression test?** — this fixes a crash with a lovely concrete repro (multi-line item, cursor past EOL). Would it be worth capturing that so it can't quietly come back? The cursor logic lives in listener-mode rather than the already-tested `common/history`, so I understand it's a bit more setup.
> 

Added a small test as well.

> 
> None of these are blockers from my side — the core fix is sound and I'd be happy to see it land. Thanks for splitting out the helper too; the previous/next duplication had been bugging me. 🙂

Your welcome :)

PR #2228, opened 2026-06-13
fix(lisp-syntax): don't apply macro indentation to keyword symbols

skyizwhite:
## Summary

When computing indentation, `find-indent-method` strips the package prefix of a package-qualified symbol (e.g. `alexandria:with-gensyms` → `with-gensyms`) so it can look up an indentation rule by the bare symbol name.

The previous regex `(?<=:)[^:]+` had no start anchor, so it also matched the part after the leading colon of a **keyword** symbol. For `:use` it returned `"use"`. If the user (or a loaded system) had defined a macro named `use` with a `&body`-style indentation rule, that rule was incorrectly applied to the `:use` clause of `defpackage`, breaking its indentation.

```lisp
(defpackage :foo
  (:use :cl
     :bar))
(in-package :foo)

(defmacro use (&body body)
  ... )
```

This replaces the lookbehind with an anchored, capturing regex `^[^:]+:{1,2}([^:].*)$` that requires a **non-empty package prefix** before the `:`/`::` separator. Keywords (which start with `:` and thus have no prefix) and unqualified symbols are now left untouched, while genuine package-qualified symbols still resolve correctly.

## Test

Adds `defpackage-use-clause-not-affected-by-use-macro` to `tests/lisp-syntax/indent-test.lisp`. It temporarily registers a `&body` indentation rule for the name `use`, then asserts that re-indenting a `defpackage` form with a `:use` clause leaves it unchanged. The registration is restored via `unwind-protect` to avoid leaking global indent-table state into other tests.

Verified the new test fails against the old regex (`:use` → `"use"`) and passes with the fix; the full `lem-tests/lisp-syntax/indent-test` suite is green.

### DIFF
--- a/extensions/lisp-syntax/indent.lisp
+++ b/extensions/lisp-syntax/indent.lisp
@@ -433,7 +433,9 @@
            (when method
              (return-from find-indent-method method))))
     (f (get-indentation name))
-    (let ((name1 (ppcre:scan-to-strings "(?<=:)[^:]+" name)))
+    (let ((name1 (let ((registers (nth-value 1 (ppcre:scan-to-strings "^[^:]+:{1,2}([^:].*)$" name))))
+                   (when (and registers (plusp (length registers)))
+                     (aref registers 0)))))
       (when name1
         (f (get-indentation name1)))
       (f (and *get-method-function*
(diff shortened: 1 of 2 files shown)


theangelperalta:
Thank you for this fix @skyizwhite 🙏

PR #2230, opened 2026-06-16
folding with overlays

mahmoodsh36:
this is an attempt to get folding support using overlays with the `:invisible` attribute and other special attributes for overlays. in emacs there are 2 approaches for folding that im aware of, first is using text attributes and the second uses overlays. this approach is inspired by the second.

currently `:invisible`, `:before-string`, `:after-string` have been implemented with the most minimal changes possible. efficiency wasnt taken into account.

reference for emacs overlay attributes [https://www.gnu.org/software/emacs/manual/html_node/elisp/Managing-Overlays.html]()

### DIFF
--- a/extensions/vi-mode/commands/utils.lisp
+++ b/extensions/vi-mode/commands/utils.lisp
@@ -61,7 +61,10 @@
             (1- len)))))
 
 (defun fall-within-line (point)
-  (when (eolp point)
+  "clamps the given point to the character before the newline at the end of the line its in."
+  ;; a buffer-line end whose newline is hidden by a fold is mid-visual-line, so we dont clamp.
+  (when (and (eolp point)
+             (not (invisible-overlay-covering point)))
     (line-end point)
     (unless (bolp point)
       (character-offset point *cursor-offset*))))
--- a/extensions/vi-mode/commands/utils.lisp
+++ b/extensions/vi-mode/commands/utils.lisp
@@ -61,7 +61,7 @@
             (1- len)))))
 
 (defun fall-within-line (point)
-  "clamps the given point to the character before the newline at the end of the line its in."
+  "Clamps the given point to the character before the newline at the end of the line it is in."
   ;; a buffer-line end whose newline is hidden by a fold is mid-visual-line, so we dont clamp.
   (when (and (eolp point)
              (not (invisible-overlay-covering point)))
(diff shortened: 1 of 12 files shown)


mahmoodsh36:
currently there is a confusion between the concept of a 'logical-line' and a 'visual-line' in the code. especially the code in logical-line.lisp could do with some renaming.

also, performance is bad. this PR doesnt address performance issues, atleast not yet.

mahmoodsh36:
pressing tab on the first line in a defun in insert-mode will fold/unfold it. by default it uses `beginning-of-defun-function` and `end-of-defun-function` so it should work for any language-mode that properly defines those.

mahmoodsh36:
the following is a demo for how virtual text can be used with the customizable cursor enter/exit behavior:
```lisp
(lem:define-command demo-reveal-overlay () ()
  (let ((buffer (lem:make-buffer "*reveal-demo*")))
    (lem:erase-buffer buffer)
    (let ((point (lem:buffer-point buffer)))
      (lem:insert-string point "some of the text in this sentence will be hidden and revealed when cursor exits/enters it")
      (lem:switch-to-buffer buffer)
      (lem:buffer-start point)
      (lem:with-point ((start point)
                       (end point))
        (lem:character-offset start 34)
        (lem:character-offset end 48)
        (lem:place-region-placeholder-overlay
         start
         end
         :placeholder "....."
         :is-line-fold nil
         :cursor-behavior :reveal)))))
```

extensions/vi-mode/commands/utils.lisp
```diff
@@ -61,7 +61,10 @@
             (1- len)))))
 
 (defun fall-within-line (point)
-  (when (eolp point)
+  "clamps the given point to the character before the newline at the end of the line its in."
```
vindarel: small english type: "its" -> "it is"

vindarel: (Also usually a docstring starts with a Capital word, here and below)

vindarel:
I see that Tab now tries to fold/unfold by default.

```diff
-(define-key *language-mode-keymap* "Tab" 'indent-line-and-complete-symbol)
+(define-key *language-mode-keymap* "Tab" 'fold-or-indent-or-complete)
```

is there a way to disable folding (other than re-binding Tab)? I think a boolean parameter would be useful.

PR #2234, opened 2026-06-18
display-time-mode

VisenDev:
I need to figure more about how the mode line works to finish implementing this. I structured my code off of modeline-battery-mode but it seems like that mode doesn't work.

EDIT: Okay I figured out how to do it now


VisenDev:
Screenshot_20260618_151504

VisenDev:
Hmm okay the text is much less legible when I look at it on my other monitor. I've got to fix that

EDIT: Should be fixed now

vindarel:
Well done!

Can you add a note about usage in the documentation website? :pray:

PR #2237, opened 2026-06-20
Add print object method for prefix class

VisenDev:
I was trying to inspect how the keybindings were implemented, and I noticed it was quite difficult to look through the different keymaps because the list of the prefixes didn't give any information about what key corresponded to those prefixes. 

I added a print object method that will now show what key the prefix uses in its printed representation
Screenshot_20260619_201805

### DIFF
--- a/src/keymap.lisp
+++ b/src/keymap.lisp
@@ -34,6 +34,10 @@ NIL to append it to the key sequence normally.")
     :initform nil
     :documentation "extra metadata that a prefix may hold.")))
 
+(defmethod print-object ((object prefix) stream)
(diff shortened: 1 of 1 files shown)


vindarel:
The class definition:

```lisp
(defclass prefix ()
  ((key
    :initarg :key
    :documentation "the key defined for the prefix. could be a function that returns a key.")
  …
```
`:key` has no initform, it is unbound by default -> let's be extra cautious and please add a check with `slot-boundp` before `slot-value`.

VisenDev:
> let's be extra cautious and please add a check with `slot-boundp` before `slot-value`.

Good point, that should be fixed now

Issue #2231, opened 2026-06-16
Emacs Style Help Keybindings Mode


VisenDev:
It would be nice if there was a mode/extension that rebound `C-h` to support emacs help commands, like `C-h b` for `describe-bindings`, `C-h k` for `describe-key`, or `C-h f` for `describe-function`. Currently the discoverability of lem is a bit limited coming from emacs. I'm ending up needing to grep the lem source code to find the source code for various things.

I might work on this myself if I get some time, but I wanted to create an issue to bring awareness and discussion.

VisenDev:
related: https://github.com/lem-project/lem/issues/407

I am not suggesting that `C-h` do this by default, but rather that there should be a mode to do this if desired. Also, many of the calls like `describe-function` or `describe-variable` aren't implemented in lem, so this issue at the very least concerns adding those, even if the keybindings are not added

VisenDev:
completed by #2240

PR #2240, opened 2026-06-21
Emacs Help Mode

VisenDev:
Initial work to address https://github.com/lem-project/lem/issues/2231

I'm not sure if the only way to do this is to overwrite and restore the `ctrl-h` binding in the *global-map* or if using a minor mode keymap will help me here. I'd appreciate any advice on whats the best way to replace just a single keybinding prefix like this

### DIFF
--- a/extensions/emacs-help-mode/main.lisp
+++ b/extensions/emacs-help-mode/main.lisp
@@ -10,9 +10,9 @@
 (define-key *ctrl-h-keymap* "b" 'describe-bindings)
 (define-key *ctrl-h-keymap* "m" 'describe-mode)
 (define-key *ctrl-h-keymap* "a" 'apropos-command)
+(define-key *ctrl-h-keymap* "v" 'apropos-variable)
 
-;; TODO add describe-variable command for "v" and
-;; add describe-function command for "f"
+;; TODO add describe-function command for "f"
 
 (defun enable (&optional recursed-p)
   "Enables emacs help mode"
--- a/extensions/emacs-help-mode/main.lisp
+++ b/extensions/emacs-help-mode/main.lisp
@@ -3,6 +3,7 @@
 (in-package #:lem-emacs-help-mode)
 
 (defvar *previous-ctrl-h-suffix* nil)
+(defvar *previous-describe-output-override* nil)
 (defvar *ctrl-h-keymap* (make-keymap))
 
 (define-key *ctrl-h-keymap* "k" 'describe-key)
@@ -22,7 +23,10 @@
     (cond
       (index
        (setf *previous-ctrl-h-suffix* (prefix-suffix (nth index prefixes)))
-       (setf (prefix-suffix (nth index prefixes)) *ctrl-h-keymap*))
+       (setf (prefix-suffix (nth index prefixes)) *ctrl-h-keymap*)
+       (setf *previous-describe-output-override*
+             lem-core/commands/help:*describe-output-type-override*)
+       (setf lem-core/commands/help:*describe-output-type-override* :buffer))
       (recursed-p
        (error "Infinite loop"))
       (t
@@ -37,7 +41,10 @@
                           :test #'lem-core:key-equal)))
     (assert index)
     (setf (prefix-suffix (nth index prefixes)) *previous-ctrl-h-suffix*)
-    (setf *previous-ctrl-h-suffix* nil)))
+    (setf *previous-ctrl-h-suffix* nil)
+    (setf lem-core/commands/help:*describe-output-type-override*
+          *previous-describe-output-override*)
+    (setf *previous-describe-output-override* nil)))
 
 (define-minor-mode emacs-help-mode
     (:name "EHelp"
--- a/extensions/emacs-help-mode/main.lisp
+++ b/extensions/emacs-help-mode/main.lisp
@@ -2,53 +2,38 @@
   (:use #:cl #:lem))
 (in-package #:lem-emacs-help-mode)
 
-(defvar *previous-ctrl-h-suffix* nil)
-(defvar *previous-describe-output-override* nil)
+(defvar *emacs-help-mode-keymap* (make-keymap))
 (defvar *ctrl-h-keymap* (make-keymap))
+(defvar *previous-describe-output-override* nil)
 
 (define-key *ctrl-h-keymap* "k" 'describe-key)
 (define-key *ctrl-h-keymap* "b" 'describe-bindings)
 (define-key *ctrl-h-keymap* "m" 'describe-mode)
 (define-key *ctrl-h-keymap* "a" 'apropos-command)
 (define-key *ctrl-h-keymap* "v" 'apropos-variable)
-
 ;; TODO add describe-function command for "f"
+
+(define-key *emacs-help-mode-keymap* "C-h" *ctrl-h-keymap*)
 
 (defun enable (&optional recursed-p)
   "Enables emacs help mode"
-  (let* ((prefixes (keymap-prefixes *global-keymap*))
-         (ctrl-h (first (parse-keyspec "C-h")))
-         (index (position ctrl-h prefixes :key #'prefix-key :test #'key-equal)))
 
-    (cond
-      (index
-       (setf *previous-ctrl-h-suffix* (prefix-suffix (nth index prefixes)))
-       (setf (prefix-suffix (nth index prefixes)) *ctrl-h-keymap*)
-       (setf *previous-describe-output-override*
-             lem-core/commands/help:*describe-output-type-override*)
-       (setf lem-core/commands/help:*describe-output-type-override* :buffer))
-      (recursed-p
-       (error "Infinite loop"))
-      (t
-       (push (make-prefix :key ctrl-h) (keymap-prefixes *global-keymap*))
-       (enable t)))))
+  ;; TODO, figure out a more functional way to overwrite this
+  ;; variable which does not cause mutable global state
+  (setf *previous-describe-output-override*
+        lem-core/commands/help:*describe-output-type-override*)
+  (setf lem-core/commands/help:*describe-output-type-override* :buffer))
 
 (defun disable ()
   "Disables emacs help mode"
-  (let* ((prefixes (keymap-prefixes *global-keymap*))
-         (ctrl-h (first (parse-keyspec "C-h")))
-         (index (position ctrl-h prefixes :key #'prefix-key
-                          :test #'lem-core:key-equal)))
-    (assert index)
-    (setf (prefix-suffix (nth index prefixes)) *previous-ctrl-h-suffix*)
-    (setf *previous-ctrl-h-suffix* nil)
-    (setf lem-core/commands/help:*describe-output-type-override*
-          *previous-describe-output-override*)
-    (setf *previous-describe-output-override* nil)))
+  (setf lem-core/commands/help:*describe-output-type-override*
+        *previous-describe-output-override*)
+  (setf *previous-describe-output-override* nil))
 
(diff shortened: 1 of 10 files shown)


extensions/emacs-help-mode/main.lisp
```diff
@@ -0,0 +1,54 @@
...
+
+    (cond
+      (index
+       (setf *previous-ctrl-h-suffix* (prefix-suffix (nth index prefixes)))
+       (setf (prefix-suffix (nth index prefixes)) *ctrl-h-keymap*)
+       (setf *previous-describe-output-override*
+             lem-core/commands/help:*describe-output-type-override*)
+       (setf lem-core/commands/help:*describe-output-type-override* :buffer))
+      (recursed-p
+       (error "Infinite loop"))
```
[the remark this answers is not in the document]
VisenDev: /dismiss this error shouldn't be user triggerable, and should only be triggered in the event of a bug in the code

src/commands/help.lisp
```diff
@@ -84,6 +119,10 @@
         (print-modes "Major modes" major-modes)
         (print-modes "Minor modes" minor-modes)))))
 
+(define-command describe-all-modes () ()
```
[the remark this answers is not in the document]
VisenDev: /dismiss Documentation has been added

src/commands/help.lisp
```diff
@@ -4,23 +4,58 @@
...
 
 (define-key *global-keymap* "C-x ?" 'describe-key)
 
+(declaim (type (member nil :buffer :popup :message)
+               *describe-output-type-override*))
+(defvar *describe-output-type-override* nil
+  "When non nil, describe commands will always
+   send their information to this type of output")
+
+(defmacro with-describe-output-stream
```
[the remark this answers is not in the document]
VisenDev: /dismiss This code has been refactored to use a `call-with-*` helper

VisenDev:
@vindarel Could you give me some assistance on how to globally rebind `C-h`, currently I am just replacing the binding in the global map, but would adding a minor mode keymap accomplish the same thing?


EDIT: I just decided to try using minor-mode-keymap and see if that did what I wanted, and it seemed to do so.

src/commands/help.lisp
```diff
@@ -4,23 +4,67 @@
...
            :describe-mode
            :apropos-command
+           :describe-command
+           :apropos-variable
+           :describe-variable
            :lem-version
-           :list-modes)
+           :list-modes
+           :describe-all-modes
+           :*describe-output-type-override*)
```
vindarel, line 12: Is it really necessary (for end users) to export this variable? You can access it with `::`.

src/commands/help.lisp
```diff
@@ -4,23 +4,67 @@
...
   (:lock t))
 (in-package :lem-core/commands/help)
 
 (define-key *global-keymap* "C-x ?" 'describe-key)
 
+(declaim (type (member nil :buffer :popup :message)
+               *describe-output-type-override*))
+(defvar *describe-output-type-override* nil
+  "When non nil, describe commands will always
+   send their information to this type of output")
```
vindarel: We welcome example choices in the docstring: `:popup` etc.

src/commands/help.lisp
```diff
@@ -111,24 +159,83 @@
...
+(define-command describe-command () ()
+  "Alias for apropos-command"
+  (call-command 'apropos-command nil))
+
+(defun lem-symbol-p (symbol)
+  (let* ((pkg (symbol-package symbol))
+         (name (package-name pkg)))
+    (string-equal "LEM" (subseq name 0 3))))
+
+(defun is-variable-p (symbol)
```
vindarel: needs "lem" in name like `lem-variable-p`.

src/commands/help.lisp
```diff
@@ -111,24 +159,83 @@
...
+  "Returns true if the symbol is a lem variable or an editor variable"
+  (or (lem/common/var:editor-variable-p (get symbol 'editor-variable))
+      (and (lem-symbol-p symbol)
+           (boundp symbol)
+           (not (constantp symbol)))))
+
+(defparameter *all-variables-cache* nil
+  "Cached list of all strings that represent variables")
+(defun all-variables ()
```
vindarel: plz a docstring

src/commands/help.lisp
```diff
@@ -111,24 +159,83 @@
...
+           :when (is-variable-p sym)
+           :collect (string-downcase
+                     (format nil "~a:~a"
+                            (package-name (find-package pkg))
+                            (symbol-name sym))))
+      :into syms
+      :finally (setf *all-variables-cache* syms)))
+  *all-variables-cache*)
+
+(define-command apropos-variable () ()
```
vindarel: missing docstring

(and also misleading name?)

src/commands/help.lisp
vindarel: This is also about lem variables, so it needs a clearer name.

src/commands/help.lisp
```diff
@@ -4,23 +4,67 @@
...
+     (with-pop-up-typeout-window (stream (make-buffer buffer-name) :erase t)
+       (funcall function stream)))
+    ((:buffer)
+     (let ((stream (make-string-output-stream)))
+       (unwind-protect
+            (funcall function stream)
+         (let ((output-buffer (make-buffer buffer-name)))
+           (erase-buffer output-buffer)
+           (insert-string (buffer-point output-buffer) (get-output-stream-string stream))
+           (pop-to-buffer output-buffer)))))))
```
vindarel: needs the `t` fall-through case, showing a warning message to the user (`message` or similar).

src/commands/help.lisp
vindarel: If you do mean Lem users can use it (to have help text in buffers rather than popups for instance), I'd push for a better name.

Also, what about giving prefix arguments to the interactive commands, to change the output type?

src/key.lisp
```diff
@@ -86,6 +86,17 @@
        (eq (key-shift key) shift)
        (equal (key-sym key) sym)))
 
+(defun key-equal (key-one key-two)
+  "Returns t if key-one == key-two, nil otherwise"
```
vindarel: Check but you should be able to use `equalp`

src/commands/help.lisp
```diff
@@ -111,24 +159,83 @@
...
+
+(defun is-variable-p (symbol)
+  "Returns true if the symbol is a lem variable or an editor variable"
+  (or (lem/common/var:editor-variable-p (get symbol 'editor-variable))
+      (and (lem-symbol-p symbol)
+           (boundp symbol)
+           (not (constantp symbol)))))
+
+(defparameter *all-variables-cache* nil
+  "Cached list of all strings that represent variables")
```
vindarel: It's also a bad name (it's specific to lem variables),

but I'm afraid of a cache (you know, "2 difficult problems in computer science: naming, caching and off-by-one errors).

What happens in long-lived sessions, when we compile and load Lem code on the fly? This cache would need to be invalidated. I'm afraid it's a bad good idea. 

Is iterating over all packages and symbols even slow?

I assume the goal of this is to offer auto-completion. I wonder if there's a better way.

WDYT?

src/commands/help.lisp
VisenDev: I meant for it to be customizable for lem users yes. If you have any better name suggestions I would be happy to hear them, I had a lot of trouble naming this variable.

As to prefix arguments, I'm not sure what would be involved for that, do you mean prefix arguments like how you can do `C-u M-x foo` in emacs to give extra arguments for the command?

src/commands/help.lisp
VisenDev: Can do

src/commands/help.lisp
VisenDev: can do

src/commands/help.lisp
VisenDev: Added docs and renamed it to `describe-variable`

src/commands/help.lisp
VisenDev: Renamed

src/key.lisp
VisenDev: I didn't realize this was a struct, good point

src/commands/help.lisp
VisenDev: The goal is to offer auto-completion yes. I'll do some testing and see how the performance is without caching

VisenDev: Okay I removed the cache, generating the list of variables takes about 0.003 seconds on my (admittedly quite fast) pc, so its probably not an issue, even on less powerful machines.

src/commands/help.lisp
VisenDev: Added an otherwise clause displaying an editor-error

src/commands/help.lisp
vindarel: Hey this command is about lem variables, I suggest renaming with "lem" inside like `describe-lem-variable`.

src/commands/help.lisp
vindarel: not easy… `documentation-output` ?

yes I mean prefix arguments like this. The use could be: M-x describe-lem-variables uses the default documentation-output, say the popup. Now calling `C-u 1 M-x …` would use a `:buffer`. (maybe not a great idea, usability-wise)

Other idea: to have a command that changes the default output. For ex: `M-x set-documentation-output RET buf <TAB> -> buffer RET` -> changes it to :buffer for the current session. And the value could be serialized in Lem's config (a very easy mechanism I can point if you don't know it yet). WDYT?

VisenDev: Yes a `set-documentation-output` command sounds like a good idea. I am not familiar with the serialized config mechanism - where should I look to find how to use it?

src/commands/help.lisp
```diff
@@ -4,23 +4,70 @@
...
 (in-package :lem-core/commands/help)
 
 (define-key *global-keymap* "C-x ?" 'describe-key)
 
+(declaim (type (member nil :buffer :popup :message)
+               *describe-output-type-override*))
+(defvar *describe-output-type-override* nil
+  "When non nil, describe commands will always
+   send their information to this type of output.
+   Acceptable values: (nil :buffer :popup :message) ")
```
vindarel, line 24: This sentence that appears in a function below is very useful to have here: " allowing the user to,
   for example, always output documentation descriptions to a full buffer, rather
   then just a temporary message or popup"

src/commands/help.lisp
vindarel: We can drop the new command and the prefix argument, but let's have a more explicit parameter name, and hopefully shorter.

trying: 

```
describe-commands-interface
documentation-interface-type
documentation-interface
documentation-output-style
```

src/commands/help.lisp
VisenDev: Updated.

src/commands/help.lisp
VisenDev: Okay, using documentation-output-style for now

src/commands/help.lisp
VisenDev: Updated

vindarel:
Looks good. Merge conflicts now.

VisenDev:
> Looks good. Merge conflicts now.

Are you sure? It says "No conflicts with base branch" for me

vindarel:
> Are you sure? It says "No conflicts with base branch" for me

got it: I couldn't rebase, but I can "merge" or "squash and merge" without issues.

vindarel:
Thank you!

Idea: search and replace all `with-pop-up-typeout-window` with `with-describe-output-stream` ;)

PR #2252, opened 2026-06-26
Revert "Add hyperspec documentation lookup"

vindarel:
Reverts lem-project/lem#2250

since the commands exist already. And are documented.

https://lem-project.github.io/usage/common_lisp/#hyperspec-lookup

sorry for the noise.


VisenDev:
Oh my bad, I didn't know this. The `C-c C-d h` keybinding should still be added then

PR #2262, opened 2026-07-04
support for displaying images in webview

mahmoodsh36:


mahmoodsh36:
this introduces functionality to display in webview, the functionality exists in the sdl2 frontend but not in webview.

the functionality so far is very basic and only works properly with images that originate from attributes of virtual text that originates from an overlay, by virtual text i mean the :before-item/:after-item stuff that was introduced in #2230, because the cursor is meant to skip over those virtual items. and currently when the cursor goes over non-virtual images the buffer gets confused.

the core currently uses a uniform grid for displaying text which makes it hard to work with arbitrarily sized objects so we round image sizes to multiples of cell sizes.

selecting text with cursor doesnt work because images dont align with the buffer flow so the client doesnt properly resolve cursor click position relevant to the text grid. hoping to addess this soon.

mahmoodsh36:
example usage:
```lisp
(lem:define-command webview-image-demo () ()
  (let* ((image (namestring (asdf:system-relative-pathname :lem #p"resources/lem.png")))
         (attr (lem:make-attribute :plist (list :image image :pixel-width 300 :pixel-height 150)))
         (buffer (lem:make-buffer "*image-demo*"))
         (point (lem:buffer-point buffer)))
    (lem:switch-to-buffer buffer)
    (lem:erase-buffer buffer)
    (lem:insert-string point (format nil "before X after~%more text~%"))
    (lem:with-point ((s point)
                     (e point))
      (lem:buffer-start s)
      (lem:character-offset s 7)
      (lem:move-point e s)
      (lem:character-offset e 1)
      (let ((ov (lem:make-overlay s e attr)))
        (lem:overlay-put
         ov
         :cursor-enter-functions (list 'lem:move-point-out-of-overlay))))
    (lem:buffer-start point)))
```

mahmoodsh36:
this PR contains more than just 'image support for webview', i unified some of the rendering/layout code between the different frontends. the webview doesnt use a grid of uniform cells anymore, and it mainly uses pixels for sizing things instead of cells, like the sdl2 frontend does.

this PR is a follow-up for #2230

mahmoodsh36:
current issues:
- sometimes the resolved column for a click on a specific row is miscomputed
- overlay nesting with different visibility settings is broken. for example when i tried to fold an org-block with latex previews in it, things didnt work as expected. we need to take into account overlays that arent necessarily disjoint that may have different visibility settings.

i think i will come back to this and fix things and perhaps rewrite some of the work since i think things can be made more ideal

PR #2263, opened 2026-07-05
Add advice to mark-set command for multicursor selection 

Plipus:
Hello,
While using lem to learn common lisp, I noticed that marks doesn't seem to work with multicursor.
Adding this advice seems to fix the issue, but as I’m still exploring both the codebase and Common Lisp,  I might have overlooked something.
Is there are any implications or edge cases I should be aware of?

### DIFF
--- a/src/commands/mark.lisp
+++ b/src/commands/mark.lisp
@@ -12,13 +12,13 @@
 (define-key *global-keymap* "C-x C-x" 'exchange-point-mark)
 (define-key *global-keymap* "C-x h" 'mark-set-whole-buffer)
 
-(define-command (mark-set (:advice-classes editable-advice)) (p) (:universal-nil)
+(define-command mark-set (p) (:universal-nil)
   "Sets a mark at the current cursor position."
   (if p
       (move-point (current-point) (pop-buffer-point))
       (progn
         (set-cursor-mark (current-point) (current-point))
-        (message "Mark set"))))
+        (when-real-cursor (message "Mark set")))))
 
 (define-command exchange-point-mark () ()
   "Exchange the current cursor position with the marked position."
--- a/src/commands/mark.lisp
+++ b/src/commands/mark.lisp
@@ -12,7 +12,7 @@
 (define-key *global-keymap* "C-x C-x" 'exchange-point-mark)
 (define-key *global-keymap* "C-x h" 'mark-set-whole-buffer)
 
-(define-command mark-set (p) (:universal-nil)
+(define-command (mark-set (:advice-classes editable-advice)) (p) (:universal-nil)
   "Sets a mark at the current cursor position."
   (if p
       (move-point (current-point) (pop-buffer-point))
--- a/src/cursors.lisp
+++ b/src/cursors.lisp
@@ -123,3 +123,7 @@
                     (if not-lastp
                         (write-line string out)
                         (write-string string out))))))))
+
+(defmacro when-real-cursor (&body body)
+  `(unless (typep (current-point) 'fake-cursor)
+     ,@body))
--- a/src/cursors.lisp
+++ b/src/cursors.lisp
@@ -125,5 +125,7 @@
                         (write-string string out))))))))
 
 (defmacro when-real-cursor (&body body)
+  "Execute BODY only if the current cursor is a real one (i.e. not a fake-cursor)."
   `(unless (typep (current-point) 'fake-cursor)
      ,@body))
+
--- a/src/internal-packages.lisp
+++ b/src/internal-packages.lisp
@@ -621,7 +621,8 @@
    :make-fake-cursor
    :delete-fake-cursor
    :merge-cursor-killrings
-   :clear-cursors)
+   :clear-cursors
+   :when-real-cursor)
   ;; typeout.lisp
   (:export
    :*typeout-mode-keymap*


src/commands/mark.lisp
```diff
@@ -12,7 +12,7 @@
 (define-key *global-keymap* "C-x C-x" 'exchange-point-mark)
 (define-key *global-keymap* "C-x h" 'mark-set-whole-buffer)
 
-(define-command mark-set (p) (:universal-nil)
+(define-command (mark-set (:advice-classes editable-advice)) (p) (:universal-nil)
```
theangelperalta: This is redundant: `mark-set` already has an `execute :around` doing `process-each-cursors` at line 38-41, and `editable-advice` adds the same wrapper — so this nests two `process-each-cursors` passes. In a headless test it didn't change the multicursor result (still crashes in the per-cursor `(message "Mark set")` below).

theangelperalta (commented):
Thanks for digging into this! I looked into it and I don't think this change is the right fix — I believe it's redundant, and the real bug is elsewhere.

`mark-set` already wraps itself in `process-each-cursors` via its own `execute :around` method (`src/commands/mark.lisp:38-41`, present since 2023). `editable-advice` supplies that *same* `process-each-cursors` wrapper, so this PR just stacks a second, nested one.

I ran it headless (fake-interface) to check:
- On `main`, marks are already set for every cursor — `mark-set` with 3 cursors returns `((CURSOR T) (FAKE-CURSOR T) (FAKE-CURSOR T))`.
- The multicursor failure is actually a **crash in `(message "Mark set")`**: it's called once per cursor inside `process-each-cursors` with the buffer point rebound to a fake cursor, and the `:follow-cursor` popup positioning then compares points across buffers → `%always-same-buffer` assertion. Dropping the `message` call makes it pass.
- Applying this PR did **not** stop that crash in my repro.

So the fix is probably to stop calling `message` per-cursor — e.g. move it out of the command body so it runs once after all cursors are processed. Could you share your exact repro (frontend + keys)? I wasn't able to find a case where adding `editable-advice` changes behavior.

Plipus:
Thanks for your response !
After another try this afternoon, it seems that I can't reproduce the bug on the main branch...
However, after multiple additional attempts, I did encounter a crash. I think my bug fix was just an illusion caused by insufficient testing from my side, sorry.

I will try to fix the message problem, moreover I have another branch where I try to fix multicursor for vi-mode, and I have the same kind of problem with part of commands that can't by runned multiple times (With my fix, `d` with multiple cursors will ask a motion for each cursor, instead of applying the same motion for every cursor).

But extracting the point related stuff in another command will expose two commands in M-x, the wrapper and the subcommand, it's not a beautiful solution. Or can we hide a command from M-x ?

I see that there is a *post-command-hook*, can it be the solution for extracting the "run only once" part ?
But I the command must be aware that an advice can be applied and adapt its implementation, it's removing the beauty of the advice in my opinion.
I'm new in common lisp, is there a pattern that I can apply in this situation ?

theangelperalta:
No worries at all — thanks for confirming, and these are good questions. A few thoughts:

**Hiding a command from M-x / avoiding two commands.** You don't need a second *command* — only `define-command` registers into the M-x table (`register-command` → `add-command`, unconditionally), and there's no hidden/visible flag. So the idiom is to factor the internal piece into a plain `defun` and keep a single command. The advice already gives you the per-cursor loop, so a helper doesn't need to be a command.

**The "run once" part (the `message`).** The body re-runs once per cursor because `process-each-cursors` calls it for each fake cursor *plus once for the real cursor*. You can use the point type as the discriminator:

```lisp
(unless (typep (current-point) 'fake-cursor)
  (message "Mark set"))
```

On the fake-cursor passes `(current-point)` is the fake cursor; only the final real pass isn't — so the message fires exactly once, and it also fixes the popup crash since `:follow-cursor` then anchors to the real cursor.

**`*post-command-hook*`.** It does run exactly once (it's in `call-command`, outside the `process-each-cursors` `:around`), so it is technically a "run once" spot. But it's a global hook — the command would have to register it and check `(this-command)`, which is more machinery and less local than the guard above. I'd keep it for cross-cutting concerns rather than per-command output.

**The vi-mode `d`-prompts-per-cursor problem.** That's the harder, general version: an argument/prompt should be read once and reused. Arguments are read inside the per-cursor body, so naive prompting re-prompts. The pattern is to read the motion **once, before iterating cursors**, then apply the captured value with `do-each-cursors`. (This is roughly what Emacs `multiple-cursors` does: run for the real cursor first, record its input, then replay the recorded input for the fake cursors so they don't re-prompt.) Note that vi-mode doesn't use `process-each-cursors` at all today, so you have room to structure it that way from the start.

Happy to take a look at the vi-mode branch if you push it as draft PR

Plipus:
I was thinking about second command because I wanted to re-use existing advice, but it's clear that it's not a good solution.
I rewrote the PR with the cursor type check, it's seems to be the simpler solution. And as I will probably use it for the vi-mode, I extract it in a macro

theangelperalta (commented):
Right scope — each cursor still gets its mark, only the echo runs once. Fixes both the duplicate message and the `:follow-cursor` crash. Load order checks out (`cursors` → `command-advices` → `commands/mark`), so the macro compiles before use.

Two nits: add a trailing newline to `cursors.lisp`, and a docstring on the now-public macro.

Heads up for vi-mode reuse: this is a "run once" primitive — the `d`-prompts-per-cursor case needs input read once and *replayed*, which this won't cover. 👍

Plipus:
Thank you for your response, I applied your recommandations !

vindarel:
Is this ready to be merged?

PR #2271, opened 2026-07-21
Tramp for Lem

MasterCsquare:
# lem-tramp

Transparent remote file editing for Lem. Works like
[Emacs TRAMP](https://www.gnu.org/software/tramp/) — type a remote path
into `C-x C-f` and edit the file as if it were local.

## Supported Methods

| Method | Syntax | Description |
|--------|--------|-------------|
| `ssh` | `/ssh:user@host:/remote/path` | Edit files on a remote host via SSH |
| `sudo` | `/sudo::/local/path` | Edit local files with root privileges via sudo |

## Usage

Open a file with `C-x C-f` using the TRAMP path syntax:

```
C-x C-f /ssh:root@example.com:/etc/nginx/nginx.conf
C-x C-f /sudo::/etc/hostname
```

The remote file is loaded into a buffer.  Edit normally, then save with `C-x C-s`.

### SSH (`/ssh:`)

**Key-based auth** (recommended) — no prompt, works automatically if you have
SSH keys set up and your key is in the remote host's `authorized_keys`.

**Password auth** — requires the `sshpass` utility:

```bash
# Arch
sudo pacman -S sshpass
# Debian/Ubuntu
sudo apt install sshpass
```

When key auth is unavailable, a password prompt appears in the minibuffer.
The password is cached for the session; subsequent file operations on the
same host reuse it without re-prompting.

### Sudo (`/sudo::`)

A password prompt appears if your sudo timestamp has expired (typically
5-15 minutes since your last `sudo` invocation in a terminal).  If you
have passwordless sudo configured (`NOPASSWD` in sudoers), no prompt is
shown.

## How It Works

lem-tramp hooks into Lem's virtual filesystem layer:

- **Reading** — runs `cat /remote/path` via SSH (or sudo), loads the
  output directly into an in-memory buffer.
- **Writing** — pipes the buffer content through `cat > /remote/path`.
- **Directory listing** — uses `ls -1a` for completion and directory-mode.
- **Metadata** — uses `stat -c '%s %Y'` for file size and modification time.

No temporary files are created on either the local or remote side.

## Dependencies

- **sshpass** — only needed for SSH password authentication
- **flexi-streams**, **str**, **babel**, **ppcre** — Common Lisp libraries

## Performance

SSH connections use `ControlMaster` multiplexing — the first command
establishes the TCP connection, and all subsequent commands reuse it.
A 5-second filesystem cache eliminates redundant `test -d` / `test -f` /
`stat` calls when Lem probes the same path multiple times during a single
file-open operation.

### DIFF
--- a/extensions/tramp/tramp.lisp
+++ b/extensions/tramp/tramp.lisp
@@ -175,8 +175,14 @@ Marks connection as needing password and prompts."
   (or (tramp-get-password method user host)
       (ecase method
         (:sudo
-         (or (tramp-prompt-password method user host)
-             (error 'editor-abort)))
+         (if (eql 0 (nth-value 2
+                       (uiop:run-program '("sudo" "-n" "true")
+                                         :output nil
+                                         :error-output nil
+                                         :ignore-error-status t)))
+             nil  ;; passwordless sudo
+             (or (tramp-prompt-password method user host)
+                 (error 'editor-abort))))
         (:ssh
          ;; Lazy auth: authenticated on first actual command, not here
          nil))))
--- a/extensions/tramp/tramp.lisp
+++ b/extensions/tramp/tramp.lisp
@@ -253,42 +253,34 @@ Exit code 5   = sshpass incorrect password."
   "Run an SSH command with automatic auth handling.
 Tries key auth first; on auth failure, clears cached password and retries.
 Signals editor-error if authentication ultimately fails."
-  (flet ((run-with-password (pwd)
-           (let ((args (build-ssh-args :ssh user host command :password pwd)))
-             (%ssh-run user host args))))
+  (labels ((run-with-password (pwd)
+             (let ((args (build-ssh-args :ssh user host command :password pwd)))
+               (%ssh-run user host args)))
+           (run-with-retry (pwd)
+             (multiple-value-bind (ec out) (run-with-password pwd)
+               (if (ssh-auth-failure-p ec)
+                   (let ((new-pwd (progn (clear-password :ssh user host)
+                                         (prompt-password :ssh user host))))
+                     (if new-pwd
+                         (multiple-value-bind (ec2 out2) (run-with-password new-pwd)
+                           (if (ssh-auth-failure-p ec2)
+                               (editor-error "Authentication failed for /ssh:~@[~A@~]~A"
+                                             (or user "") host)
+                               (values ec2 out2)))
+                         (error 'editor-abort)))
+                   (values ec out)))))
     (multiple-value-bind (password auth-tried) (ssh-ensure-auth :ssh user host)
       (if auth-tried
-          ;; Auth method known — run directly
-          (multiple-value-bind (exit-code stdout) (run-with-password password)
-            (if (ssh-auth-failure-p exit-code)
-                ;; Cached password is wrong — re-prompt and retry once
-                (let ((new-pwd (progn (clear-password :ssh user host)
-                                      (prompt-password :ssh user host))))
-                  (if new-pwd
-                      (multiple-value-bind (ec2 out2) (run-with-password new-pwd)
-                        (if (ssh-auth-failure-p ec2)
-                            (editor-error "Authentication failed for /ssh:~@[~A@~]~A"
-                                          (or user "") host)
-                            (values ec2 out2)))
-                      (error 'editor-abort)))
-                (values exit-code stdout)))
-          ;; Auth method unknown — try key auth first
-          (multiple-value-bind (exit-code stdout)
-              (run-with-password nil)
-            (if (ssh-auth-failure-p exit-code)
-                ;; Auth failure → prompt password and retry
+          (run-with-retry password)
+          (multiple-value-bind (ec out) (run-with-password nil)
+            (if (ssh-auth-failure-p ec)
                 (let ((pwd (ssh-remember-auth-failure user host)))
                   (if pwd
-                      (multiple-value-bind (ec2 out2) (run-with-password pwd)
-                        (if (ssh-auth-failure-p ec2)
-                            (editor-error "Authentication failed for /ssh:~@[~A@~]~A"
-                                          (or user "") host)
-                            (values ec2 out2)))
-                      (values exit-code stdout)))
-                ;; Key auth succeeded or command failed for other reasons
+                      (run-with-retry pwd)
+                      (values ec out)))
                 (progn
                   (ssh-remember-auth-success user host)
-                  (values exit-code stdout))))))))
+                  (values ec out))))))))
 
 ;;; Sudo command execution (pipe-based, no temp files)
 
--- a/extensions/tramp/tramp.lisp
+++ b/extensions/tramp/tramp.lisp
@@ -243,24 +243,47 @@ Marks connection as needing password and prompts."
     (error (c)
       (values 255 (princ-to-string c)))))
 
+(defun ssh-auth-failure-p (exit-code)
+  "Return T if EXIT-CODE indicates an SSH authentication failure.
+Exit code 255 = SSH BatchMode auth failure / connection refused.
+Exit code 5   = sshpass incorrect password."
+  (or (= exit-code 255) (= exit-code 5)))
+
 (defun %ssh-run-with-auth-retry (user host command)
   "Run an SSH command with automatic auth handling.
-Tries key auth first; on exit-255 failure, prompts for password and retries.
-Returns (values exit-code stdout-string)."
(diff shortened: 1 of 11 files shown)


extensions/tramp/tramp.lisp
```diff
@@ -0,0 +1,687 @@
+(defpackage :lem-tramp
+  (:use :cl :lem)
+  (:export :tramp-mode))
+(in-package :lem-tramp)
+
+(setf (documentation *package* t)
+      "TRAMP-like remote file editing for Lem.
+Supports /ssh:user@host:/path and /sudo::/path syntax for transparent
+remote file access via SSH and sudo.
```
vindarel, line 13: This looks strange to me. Why not a `(:documentation "…")` clause in the defpackage?

extensions/tramp/tramp.lisp
```diff
@@ -0,0 +1,687 @@
...
+:password (needs password), or nil (unknown).")
+
+(defun sshpass-available-p ()
+  "Check if the sshpass utility is available on the system."
+  (or (eql 0 (nth-value 2
+                        (uiop:run-program '("which" "sshpass")
+                                          :output nil
+                                          :error-output nil
+                                          :ignore-error-status t)))
+      (probe-file "/usr/bin/sshpass")
```
vindarel: this is unix only, please use a feature flag.

extensions/tramp/tramp.lisp
```diff
@@ -0,0 +1,687 @@
...
+
+(defvar *ssh-auth-method-cache* (make-hash-table :test 'equal)
+  "Cache of SSH auth methods. Values: :key (key auth works),
+:password (needs password), or nil (unknown).")
+
+(defun sshpass-available-p ()
+  "Check if the sshpass utility is available on the system."
+  (or (eql 0 (nth-value 2
+                        (uiop:run-program '("which" "sshpass")
+                                          :output nil
```
vindarel: `which` needs to be installed and that's to be taken care of. There's a Lem primitive for this "wich" kind of checks, forgot its name right now. It's also unix only.

extensions/tramp/tramp.lisp
```diff
@@ -0,0 +1,687 @@
...
+  (list "-o" "ControlMaster=auto"
+        "-o" "ControlPath=/tmp/lem-ssh-%C"
+        "-o" "ControlPersist=60"))
+
+(defun build-ssh-args (method user host command &key (use-sudo-s nil) password)
+  "Build the argument list for running a command via SSH or sudo."
+  (ecase method
+    (:ssh
+     (let ((target (if user (format nil "~A@~A" user host) host))
+           (cmd-args (if (listp command) command (list command)))
```
vindarel: shorter: `uiop:ensure-list`

extensions/tramp/tramp.lisp
```diff
@@ -0,0 +1,687 @@
...
+                        files)))
+              (unless filtered
+                (return-from tramp-file-completion nil))
+              (mapcar (lambda (f)
+                        (let ((label (enough-namestring (namestring f) input-dir)))
+                          (lem/completion-mode:make-completion-item
+                           :label (or label (namestring f)))))
+                      filtered))))
+        (funcall *tramp-original-completion-function*
+                 string directory :directory-only directory-only))))
```
vindarel, line 759: surely this function can be split in smaller parts? (to have less indentation levels and nested conditionals, etc)

src/buffer/file-utils.lisp
```diff
@@ -91,11 +105,14 @@
      (sort-files files))))
 
 (defun directory-files (pathspec)
-  (if (uiop:directory-pathname-p pathspec)
-      (list (pathname pathspec))
-      (or (mapcar (lambda (x) (virtual-probe-file x pathspec))
-                  (directory pathspec))
-          (list pathspec))))
+  (or (loop :for f :in *virtual-directory-files-functions*
```
vindarel: This loop could be a function with a meaningful name? 

plus it would save the copy-pasting of below.

src/buffer/file-utils.lisp
```diff
@@ -142,6 +165,38 @@
 
 (defparameter *virtual-file-open* nil)
 
+(defparameter *virtual-probe-file-functions* nil
+  "A list of functions for virtual probe-file.
```
vindarel, line 190: Can you add context what this is used for? For a new contributor reading this file-utils and not knowing abous lem-tramp it could be puzzling.

vindarel:
Thanks for the PR!

Can you disclose your LLM usage and methodology? 

How long have you been using it, how much tested is it?

Your PR message is useful, can you add it to a README next to the sources?

So the functionality is unix only right? We'll need a feature flag somewhere, to error gracefully for Windows users.

Is the name Tramp appropriate? Since this module isn't a copy, and Lem doesn't aim at being an Emacs copy either. Honest question.

In general I think all the variable and functions names in `tramp-…` feel too elisp-y. We have packages in CL. 

---

It would be great to have more testers before merging! (I can't test right now for the next few days)

Thanks!

MasterCsquare:
Thanks for the review, I will fix them later.

For LLM, I'm using claude + deepseek. Because deepseek is not the best(compared to chatgpt and Fable, but cheap enough), i have to guide him more, which helps me understanding his code better. LLM is just like a man playing three role at the same time, teacher, friend, slave. Teacher for he known so many things and can explain it in detail. Friend for he can help you do many things. Slave for he can get no tired at all. Almostly my workflow for LLM is simple, i do it step by step to make the  codebase larger until the first working version done. After that ,test by myself and fix the bugs. For performance improving , i ask him to analyze if it possible to improve, and check what he said, when everything he said is right, i ask him to implement it.

I only using it for about 2 days, testing is not so complete.

README will be added later.

Right, it only for Unix.

I think Tramp is a suitable name, for nothing related to emacs in this name, and can help the new incoming user to understand what it is faster.

Maybe the LLM read the emacs codes, so it's elisp-y, fix later.

Test is required, currently when input a wrong password, the feedback is not normal, there are still some bugs.

extensions/tramp/tramp.lisp
MasterCsquare: Because ANSI Common Lisp's defpackage doesn't have a :documentation option in the standard.

The standard defpackage options are only: :export, :import-from, :shadow, :shadowing-import-from, :use, and :nicknames. No :documentation slot exists.

So the portable way to attach a docstring to a package is to set it after the defpackage form:

(setf (documentation *package* t) "docstring here")

Some implementations (like SBCL) extend defpackage with a non-standard :documentation option, but Lem targets portability, so the setf documentation approach is the consistent convention across the project.

MasterCsquare:
All the Reviews and bugs have been fixed.

Testing is welcome, please report bugs if you found.

There's a typo in the commit messages, so --amend and force push.

MasterCsquare:
I found a new bug, the password prompted(sudo) still insert to the top of buffer, but you can't see it in the buffer but when you cat the file you see it. This should have been fixed before, but it's just in the situation fixing by you can't see.

but I can't reproduce this, it seems to the files opened is opened before in the buggy version, so there is the password in the head, when i open the file i never open before it works as normal.This should be noticed.

MasterCsquare:
The bug is fixed, can be reproduced by open a new file like`C-x C-f /sudo::/etc/example` and input the password for sudo,  write something then save，and then `cat /etc/example`.

MasterCsquare:
The Terminal is integrated in the new commits.

When a buffer is visiting a TRAMP path, `M-x terminal` opens a terminal in the remote file's directory on the appropriate host.This works like run `eshell` in emacs tramp but for lem. Both sudo and ssh are supported.

This make `M-x terminal` more useful.

vindarel:
Very nice.

I'm still short on time to test and report (enjoying days off and moving flats…). 

Anyone's welcome to report.

We can also let you mature the module even more ;)

MasterCsquare:
There are about 70 clones for this PR, but nobody has reported any bugs, maybe no bug (hopefully).I also haven't found any bug since the last commit.

Next goals(following PRs):
- password masking
- integrate `directory-mode`

vindarel:
It worked fine for me in my few minutes test. 

I would love to see a directory content. Would be great to merge at the same time. WDYT?

Well done anyways.

---

I think a top-level command starting with "tramp-" in its name will very much improve discoverability. It's easy enough, can you do it? Maybe sth like: `M-x tramp-connect-over-ssh RET IP? … RET user? RET path?… RET`

MasterCsquare:
maybe implement it in next PR, need to modify so many codes for support directory-mode and need to test so many things for the mode's command.

I think we make a general interactive command is better, something like `tramp-connect`, and select the method(sudo/ssh) first them the other.

Issue #2090, opened 2026-01-09
Alt bindings ignore keyboard layout in webview


not-leader:
on fedora 43 cosmic spin, an installation of lem from source as of commit `d53a5ce` has an issue where the M- bindings ignore the user set keyboard layout

eg: M-x on colemak-dh registers as M-z (the equivalent qwerty input)

vindarel:
Hi, this is only on the web view, right? I have this issue, but it works fine on ncurses.

not-leader:
yes, this is the webview sorry

vindarel:
It works now on Linux 🥳

PR #2275, opened 2026-08-14
Avoid using key code with alt modifier except on MacOS

zakwilson:
When using Alt as a modifier key, Lem ignores the system keyboard layout and looks at [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) instead. This appears to be motivated by the Option key on MacOS sending something other than the base key.

This limits the special handling to MacOS, making the system keyboard layout work in combination with Alt on other platforms. This addresses #2090 for not-MacOS, but should not close the issue until MacOS is also fixed.


vindarel:
Thanks a lot! :partying_face:

PR #2276, opened 2026-08-18
fix incorrectly displayed file names in directory mode

mahmoodsh36:
in a directory buffer, a file named "12 Day of Rain [t0Tph__v070].webm"
was listed as "12 Day of Rain \[t0Tph__v070].webm"

this is fixed by making use of `uiop:native-namestring' instead of the builtin `namestring'.

fixing this revealed other issues because it made things error out, which is what the 2 other commits are for


vindarel:
Thank you. 

I'm all for very explicit comments, specially about those details… don't hesitate to add this explanation/example to display-name, or as a top-level recommendation/warning in directory-mode. It will help the next developers for sure.

PR #2281, opened 2026-08-30
Ansible-mode + lsp for yaml

Pythonpoison48:
This pr adds a mode for ansible via ansible-language-server . Ansible mode reuse the yaml mode .It also adds a lsp for yaml


vindarel:
Thank you! :rocket: 

I'm wondering about this:

      :depends-on ("lem/core" "lem-tree-sitter" "lem-lsp-mode")

if it isn't too much and might be annoying for simple yaml usage. WDYT?

Pythonpoison48:
Hello ,

I don't know lem enough but It should be possible to disable lsp-mode and tree-sitter in case someone do not want these features ?

If not I can add a switch/flag for this .

Have a nice day

vindarel:
thanks for the answer but please ignore my question, I read the changes too fast.

Thank you again!

PR #2287, opened 2026-09-09
Support a minimal ncurses-only build of lem with no extensions loaded

### LINKED ISSUE #2286: Allow building without baked-in extensions, modes, and file manager
Provide an option to build a "batteries-excluded" version of Lem (omitting language modes, built-in LSP servers, AI mode, file manager, etc.), as these are currently baked in with no way to exclude them.

Lem already follows a similar pattern with terminal support, which is opt-in via installing the `libvterm` dependency. Offering a build configuration to exclude default features would allow users to start with essential core editing functions and add tools as needed.

VisenDev:
Closes https://github.com/lem-project/lem/issues/2286

To build:
```
make minimal-build
```


VisenDev:
Binary size difference:

```
~/lisp/lem [1] $ du -sh lem-ncurses
117M	lem-ncurses
~/lisp/lem $ du -sh lem-minimal-build 
94M	lem-minimal-build
~/lisp/lem $ 
```

vindarel:
Is that your main motivation, binary size? and/or compile times?

---

The change sounds reasonable.

Please add a line of developer documentation somewhere (at least in the Makefile, to explain it doesn't build extensions/), also on https://lem-project.github.io/installation/ncurses/, thank you.

VisenDev:
> Is that your main motivation, binary size? and/or compile times?

I was just implementing the changes proposed by https://github.com/lem-project/lem/issues/2286. The binary size demonstration was only to verify that the minimal build does actually include less things than the full build does.

> Please add a line of developer documentation somewhere (at least in the Makefile, to explain it doesn't build extensions/), also on https://lem-project.github.io/installation/ncurses/, thank you.

Can do

PR #2288, opened 2026-09-10
fix freeze from interning mode class names during redisplay

mahmoodsh36:
lem used to sometimes freeze when trying to evaluate specific code snippets. this was caused by a deadlocking issue caused by trying to intern symbols unnecessarily.

before this fix, it was reproducible by simply trying to evaluate the expression ﻿`'lem-core::nonexistent-symbol`.


vindarel:
# :partying_face: 

Fantastic! 

## :pray:

PR #2289, opened 2026-09-10
Add `undefine-all-keys` function

### LINKED ISSUE #2285: Add a public way to clear all keybindings at once
I'd like to configure Lem's keybindings completely from scratch.

My main problem is that there doesn't seem to be a public operation to
clear all bindings from a keymap at once. The public API provides
`undefine-key` and `undefine-keys`, but those require removing bindings
individually.

The main thing I want is a single supported operation that removes all
existing bindings from a keymap, so I don't have to enumerate and
undefine every binding individually.

Currently, my configuration has to do roughly this:

    (clrhash (lem-core::keymap-table lem:*global-keymap*))
    (clrhash (lem-core::keymap-function-table lem:*global-keymap*))
    (setf (lem-core::keymap-parent lem:*global-keymap*) nil)

My actual configuration also does this for major/minor mode keymaps.
This relies on internal APIs and was written against an older keymap
implementation.

What I'd like to be able to do instead is simply:

    (clear-keymap lem:*global-keymap*)

and then define only the bindings I want.

Would it make sense to provide a public `clear-keymap` / `reset-keymap`
operation that safely removes all bindings?

Ideally this should make it possible to reset a keymap in one operation,
without having to know or depend on its internal representation.

VisenDev:
Closes https://github.com/lem-project/lem/issues/2285

I chose to name the function `undefine-all-keys` rather than the proposed `clear-keymap` so that the new function would match the naming conventions established by the already existing `undefine-key` and `undefine-keys` functions


vindarel:
Inviting @soiq1781-cell to test.

mahmoodsh36:
you probably also want `(setf (keymap-children keymap) nil)` if you really wanna clear the map completely, and even then a keymap can still "contain" some keybindings if it has a 'base' keymap so you might also need `(setf (keymap-base keymap) nil)`

its also possible to play around with the generics defined for `*root-keymap*` directly so you can have full control

PR #2291, opened 2026-09-11
Add delete-region command

### LINKED ISSUE #2277: Add 'delete-region' command
Delete region kills the selected region without copying it.

VisenDev:
I also added the "C-x u" emacs keybinding for `undo` because I noticed it was missing

Closes: https://github.com/lem-project/lem/issues/2277


vindarel:
Thanks

PR #2296, opened 2026-09-12
sdl2: don't run post-display-change hooks before the first frame

MasterCsquare:
this fix sdl2 error in linux, mac need to test
about how it fix, see the diff, it's short


vindarel:
thank you

PR #2297, opened 2026-09-12
Don't fold on tab by default

VisenDev:
Folding function definitions by default whenever you press tab is a new behavior that was just added in [this commit](https://github.com/lem-project/lem/commit/258aa2af4) about 2 months ago.

While some people might like this behavior, I found it very confusing and it caused friction when trying to use tab for symbol completion and code indentation.

Because the folding behavior is bound to such a fundamental key (`Tab`) which is already used for several other tasks, I think it would be better for this to be an opt-in feature, rather than enabled by default.


mahmoodsh36:
youre probably right, its supposed to fold only when the cursor is on the first line of a defun, but since i dont use completion much i may not have noticed the friction it caused

VisenDev:
> youre probably right, its supposed to fold only when the cursor is on the first line of a defun, but since i dont use completion much i may not have noticed the friction it caused

Actually maybe its also particularly problematic for me because the folding behavior seems to occur in places where I wanted different behaviors as well. 

For example, if I am defining a function, and I want to tab complete a symbol like `lem:define-command`, and I have my cursor after the `f` in `lem:def` and I press `Tab`, the function just folds, rather than doing symbol completion.

```lisp
(defun foo () (lem:def))
 ;; CURSOR HERE      ^
```

Another thing I like to do when working on code is using `C-M-u` to keep going up a sexpr until my cursor is at the beginning of my function, then doing `C-M-Space` to select the entire function definition, then pressing `Tab` to auto-indent the entire function body to align any code blocks which may have ended up improperly indented while editing. In This following example, that workflow no longer works, when I select the body to try and fix the incorrect indentation of the let form, it just folds the function definition when I press tab.

```lisp
(defun bar ()
  (let ((asdf
     1))
    (print asdf)))
```


I think if these problems were fixed, I think the folding behavior would be less disruptive. But as it is, it seems to break these tasks

vindarel:
oh indeed.

Can we bind folding to another key? `C-Tab` ?

VisenDev:
> Can we bind folding to another key? `C-Tab` ?

What is the preferred approach @vindarel @mahmoodsh36 ? Should I disable tab folding by default or change the default folding key so it doesn't conflict with the other tab actions? I think given the problems that the current implementation of the tab folding has that it might be better to use a different key like `C-Tab` as suggested

vindarel:
Hey shall you bind folding to `C-tab` and let's merge this?

I'm annoyed by folding when the cursor is at the end of line too^^ (right before the last paren, where I want code completion)

Issue #1546, opened 2024-09-27
Directory Sorting Order?


nmccamish:
Currently, one can sort a directory by one of the three members of type `sort-method`, but there currently doesn't appear to be a way to sort it ascending or descending. Perhaps we could add a type, `sort-direction`, and implement it? It could be as simple as changing the `:test` from `#'>` to `#'<`.

vindarel:
Hello, correct. I think we could have the capital `S` key for that.

but wait, for what exactly, let's think about usage carefully:

- "s" sorts successfully with a different method.
- would "S" simply invert the current order?

what about:
- current order starts at "ascending"
- "s" sorts
- "S" toggles the sort order to "descending" and changes the current file order. 
- a repeated press to "S" only inverts the current file order.
- "s" sorts with the next method, respecting the current sord order.

or is "S" better suited to sort with the previous method, and we use another key to inverse the ordering? :thinking: 

what are other applications doing? (file manager or web app listing data, etc)

nmccamish:
In Doom Emacs, I press `S` (`dirvish-quicksort`), which pulls up a [menu of options](https://github.com/hlissner/dirvish/blob/5f046190e886fb0a2dae7e884cc7cd9bcf48ac26/extensions/dirvish-ls.el#L98-L135), with the lower-cased keys being one direction, and the upper-cased versions the other. Perhaps we could do something like that? Do we need a separate `sorting-direction` type & argument, or do we integrate that into `sorting-method`, or something else?

vindarel:
I think it's a good UX. 

We don't have a text-based menu of options, we could shave the yaks and create it beforehand, or in the meantime we could rely on the prompt commands (`prompt-for-string` etc), in order to autocomplete a choice.

nmccamish:
So would it be good to create a separate `sorting-direction`, or merge that into `sorting-method` like Dirvish does?

cxxxr:
> We don't have a text-based menu of options, we could shave the yaks and create it beforehand

What do you think of this?
https://github.com/lem-project/lem/blob/0f178254ea848b29d6f40403e79a8fd90419ef6f/src/ext/context-menu.lisp#L76

vindarel:
oh indeed it looks useful +1

vindarel:
> So would it be good to create a separate `sorting-direction`, or merge that into `sorting-method` like Dirvish does?

not 100% sure what you mean, but giving an option to the user to choose the default sorting direction is always nice.

[...]

PR #2306, opened 2026-09-16
add directory-mode sort keymap

mahmoodsh36:
this addresses https://github.com/lem-project/lem/issues/1546

this made me realize there is an issue: we might want to move the transient module to the core instead of having it as an extension because directory-mode ideally shouldve used it directly but it cant depend on it because its not a core module.

while this functionality was possible without needing to depend on the transient module (because the transient module just deals with keymaps and we just created the keymap it needs), it would've been more ideal to do this with the `define-transient` macro which is only available through the transient module and not through the core keymap functionality.


vindarel:
Oh yes, this transient :+1: 

I see a change in use: before, pressing `s` was sorting right away (by size), and a subsequent press was sorting by the next sort method. Now, pressing `s` does nothing but waits for a second press, the choice. 

I suggest keeping one direct action with `s` (either this default, either alternate size and name) and have this keymap on (capital) `S`.

WDYT? I think it's nice to have direct actions on a key, and another key with a menu.

mahmoodsh36:
having a second binding for this makes sense. tho the transient popup is supposed to show up instantly when you press 's'. im not sure why a delay was introduced by default for the transient popup (wasnt done by me).

i have this in my config to undo the upstream change
```
(setf lem/transient:*transient-popup-delay* 0)
```

Issue #2272, opened 2026-07-25
Cannot edit markdown files: "The alien function "ts_tree_root_node_out" is undefined."


Baisha-Swap:
Hi! I'm a new lem user and  have a problem with editing markdown files with it.
I use Arch, so I followed "Install from scratch" section in https://lem-project.github.io/installation/ncurses/. Then, installation succeeded.
However, I realized lem fails trying to edit xxxx.md.

```
❯ lem a.md
The alien function "ts_tree_root_node_out" is undefined.
Backtrace for: #<SB-THREAD:THREAD tid=3904757 "editor" RUNNING {120493A2C3}>
0: ("undefined function")
1: (TREE-SITTER/FFI:TS-TREE-ROOT-NODE #.(SB-SYS:INT-SAP #X7FD6AC016F50) #.(SB-SYS:INT-SAP #X7FD6AC00BA80))
2: (TREE-SITTER/NODE:TREE-ROOT-NODE #<TREE-SITTER/TYPES:TS-TREE {1204E25AE3}>)
3: (LEM-TREE-SITTER::APPLY-TREE-HIGHLIGHTS #<LEM-TREE-SITTER:TREESITTER-PARSER {1204B47C93}> #<TREE-SITTER/TYPES:TS-TREE {1204E25AE3}> #<LEM/BUFFER/INTERNAL:TEXT-BUFFER a.md /home/baisha/a.md {12047F5C23}> #<LEM/BUFFER/INTERNAL:POINT (1, 0) "" {1204DB66F3}> #<LEM/BUFFER/INTERNAL:POINT (1, 0) "" {1204DB6743}>)
4: (LEM/BUFFER/INTERNAL:SYNTAX-SCAN-REGION #<LEM/BUFFER/INTERNAL:POINT (1, 0) "" {1204DB6653}> #<LEM/BUFFER/INTERNAL:POINT (1, 0) "" {1204DB66A3}> :SYNTAX-TABLE NIL :RECURSIVE-CHECK T)
5: (LEM-CORE::SYNTAX-SCAN-WINDOW #<LEM-CORE:WINDOW {12075D8093}>)
[...]

PR #2313, opened 2026-09-21
tree-sitter: require libts-wrapper in tree-sitter-available-p

MasterCsquare:
This address issue #2272
Only tested in Archlinux


vindarel:
Thank you :pray: 

and for the development comments

PR #2315, opened 2026-09-22
fix(nix.flake warning): stdenv.* . Also switching to latest release + also cleaner formatting

yVieta:
simular issue to https://github.com/nix-community/home-manager/issues/9806
For formattig  alejandra was used: https://search.nixos.org/packages?channel=26.05&query=alej#show=alejandra


vindarel:
Hi, thank you.

Is it a good idea to run on unstable, instead of a pinned version?

I can't judge this PR myself and I'm ready to merge. Ideally some else will test and validate the changes.

Also formatting would be better on its own separate commit ;)

yVieta:
The unstable branch has the latest syntax and logic of nix which has less conflicts for flake configurations. An another option would using nix default (via default.nix) to build Lem without using flake.

